PR #18, opened 2021-10-28
Synchronous Hyperband

mseeger:
*Issue #, if available:*

*Description of changes:*
This implements synchronous Hyperband. I am putting this out for comments, but unit tests are still missing, I may not get such done before I leave for holiday.

### DIFF
--- a/benchmarks/launch_utils.py
+++ b/benchmarks/launch_utils.py
@@ -329,9 +329,10 @@ def make_searcher_and_scheduler(params) -> (dict, dict):
         random_seed_offset = 0
     random_seed = (random_seed_offset + params['run_id']) % (2 ** 32)
     scheduler_options = {'random_seed': random_seed}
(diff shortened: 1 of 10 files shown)


aaronkl (approved):

geoalgo:
I realize now that my review was never sent (from 3 weeks ago). I had essentially two comments: 
1. can `train_height` be used instead of nas201 in 
`examples/launch_nasbench201_hyperband_sync.py`? This is because 1) all schedulers examples until now use height 2) nas201 is not currently available.
2. can you add a test in `test_schedulers_api.py`?

Those could be done in a later PR but are important in my opinion.

mseeger:
Hello, I agree I need to do a little more work on this one.

I will first convert it to the new naming, then address David's issues.

aaronkl:
Can we delete this branch?

PR #37, opened 2021-11-19
Extending docs concerning remote launching

mseeger:
*Issue #, if available:*

*Description of changes:*
Fixes bug in launch_hpo.py, and extends README.md in order to explain the requirement of building a docker image for
remote tuning, and providing some hints what to do if this fails.


README.md
```diff
@@ -336,15 +336,40 @@ Or run the following example that evaluates trials on SageMaker.
...
 ```bash
 python examples/launch_height_sagemaker_remotely.py
 ```
+If you run this for the first time, it will take a while, building a docker image with the
+Syne Tune dependencies and pushing it to ECR. This has to be done only once, even if Syne
+Tune source code is modified later on.
+
+If running this example script fails, you are probably not setup to build docker images and
+push them to ECR on your local machine. In order to enable this, you need to install both
+`awscli` and `docker` (see [this link](https://docs.docker.com/get-docker/)),
```
geoalgo, line 355: Can we add a link to awscli install instructions?

README.md
```diff
@@ -336,15 +336,40 @@ Or run the following example that evaluates trials on SageMaker.
...
+and make sure the docker demon is running. However, since many Syne Tune workflows
+do not require building docker images, you may alternatively build the required image
+on a SageMaker notebook instance. To do so, log into such an instance and install
+Syne Tune there, then run:
+```bash
+cd container
+bash build_syne_tune_container.sh
+```
```
geoalgo, line 363: Can we remove the part after "However, ...". Running in a SageMaker notebook instance is not a unit step (we would have to describe all the steps precisely), also I do not think it is simpler than running a docker daemon locally.

README.md
```diff
@@ -336,15 +336,40 @@ Or run the following example that evaluates trials on SageMaker.
...
+Assuming that `launch_height_sagemaker_remotely.py` is working for you now, you
+should note that the script returns immediately after starting the experiment, which is
+running as a SageMaker training job. This allows you to run many experiments in
+parallel, possibly by using the [command line launcher](docs/command_line.md).
 
-To run on SageMaker, you can also use any custom docker images available on ECR.
-See [launch_height_sagemaker_custom_image.py](examples/launch_height_sagemaker_custom_image.py)
+You can also run Syne Tune on training code in a custom docker image, circumventing
+SageMaker frameworks entirely. See
+[launch_height_sagemaker_custom_image.py](examples/launch_height_sagemaker_custom_image.py)
```
geoalgo, line 372: I am not sure this formulation is better, could we keep the previous one? (this is not circumventing SageMaker frameworks since it uses its base class)

docs/command_line.md
```diff
@@ -272,9 +266,9 @@ experiments remotely from your laptop).
 
 ## How Results are Stored and Downloaded
 
-This is detailed in [Output of a tuning job](../README.md). If tuning is run
-remotely on SageMaker, results are uploaded to S3 in regular intervals. When
-the CLI is used, additional points apply:
+This is detailed in [Output of a tuning job](../README.md#launching-a-tuning-job).
```
geoalgo, line 269: Cool, I did not know how to add link to specific sections.

geoalgo (commented):
Thanks for doing those changes, only have small comments.

README.md
mseeger: OK, but can I add a comment in parentheses, saying that awscli and docker will be all installed on a notebook instance?

I am really thinking about our customers here. Many will be using notebooks instances anyway, exactly because they don't want to install things like docker.

What I am describing here is exactly something that affects me right now. I cannot reinstall docker, because it does not support my system anymore (I got the laptop maybe 2 years ago). And linking the docker main page is not a complete description on how to get docker running. On Mac OS, starting the docker daemon is really quite tricky, for example.

docs/command_line.md
mseeger: Haha, I found it in one of the docs you wrote, but maybe you simply copied this from elsewhere

README.md
geoalgo: Ok sure, my point is that for SageMaker users will know how to set a SM notebook but probably not others so this comment might confuse them. I think SageMaker often assume that you have a docker setup somehow.

Not sure why docker would not run on your machine, it should be supported for (much) older machine, lets discuss this one offline.

geoalgo (approved):

PR #44, opened 2021-11-24
Add tests for python 3.6, 3.7 and 3.9, update README.md badge

geoalgo:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/check-code-examples.yml
+++ b/.github/workflows/check-code-examples.yml
@@ -1,8 +1,4 @@
-
-# This workflow will install Python dependencies, run tests and lint with a single version of Python
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
-
-name: Python application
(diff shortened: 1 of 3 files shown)


.github/workflows/check-code-examples.yml
```diff
@@ -1,8 +1,4 @@
-
-# This workflow will install Python dependencies, run tests and lint with a single version of Python
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
-
-name: Python application
+name: Check code quality and examples on python3.8
```
aaronkl, line 1: Typo? Since we later also check other python versions

aaronkl (approved):

.github/workflows/check-code-examples.yml
geoalgo: No it is intentional, we only run examples against python 3.8 to save test runtime.

PR #45, opened 2021-11-24
Synchronous Hyperband

mseeger:
*Issue #, if available:*
Inherited from #18 

*Description of changes:*
Adding some tests


mseeger:
I'll still add one more test, otherwise this is good to go. Had to open another PR due to renaming.

mseeger:
Hello, I now implemented the final test (I extended test_scheduler_api.py for sync schedulers). Along the way, I also made sure on_trial_error is correctly implemented.

tst/schedulers/test_schedulers_api.py
```diff
@@ -96,3 +99,72 @@ def test_schedulers_api(scheduler):
...
+    SynchronousGeometricHyperbandScheduler(
+        config_space=config_space,
+        max_resource_level=max_t,
+        brackets=3,
+        resource_attr=resource_attr,
+        batch_size=sync_batch_size,
+        metric=metric1,
+        max_resource_attr='steps'),
+])
+def test_sync_schedulers_api(scheduler):
```
geoalgo, line 113: This solution works for me, is there a particular reason why we would not test this scheduler among others? It should still obey the same API right?

geoalgo (approved):

tst/schedulers/test_schedulers_api.py
mseeger: Good question: A sync scheduler with a certain batch_size (say: 4) can expect to receive batch_size suggest calls before any on_trial_result calls. You can expect there is a suggest and a collect phase. At least, this is what the Tuner code implies. It really calls suggest batch_size ( = num_workers) time in a row, and only later then queries results.

This is how I implemented sync HB, it is simple and clean that way. You are right that *some* sync schedulers do not make use of this (e.g., FIFOScheduler can be run sync).

PR #47, opened 2021-11-25
add script to download and convert nb201 data

aaronkl:
Script to download and convert original NASBench201 data into blackbox_repository format

### DIFF
--- a/blackbox_repository/conversion_scripts/recipes.py
+++ b/blackbox_repository/conversion_scripts/recipes.py
@@ -1,10 +1,10 @@
-from blackbox_repository.conversion_scripts.scripts.nas201_import import generate_nas201
+from blackbox_repository.conversion_scripts.scripts.nasbench201_import import generate_nasbench201
 from blackbox_repository.conversion_scripts.scripts.fcnet_import import generate_fcnet
 from blackbox_repository.conversion_scripts.scripts.icml2020_import import generate_deepar, generate_xgboost
 
 generate_blackbox_recipe = {
     "icml-deepar": generate_deepar,
     "icml-xgboost": generate_xgboost,
-    "nas201": generate_nas201,
+    "nasbench201": generate_nasbench201,
     "fcnet": generate_fcnet,
 }
--- a/blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
+++ b/blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
@@ -184,7 +184,7 @@ def generate_nasbench201():
 
     with catchtime("converting"):
         bb_dict = {}
-        for dataset in ['cifar10', 'cifar100', 'ImageNet16-120'][:1]:
+        for dataset in ['cifar10', 'cifar100', 'ImageNet16-120']:
             print(f"converting {dataset}")
             bb_dict[dataset] = convert_dataset(data, dataset)
 
--- a/blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
+++ b/blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
@@ -1,13 +1,12 @@
 import bz2
 import pickle
-
-import matplotlib.pyplot as plt
-
 import pandas as pd
 import numpy as np
+
 from blackbox_repository.blackbox_tabular import serialize, BlackboxTabular
-import sagemaker_tune.search_space as sp
 from blackbox_repository.conversion_scripts.utils import repository_path
+
+from syne_tune import search_space
 from syne_tune.util import catchtime
 
 
@@ -16,11 +15,12 @@ def str_to_list(arch_str):
     node_strs = arch_str.split('+')
     config = []
     for i, node_str in enumerate(node_strs):
-        inputs = list(filter(lambda x: x != '', node_str.split('|')))
-        for xinput in inputs: assert len(xinput.split('~')) == 2, 'invalid input length : {:}'.format(xinput)
+        inputs = [x for x in node_str.split('|') if x != '']
+        for xinput in inputs:
+            assert len(xinput.split('~')) == 2, 'invalid input length : {:}'.format(xinput)
         inputs = (xi.split('~') for xi in inputs )
 
-        config.extend(op for (op, IDX) in inputs)
+        config.extend(op for (op, idx) in inputs)
 
     return config
 
@@ -63,14 +63,11 @@ def convert_dataset(data, dataset):
     n_seeds = 3
 
     objective_evaluations = np.empty((n_hps, n_seeds, n_fidelities, n_objectives)).astype('float32')
+    name_index = {name: i for i, name in enumerate(objective_names)}
 
     def save_objective_values_helper(name, values):
         assert values.shape == (n_hps, n_seeds, n_fidelities)
 
-        name_index = dict(zip(
-            objective_names,
-            range(len(objective_names)))
-        )
         objective_evaluations[..., name_index[name]] = values
 
     ve = np.empty((n_hps, n_seeds, n_fidelities)).astype('float32')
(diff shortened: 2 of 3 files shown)


geoalgo (commented):
Could we add this one to `https://github.com/awslabs/syne-tune/blob/main/blackbox_repository/conversion_scripts/recipes.py` and replace (or remove) nas201_import.py?

blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
```diff
@@ -0,0 +1,212 @@
...
+def str_to_list(arch_str):
+
+    node_strs = arch_str.split('+')
+    config = []
+    for i, node_str in enumerate(node_strs):
+        inputs = list(filter(lambda x: x != '', node_str.split('|')))
+        for xinput in inputs: assert len(xinput.split('~')) == 2, 'invalid input length : {:}'.format(xinput)
+        inputs = (xi.split('~') for xi in inputs )
+
+        config.extend(op for (op, IDX) in inputs)
```
geoalgo, line 23: why using upper-case IDX?

blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
```diff
@@ -0,0 +1,212 @@
...
+from syne_tune.util import catchtime
+
+
+def str_to_list(arch_str):
+
+    node_strs = arch_str.split('+')
+    config = []
+    for i, node_str in enumerate(node_strs):
+        inputs = list(filter(lambda x: x != '', node_str.split('|')))
+        for xinput in inputs: assert len(xinput.split('~')) == 2, 'invalid input length : {:}'.format(xinput)
```
geoalgo, line 20: can you make it two lines? (we should add black formatting to our code to avoid the need of mentioning such points).

blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
```diff
@@ -0,0 +1,212 @@
...
+from blackbox_repository.conversion_scripts.utils import repository_path
+from syne_tune.util import catchtime
+
+
+def str_to_list(arch_str):
+
+    node_strs = arch_str.split('+')
+    config = []
+    for i, node_str in enumerate(node_strs):
+        inputs = list(filter(lambda x: x != '', node_str.split('|')))
```
geoalgo, line 19: list comprehension are generally recommended over list + filter:
```suggestion
        inputs = [x for x in node_str.split('|') if x != '']
```

blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
```diff
@@ -0,0 +1,212 @@
...
+                except KeyError:
+                    validation_error = np.nan
+                    train_error = np.nan
+                    runtime = np.nan
+                ve[ai, si, ei] = validation_error
+                te[ai, si, ei] = train_error
+                rt[ai, si, ei] = runtime
+
+    def impute(values):
+        idx = np.isnan(values)
```
geoalgo, line 97: Could you add a comment on this one explaining exactly where it comes from (missing seeds) and what is the strategy employed (taking the mean).

blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
```diff
@@ -0,0 +1,212 @@
...
+    n_fidelities = len(fidelity_values)
+    n_objectives = len(objective_names)
+    n_seeds = 3
+
+    objective_evaluations = np.empty((n_hps, n_seeds, n_fidelities, n_objectives)).astype('float32')
+
+    def save_objective_values_helper(name, values):
+        assert values.shape == (n_hps, n_seeds, n_fidelities)
+
+        name_index = dict(zip(
```
geoalgo, line 70: Nit, but the dict can be taken out or passed as argument: it is constructed every time but does not change (wont change performance but might make things clearer).

geoalgo: I also think this would be easier to read:
```suggestion
        name_index = {name: i for i, name in enumerate(objective_names)}
```

blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
```diff
@@ -0,0 +1,212 @@
...
+            range(len(objective_names)))
+        )
+        objective_evaluations[..., name_index[name]] = values
+
+    ve = np.empty((n_hps, n_seeds, n_fidelities)).astype('float32')
+    te = np.empty((n_hps, n_seeds, n_fidelities)).astype('float32')
+    rt = np.empty((n_hps, n_seeds, n_fidelities)).astype('float32')
+
+    for ai in range(n_hps):
+        for si, seed in enumerate([777, 888, 999]):
```
geoalgo, line 81: Isnt it very slow compared to assigning all the epochs at once? We can live with it if its not too slow, e.g. if it does not take hours.

blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
```diff
@@ -0,0 +1,212 @@
+import bz2
+import pickle
+
+import matplotlib.pyplot as plt
```
geoalgo, line 4: Can you put matplotlib inside the plot function? (it can fail some flake8 tests)

geoalgo (commented):
Thanks a lot for doing this. My most import comment is whether we can add this to recipes.py and remove the previous nas201 import script.

blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
```diff
@@ -0,0 +1,212 @@
...
+        'latency',
+        'flops',
+        'params',
+
+    ]
+
+    fidelity_values = np.arange(1, 201)
+    n_fidelities = len(fidelity_values)
+    n_objectives = len(objective_names)
+    n_seeds = 3
```
mseeger, line 63: One question about the seed. Currently, we are not exposing this via load, where I can only select the dataset.

But I think this can be done later, it may need some thinking. We can only do it if tables for all seeds are complete.

blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
```diff
@@ -0,0 +1,212 @@
...
+        for j, hp in enumerate(config):
+            hps[f"hp_x{j}"].append(hp)
+
+    hyperparameters = pd.DataFrame(
+        data=hps,
+        columns=hp_cols
+    )
+
+    objective_names = [
+        'valid_error',
```
mseeger, line 51: One question, before I do something wrong downstream.
In the old dataset (on our S3 bucket), we had lc_valid_epoch_X, which was accuracy * 100, and mode was 'max'.
Here, we now have validation error, and the mode is 'min' ?

Also, 'runtime' is runtime since start?

aaronkl: yes that's right we now return the validation error. And yes runtime is the runtime since start

mseeger (approved):
Great. Once this is in, I send another PR which refactors launch_hpo.py and the benchmarks

mseeger:
Oh one more thing, could you refactor examples/launch_simulated_benchmark.py and make sure it runs? There have been name changes (e.g., 'nas201' -> 'nasbench201'; metric names).

blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
```diff
@@ -0,0 +1,216 @@
...
+                        f.write(chunk)
+
+        download_file_from_google_drive('1vzyK0UVH2D3fTpa1_dSWnp1gvGpAxRul', file_name)
+
+    f = bz2.BZ2File(file_name, 'rb')
+    data = pickle.load(f)
+
+    with catchtime("converting"):
+        bb_dict = {}
+        for dataset in ['cifar10', 'cifar100', 'ImageNet16-120'][:1]:
```
mseeger, line 187: 'cifar10' was 'cifar10-valid' before. This was always annoying, I am happy to change it. But are you sure this works here? In the current code by David, there is no explicit mentioning of the dataset names, so I suppose it is still 'cifar10-valid' there.

blackbox_repository/conversion_scripts/recipes.py
```diff
@@ -1,10 +1,10 @@
-from blackbox_repository.conversion_scripts.scripts.nas201_import import generate_nas201
+from blackbox_repository.conversion_scripts.scripts.nasbench201_import import generate_nasbench201
 from blackbox_repository.conversion_scripts.scripts.fcnet_import import generate_fcnet
 from blackbox_repository.conversion_scripts.scripts.icml2020_import import generate_deepar, generate_xgboost
 
 generate_blackbox_recipe = {
     "icml-deepar": generate_deepar,
     "icml-xgboost": generate_xgboost,
-    "nas201": generate_nas201,
+    "nasbench201": generate_nasbench201,
```
geoalgo, line 8: If we make this change, we should also update the example: https://github.com/awslabs/syne-tune/blob/main/examples/launch_simulated_benchmark.py#L86

blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
geoalgo: I agree cifar10 is easier. 

Not sure I got your other comment Matthias, the name of the dataset appears in the mapping in bb_dict in the repository.

geoalgo, line 187: Is the `[:1]` intended? It seems like a left-over.

geoalgo (commented):
LGTM, I only have minor comments.

blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
mseeger: So it is still cifar10-valid ? What I meant is that Aaron's code has these names explicitly in the code, but in David's code I do not find them, so I suppose they come from the file that is loaded?

blackbox_repository/conversion_scripts/recipes.py
mseeger: +1

blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
aaronkl: AFAIK there are two different version of cifar10 in nb201, and uses the validation error as optimization criterion and on e the test error. To be consistent with the other two datasets cifar100 and Imagenet-16, I switched now to the version based on the test error. Unfortunately for none of the dataset we have another hold out test error as we have for  FCNet based NAS-HPO-Bench

geoalgo (approved):

blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
geoalgo: Different seeds can be accessed from blackbox by passing `seed=...` as argument (we took roughly the same interface as HPOBench to improve compatibility).

PR #52, opened 2021-11-29
fix dict keys

aaronkl:
use correct dict keys


mseeger (approved):
OK, would this address the nan issues I am observing?

aaronkl:
yes it should solve the nan issue

PR #54, opened 2021-11-30
Allow default region to be configured in the AWS CLI in addition to AWS_DEFAULT_REGION

master:
*Issue #, if available:*

*Description of changes:*
Default AWS region can be specified in the AWS CLI config (`~/.aws/config`) in addition to environment variables. `boto3.Session()` provides an interface to both.

### DIFF
--- a/syne_tune/remote/remote_main.py
+++ b/syne_tune/remote/remote_main.py
@@ -14,6 +14,7 @@
 Entrypoint script that allows to launch a tuning job remotely.
 It loads the tuner from a specified path then runs it.
 """
+import os
 import logging
 from argparse import ArgumentParser
 from pathlib import Path
@@ -27,10 +28,13 @@ if __name__ == '__main__':
     parser.add_argument('--store_logs', dest='store_logs', action='store_true', default=False)
     parser.add_argument('--log_level', type=int, default=logging.INFO)
     parser.add_argument('--no_tuner_logging', type=str, default='False')
+    parser.add_argument('--region_name', type=str, default='us-west-2')
     args, _ = parser.parse_known_args()
 
     root = logging.getLogger()
     root.setLevel(args.log_level)
+
+    os.environ['AWS_DEFAULT_REGION'] = args.region_name
 
     tuner_path = Path(args.tuner_path)
     logging.info(f"load tuner from path {args.tuner_path}")
--- a/syne_tune/backend/sagemaker_backend/sagemaker_backend.py
+++ b/syne_tune/backend/sagemaker_backend/sagemaker_backend.py
@@ -238,7 +238,7 @@ class SagemakerBackend(Backend):
             self.sm_estimator.dependencies = [Path(dep).name for dep in self.sm_estimator.dependencies]
 
     def initialize_sagemaker_session(self):
-        if 'AWS_DEFAULT_REGION' not in os.environ:
+        if boto3.Session().region_name is None:
             # avoids error "Must setup local AWS configuration with a region supported by SageMaker."
             # in case no region is explicitely configured
             os.environ['AWS_DEFAULT_REGION'] = 'us-west-2'
--- a/syne_tune/remote/remote_launcher.py
+++ b/syne_tune/remote/remote_launcher.py
@@ -108,7 +108,7 @@ class RemoteLauncher:
         """
         self.prepare_upload()
 
-        if 'AWS_DEFAULT_REGION' not in os.environ:
+        if boto3.Session().region_name is None:
             # launching in this is needed to send a default configuration on the tuning loop running on Sagemaker
             # todo restore the env variable if present to avoid a side effect
             os.environ['AWS_DEFAULT_REGION'] = 'us-west-2'
--- a/syne_tune/remote/remote_launcher.py
+++ b/syne_tune/remote/remote_launcher.py
@@ -231,7 +231,8 @@ class RemoteLauncher:
         """
         docker_image_name = "syne-tune-cpu-py36"
         account_id = boto3.client("sts").get_caller_identity()["Account"]
-        image_uri = f"{account_id}.dkr.ecr.us-west-2.amazonaws.com/{docker_image_name}"
+        region_name = boto3.Session().region_name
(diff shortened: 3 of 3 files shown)


geoalgo (commented):
LGTM, thanks for proposing this improvement.

Unfortunately, we do not have integration tests for SM backend (we need to set credentials for this). Could you check that this example still runs when setting `distribute_trials_on_sagemaker=True`?
https://github.com/awslabs/syne-tune/blob/main/examples/launch_height_sagemaker_remotely.py#L51

Let me know if this is working, we can then merge your change.

master:
Thanks for the pointers! Indeed, running with distributed trials required a few more changes. I updated the PR now and tested that it works with both distributed and non-distributed remote tuning in various regions.

syne_tune/remote/remote_main.py
```diff
@@ -27,11 +28,14 @@
     parser.add_argument('--store_logs', dest='store_logs', action='store_true', default=False)
     parser.add_argument('--log_level', type=int, default=logging.INFO)
     parser.add_argument('--no_tuner_logging', type=str, default='False')
+    parser.add_argument('--region_name', type=str, default='us-west-2')
```
geoalgo, line 31: It seems to me that we can avoid adding this extra argument and case in `remote_main.py` by just setting the environment variable in the caller script.

The following would be equivalent:
``` 

        # avoids error "Must setup local AWS configuration with a region supported by SageMaker."
        # in case no region is explicitely configured by providing a default region
        environment = estimator_kwargs.get("environment", {})
        if "AWS_DEFAULT_REGION" not in environment:
          environment["AWS_DEFAULT_REGION"] = boto3.Session().region_name
        tuner_estimator = PyTorch(
            # path which calls the tuner
            entry_point="remote_main.py",
            source_dir=str(self.remote_script_dir()),
            instance_type=self.instance_type,
            instance_count=1,
            role=self.role,
            py_version="py3",
            framework_version='1.6',
            image_uri=self.syne_tune_image_uri(),
            hyperparameters=hyperparameters,
            checkpoint_s3_uri=checkpoint_s3_uri,
            environment=environment,
            **self.estimator_kwargs,
        )
```

but does not require any change in`remote_main.py` while requiring a similar change in `remote_tuner.py` (e.g. setting an env variable instead of a hyperparameter).

Could we make this change unless you see a downside of this approach?

geoalgo:
Thanks testing it (and sorry for not having yet an automatic test for this). The change looks good to me, I only have a suggestion that would avoid adding extra-logic to remote main while being equivalent.

syne_tune/remote/remote_main.py
master: I experimented with this idea and I didn't manage to get it to work. My understanding is that `remote_main.py` loads a `Tuner` instance, that triggers deserialization of all underlying objects: https://github.com/awslabs/syne-tune/blob/main/syne_tune/remote/remote_main.py#L37

`SagemakerBackend` on loading creates a new `sagemaker.Session()` instance: https://github.com/awslabs/syne-tune/blob/main/syne_tune/backend/sagemaker_backend/sagemaker_backend.py#L230

That overrides the region that was set before: https://github.com/awslabs/syne-tune/blob/main/syne_tune/backend/sagemaker_backend/sagemaker_backend.py#L251

Since this code is getting executed on a remote Sagemaker instance, where neither `AWS_DEFAULT_REGION` nor `awscli` config are available, it falls back to the default here: https://github.com/awslabs/syne-tune/blob/main/syne_tune/backend/sagemaker_backend/sagemaker_backend.py#L244

The only workaround I could come up with is propagating `region_name` with a CLI argument. I think another potential solution is to query a special AWS EC2 service as suggested here: https://stackoverflow.com/a/9735663 However, I don't know if this URL is available for Sagemaker containers.

geoalgo: Thanks for the detailed answer and the investigation. Let's ship the version as it is as it is functional and perhaps we can try to do this change later.

Regarding your analysis, it is not 100% clear to me why the solution with the environment variable would not work. In particular the following point:

> Since this code is getting executed on a remote Sagemaker instance where neither AWS_DEFAULT_REGION nor awscli config are available

is not clear to me because setting `environment={"foo": "A3}` should set the environment variable "foo=A3" on the remote instance.

geoalgo (approved):

syne_tune/remote/remote_main.py
geoalgo: For instance, I just tried this standalone script which launches an estimator after setting the `AWS_DEFAULT_REGION` environment variable. 
When running on my machine the script displays "us-west-2" (default region on my machine) and "us-east-1" (region on the remote machine set through the environment variable).

```python
import os
from pathlib import Path

import boto3
from sagemaker.pytorch import PyTorch

def get_execution_role():
    """
    :return: sagemaker execution role that is specified with the environment variable `AWS_ROLE`, if not specified then
    we infer it by searching for the role associated to Sagemaker. Note that
    `import sagemaker; sagemaker.get_execution_role()`
    does not return the right role outside of a Sagemaker notebook.
    """
    client = boto3.client("iam")
    sm_roles = client.list_roles(PathPrefix="/service-role/")['Roles']
    for role in sm_roles:
        if 'AmazonSageMaker-ExecutionRole' in role['RoleName']:
            return role['Arn']

if __name__ == '__main__':
    print(boto3.Session().region_name)

    is_running_on_sagemaker = "SM_INPUT_DATA_CONFIG" in os.environ
    if not is_running_on_sagemaker:
        est = PyTorch(
            entry_point=str(Path(__file__)),
            py_version="py3",
            framework_version='1.6',
            instance_count=1,
            instance_type="ml.m5.large",
            role=get_execution_role(),
            environment={"AWS_DEFAULT_REGION": "us-east-1"}
        )
        est.fit()
```

geoalgo:
Merging it for now, I will integrate the fix to avoid passing the region as parameter after testing it.

syne_tune/remote/remote_main.py
master: Yes, you are right - there was a bug in my previous attempt that caused a wrong behavior. I'm opening a new PR with a refactored implementation.

PR #57, opened 2021-11-30
Much faster surrogate in bb-repo, dont serialize blackbox but only re…

geoalgo:
…quired information, allow to pass surrogate.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/blackbox_repository/blackbox_surrogate.py
+++ b/blackbox_repository/blackbox_surrogate.py
@@ -127,7 +127,9 @@ class BlackboxSurrogate(Blackbox):
             # convert prediction to dictionary
             return dict(zip(self.objectives_names, prediction))
         else:
-            # returns all fidelities
+            # when no fidelity is given and a fidelity space exists, we return all fidelities
+            # we construct a input dataframe with all fidelity for the configuration given to call the transformer
+            # at once which is more efficient due to vectorization
             surrogate_input_df = pd.DataFrame([surrogate_input] * len(self.fidelity_values))
             surrogate_input_df[next(iter(self.fidelity_space.keys()))] = self.fidelity_values
             objectives_values = self.surrogate_pipeline.predict(surrogate_input_df)
--- a/blackbox_repository/blackbox_surrogate.py
+++ b/blackbox_repository/blackbox_surrogate.py
@@ -128,15 +128,10 @@ class BlackboxSurrogate(Blackbox):
             return dict(zip(self.objectives_names, prediction))
         else:
             # returns all fidelities
-            objectives_values = []
-            for fidelity in self.fidelity_values:
-                surrogate_input = configuration.copy()
-                surrogate_input.update({next(iter(self.fidelity_space.keys())): fidelity})
-                # converts the returned nd-array with shape (1, num_metrics) to the list of objectives values
-                prediction = self.surrogate_pipeline.predict(pd.DataFrame([surrogate_input])).reshape(-1).tolist()
-                objectives_values.append(prediction)
-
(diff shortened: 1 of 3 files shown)


blackbox_repository/blackbox_surrogate.py
```diff
@@ -128,15 +128,10 @@ def _objective_function(
...
-            objectives_values = []
-            for fidelity in self.fidelity_values:
-                surrogate_input = configuration.copy()
-                surrogate_input.update({next(iter(self.fidelity_space.keys())): fidelity})
-                # converts the returned nd-array with shape (1, num_metrics) to the list of objectives values
-                prediction = self.surrogate_pipeline.predict(pd.DataFrame([surrogate_input])).reshape(-1).tolist()
-                objectives_values.append(prediction)
-
-            return np.stack(objectives_values)
+            surrogate_input_df = pd.DataFrame([surrogate_input] * len(self.fidelity_values))
```
mseeger, line 131: Would be good to have a comment which explains how this works here?

blackbox_repository/tabulated_benchmark.py
```diff
@@ -111,8 +118,25 @@ def blackbox(self) -> Blackbox:
...
+        return {
+            'blackbox_name': self.blackbox_name,
+            'dataset': self.dataset,
+            'surrogate': self._surrogate,
+        }
+
+    def __setstate__(self, state):
+        self.blackbox_name = state['blackbox_name']
+        self.dataset = state['dataset']
+        self._blackbox = None
```
mseeger, line 138: Very elegant, works together with this loading when first needed!

blackbox_repository/tabulated_benchmark.py
```diff
@@ -111,8 +118,25 @@ def blackbox(self) -> Blackbox:
...
+    def __getstate__(self):
+        # we serialize only required metadata information since the blackbox data is contained in the repository and
+        # its raw data does not need to be saved.
+        return {
+            'blackbox_name': self.blackbox_name,
+            'dataset': self.dataset,
+            'surrogate': self._surrogate,
+        }
+
+    def __setstate__(self, state):
```
mseeger, line 135: I'd like to learn here. Can you explain to me how Python __setstate__ works? OK, so I want to de-serialize an object, and I have stored 'state'. How do I get 'self' here? Does this call __init__ before calling __setstate__ ?

mseeger (dismissed):
Thanks a lot, this will be very useful!

blackbox_repository/tabulated_benchmark.py
geoalgo: Of-course, when pickle/dill is called to serialize `getstate` is called recursively on the object (conversely `setstate` is called when deserializing). This allows to override the serialization mechanism to store any object for instance with json or dictionary. This is a mechanism that we could use in case we want to save schedulers with a specific format, this is only required if we want to save only required information (e.g. there is no need if all fields are basic types that we all want to store).

> How do I get 'self' here? Does this call init before calling setstate ?

When the object is deserialized, `setstate` is called and it is your responsability to set fields in `self`.

Hope it helps, here is another explanation with minimal examples https://stackoverflow.com/questions/1939058/simple-example-of-use-of-setstate-and-getstate we can chat about this offline if needed.

blackbox_repository/blackbox_surrogate.py
geoalgo: Sure, sent a commit with such a comment. Can you approve again?

mseeger (approved):

PR #58, opened 2021-12-01
Support NASBench201 blackbox as benchmark again

mseeger:
*Issue #, if available:*

*Description of changes:*
- Starts reorganization of benchmarks (separation between scripts and meta-data)
- Several smaller things


blackbox_repository/tabulated_benchmark.py
```diff
@@ -143,12 +199,20 @@ def __init__(
...
+            on how to use.
         :param elapsed_time_attr:
+        :param time_this_resource_attr:
+        :param kwargs: Arguments for :class:`SimulatorBackend`
         """
-        super().__init__(elapsed_time_attr=elapsed_time_attr)
+        super().__init__(
+            elapsed_time_attr=elapsed_time_attr,
+            time_this_resource_attr=time_this_resource_attr,
+            **kwargs)
```
geoalgo, line 217: We could pick an explicit naming and remove the comment.
```suggestion
            **simulatorbackend_kwargs)
```

examples/training_scripts/resnet_cifar10/resnet_cifar10.py
```diff
@@ -50,7 +50,6 @@ def resnet_cifar10_default_params(params=None):
         'epochs': 27,
         'grace_period': 1,
         'reduction_factor': 3,
-        'max_resource_attr': 'epochs',
```
geoalgo, line 53: If we move nas201 in `benchmark/definitions/` I think we should also move those definitions but this could be done in a separate PR.

blackbox_repository/tabulated_benchmark.py
```diff
@@ -12,43 +12,67 @@
...
 
+logger = logging.getLogger(__name__)
+
 
 class _BlackboxSimulatorBackend(SimulatorBackend):
 
-    def __init__(self, elapsed_time_attr: str):
+    def __init__(self, elapsed_time_attr: str,
+                 time_this_resource_attr: Optional[str] = None,
+                 max_resource_attr: Optional[str] = None,
```
geoalgo, line 30: I would prefer to have `time_attr` and `is_cumulative` arguments instead of two strings since the state of two strings may be inconsistent (you may have two None or two strings and one is ignored).  I also believe that the explanation is simpler: "`time_attr` must represent the time spent since the beggining if `is_cumulative` is True or the time spent in the step otherwise."

benchmarks/benchmark_factory.py
```diff
@@ -51,6 +57,11 @@ def benchmark_factory(params):
         f"benchmark_name = {name} not supported, choose from:\n{supported_benchmarks()}"
     if name.startswith('nasbench201_'):
         dataset_name = name[len('nasbench201_'):]
+        if dataset_name == 'cifar10-valid':
```
aaronkl, line 60: Can we just delete this?

examples/training_scripts/nasbench201/nasbench201.py
```diff
@@ -17,156 +17,37 @@
...
-        print(
-            "WARNING: Could not obtain lock for dataset files. Trying anyway...",
-            flush=True)
-        fname_local = download_datafile(
-            dataset_path, dataset_name, dataset_s3_bucket)
-    return pandas.read_csv(fname_local)
+METRIC_ELAPSED_TIME = 'metric_elapsed_time'
 
 
 def objective(config):
```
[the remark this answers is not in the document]
aaronkl: While I agree, to benchmark the simulator backend, it would be nice to have a version of the benchmark that allows us to sleep during the evaluation

aaronkl (dismissed):

examples/training_scripts/nasbench201/nasbench201.py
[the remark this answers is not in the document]
geoalgo: Assuming we need it, we could still have a code that is generic compare to having to write a main script that sleep for every benchmark (which can be done in another PR). The use-case is not 100% clear to me, is it to test the simulator to show that it gives the same result when you sleep or when you dont?

blackbox_repository/tabulated_benchmark.py
```diff
@@ -12,43 +12,67 @@
...
+            **kwargs,
         )
+        self._time_this_resource_attr = time_this_resource_attr
+        self._max_resource_attr = max_resource_attr
 
     @property
     def resource_attr(self):
         return next(iter(self.blackbox.fidelity_space.keys()))
 
-    def config_objectives(self, config: dict) -> List[dict]:
```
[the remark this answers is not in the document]
mseeger: OK, but it would call the one in utils? I'd like to avoid copy&paste.

examples/training_scripts/nasbench201/nasbench201.py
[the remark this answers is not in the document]
mseeger: Aaron nailed it. We'd like to be able to use (some) blackboxes also with the other back-ends.

Of course, this does not mean we have to wrap every blackbox in this way, just maybe a few ones that we'd like to use in this way. It is really optional.

blackbox_repository/tabulated_benchmark.py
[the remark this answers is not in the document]
geoalgo: Ok, lets keep it (I was thinking also that we may not need the util, but we do if we want the functionality to be able to sleep).

blackbox_repository/utils.py
```diff
@@ -0,0 +1,46 @@
...
+        blackbox: Blackbox, config: dict, resource_attr: str,
+        fidelities: Optional[np.array] = None) -> List[dict]:
+    """
+    Returns all results for configuration `config` at fidelities in
+    `fidelities`.
+
+    :param blackbox: Blackbox
+    :param config: Configuration
+    :param resource_attr: Name of resource attribute
+    :param fidelities: Defaults to all fidelity values
```
geoalgo, line 29: The intent of this one is not very clear, more than default this is more of a filter (only fidelities that are in this vector are returned). Could you update the comment to better reflect that?

examples/training_scripts/nasbench201/nasbench201.py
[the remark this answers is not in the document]
aaronkl: I agree this might be a bit specific and only relevant for a subset of the benchmarks, since it's only to benchmark the simulator backend.

blackbox_repository/tabulated_benchmark.py
mseeger: Good idea

examples/training_scripts/resnet_cifar10/resnet_cifar10.py
mseeger: Exactly, I have that on my list, to separate code from meta-data for all the benchmarks.

blackbox_repository/tabulated_benchmark.py
mseeger: But then, if is_cumulative = False, you need to specify what elapsed_time_attr should be. The problem is that this name is also needed elsewhere, in the schedulers. So, for example, if I somehow construct elapsed_time_attr here, then I'd also have to tell that to schedulers.

Do you have a suggestion?

One thing to note is that it is already a bit of a stretch that a back-end determines what these attribute names are. The other back-ends do not. I see the point here, because due to the BB table, you know the column names. But it is important that also the scheduler needs to know these names, and it cannot just ask the back-end, because normally the back-end would not know.

benchmarks/benchmark_factory.py
mseeger: I wanted to try and be backward compatible.

blackbox_repository/utils.py
mseeger: OK

blackbox_repository/tabulated_benchmark.py
mseeger: Note that elapsed_time_attr must never be None, this would fail. The simulator back-end does not work if results do not contain this attribute.

examples/training_scripts/nasbench201/nasbench201.py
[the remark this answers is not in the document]
mseeger: Good idea, David, to create a generic wrapper (it will be a little difficult, because each BB has its own config arguments).

For now, we'd need this in order to compare simulator against not (with sleep).

The case dont_sleep = True is indeed a bit obsolete now. Aaron and me wanted to run some experiments, though, we were do not sleep and use the local back-end, and show that this does not lead to the right results. This would be easier than (say) feeding a BB directly into the local back-end.

blackbox_repository/tabulated_benchmark.py
geoalgo: > The problem is that this name is also needed elsewhere, in the schedulers.

Not 100% clear what you are referring to here, for me schedulers are using `resource_attr` which is typically the epoch but not the runtime as in this case.

I understood your other point: simulator backend currenly only handles cumulative time. I propose to use this code for now, but to add support for adding cumulative time in blackbox in case only a per-step information is present so that we can avoid handling this case here. Does this solution work for you?

benchmarks/benchmark_factory.py
geoalgo: Agree with Aaron that it may be easier to remove this case, this complicates our setup forever and I do not think we loose much by changing the name of one dataset.

examples/training_scripts/nasbench201/nasbench201.py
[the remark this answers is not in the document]
geoalgo: Ok I fine not doing for now but is it something we can remove once we are not using anymore since supporting this adds a extra cases to maintain?

blackbox_repository/utils.py
```diff
@@ -0,0 +1,46 @@
...
+    :param fidelities: Only results at these fidelities are returned.
+        The default is to return results at all fidelity values
+    :return: List of result dicts
+
+    """
+    assert blackbox.fidelity_values is not None, \
+        "Blackbox must come with fidelities"
+    res = []
+    if fidelities is not None:
+        fidelities = set(fidelities)
```
geoalgo, line 38: This may be slow if you have lots of fidelities (I am running an example with 1K fidelities, you need to do 1K inclusion test in a set of 1K elements). Could you add a comment similar to the one suggested?

```suggestion
        # TODO we test if fidelities belongs to a set for simplicity, in case of performance bottleneck, we could pass
        #  fidelity_min, fidelity_max instead
        fidelities = set(fidelities)
```

geoalgo (approved):
Thanks for the PR. Could you just commit the comment on the performance of `set(fidelities)`?

blackbox_repository/tabulated_benchmark.py
mseeger: I mean the cost-aware schedulers. They have an attribute called cost_attr, and for the use cases we have right now, this is equal to elapsed_time_attr.

And elapsed_time_attr is also used with simulation-based scoring (which is not yet merged in). There, you essentially need to train a cost model (like in cost-aware) to fit the training times.

benchmarks/benchmark_factory.py
mseeger: OK, then I delete it.

blackbox_repository/utils.py
mseeger: Uups, you are right, this is in fact pretty bad. I'll change the code.

mseeger: The change you suggest works fine for me.

blackbox_repository/tabulated_benchmark.py
geoalgo: Ok, it is a bit independent in my view (the different names seems to indicate that). 

In any case, is the idea of using the code for now, but to add support for adding cumulative time in blackbox later an option? (I could add support for this)

geoalgo (approved):

PR #59, opened 2021-12-01
Fixed problem in Tuner, which can cause trials to remain paused, even…

mseeger:
… though they should be resumed

*Issue #, if available:*

*Description of changes:*
I did a test whether the problem described here really happens. It does, and it causes the dropped resumes I've been seeing in SimulatorBackend. I suppose the reason we see this here is that with simulator and 8 workers, the chance of such a immediate pause->resume is higher than normal.


geoalgo (approved):
LGTM, thanks for the fix.

PR #60, opened 2021-12-02
Fix blackbox backend for remote launcher

geoalgo:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/blackbox_repository/tabulated_benchmark.py
+++ b/blackbox_repository/tabulated_benchmark.py
@@ -66,6 +66,7 @@ class _BlackboxSimulatorBackend(SimulatorBackend):
         )
         self._time_this_resource_attr = time_this_resource_attr
         self._max_resource_attr = max_resource_attr
+        self.simulatorbackend_kwargs = simulatorbackend_kwargs
 
     @property
     def resource_attr(self):
@@ -162,7 +163,6 @@ class BlackboxRepositoryBackend(_BlackboxSimulatorBackend):
         self.dataset = dataset
         self._blackbox = None
         self._surrogate = surrogate
-        self.simulatorbackend_kwargs = simulatorbackend_kwargs
 
     @property
     def blackbox(self) -> Blackbox:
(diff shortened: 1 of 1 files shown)


blackbox_repository/tabulated_benchmark.py
```diff
@@ -130,12 +130,15 @@ def __getstate__(self):
             'blackbox_name': self.blackbox_name,
             'dataset': self.dataset,
             'surrogate': self._surrogate,
+            'elapsed_time_attr': self.elapsed_time_attr,
         }
 
     def __setstate__(self, state):
+        super().__init__(elapsed_time_attr=state['elapsed_time_attr'])
```
mseeger, line 137: Thanks, this would work, but could you also add simulator_config and tuner_sleep_time, which are two other members of SimulatorBackend we would want to get and set. Then, this looks good.

geoalgo: Can we do this in another PR? Setting `simulator_config`, `tuner_sleep_time` is currently not done in `_BlackboxSimulatorBackend` (the default are used). I do not mind if we allow setting them in BlackboxSimulatorBackend but this would be a separate change.

mseeger: That is not quite true, they are set via **simulatorbackend_kwargs. And this is used by launch_hpo.py, so not doing it would break for example the setting of tuner_sleep_time.

It is just two more lines. Thanks.

BTW: Not in the PR, but would it not also work to call super().__setstate__, super().__getstate__ in the right way? I just don't know how state is organized then. Maybe it is just one flat dict ...

In any case, thanks for this overall change, because it certainly avoids huge Dill files when serializing the Tuner, so we really want to have it.

geoalgo: > That is not quite true, they are set via **simulatorbackend_kwargs. And this is used by launch_hpo.py, so not doing it would break for example the setting of tuner_sleep_time.

To be clear I do not mind doing calling `simulatorbackend_kwargs` but this was not in "main" when the PR nor this comment was sent. It was just merged with a recent PR:
 https://github.com/awslabs/syne-tune/blob/ba0b25431143c26865d28ea403f6c0b27a10e20b/blackbox_repository/tabulated_benchmark.py#L32

Now that the PR is merged, I can update with this change.

mseeger: Ah you are right!

mseeger (approved):
Thanks a lot, this looks good now

PR #64, opened 2021-12-02
Small fix

mseeger:
*Issue #, if available:*

*Description of changes:*
Another two fixes required to make things work again.

### DIFF
--- a/syne_tune/backend/simulator_backend/simulator_callback.py
+++ b/syne_tune/backend/simulator_backend/simulator_callback.py
@@ -18,6 +18,7 @@ from syne_tune.backend.simulator_backend.simulator_backend import \
 from syne_tune.tuner import Tuner
 from syne_tune.constants import ST_TUNER_TIME
 from syne_tune.stopping_criterion import StoppingCriterion
+from syne_tune.optimizer.schedulers.fifo import FIFOScheduler
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


syne_tune/backend/simulator_backend/simulator_callback.py
```diff
@@ -91,6 +92,16 @@ def on_tuning_start(self, tuner: "Tuner"):
         assert tuner.sleep_time == 0, \
             "Initialize Tuner with sleep_time = 0 if you use the SimulatorBackend"
         self._time_keeper = backend.time_keeper
+        scheduler = tuner.scheduler
```
geoalgo, line 95: Ok for now to add a special-case, but I propose to add a callback to avoid adding this special case (I can do that in a next PR).

geoalgo (approved):

syne_tune/backend/simulator_backend/simulator_callback.py
mseeger: Thanks!
The callback is specific to the SimulatorBackend, which always has this time keeper.
But you are right that this testing whether scheduler is FIFOScheduler is not good, and could be improved. However, that would likely mean an extension of the Scheduler API (say, a subclass SchedulerWithTimeKeeper). So, any scheduler which also needs time, would inherit from that, and would be treated properly in the callback.

mseeger: In general, I think, the lesson here (to me) is to avoid dependencies between the components interacting in the Tuner (mainly scheduler and backend). If this cannot be avoided, these dependencies MUST be created only at the start of tuning, and not at construction.

This is because the deserialization creates a new backend object.

I think this is a good forcing function for modular design!

PR #67, opened 2021-12-07
Fix bug in fcnet_import

mseeger:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/blackbox_repository/conversion_scripts/scripts/fcnet_import.py
+++ b/blackbox_repository/conversion_scripts/scripts/fcnet_import.py
@@ -21,6 +21,15 @@ from blackbox_repository.conversion_scripts.utils import repository_path
 from syne_tune.util import catchtime
 
 
+BLACKBOX_NAME = 'fcnet'
+
+METRIC_VALID_LOSS = 'metric_valid_loss'
+
+METRIC_ELAPSED_TIME = 'metric_elapsed_time'
+
+RESOURCE_ATTR = 'hp_epoch'
+
+
 def convert_dataset(dataset_path: Path, max_rows: int = None):
     data = h5py.File(dataset_path, "r")
     keys = data.keys()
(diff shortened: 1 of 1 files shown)


blackbox_repository/conversion_scripts/scripts/fcnet_import.py
```diff
@@ -80,9 +86,9 @@ def save_objective_values_helper(name, values):
     # (n_hps, n_seeds, n_epochs)
     # todo utilize expand dim instead of reshape
     epochs = np.repeat(np.arange(1, 101).reshape(1, -1), n_hps * n_seeds, axis=0).reshape(n_hps, n_seeds, -1)
-    runtime_per_epoch = (100 / epochs) * runtime.reshape((n_hps, n_seeds, 1))
+    elapsed_time = (epochs / 100) * runtime.reshape((n_hps, n_seeds, 1))
     
-    save_objective_values_helper('runtime', runtime_per_epoch)
+    save_objective_values_helper('elapsed_time', elapsed_time)
```
geoalgo, line 91: Could we keep the original name `runtime`? The reasoning was to import datasets as close as possible from the original source and modify it only downstream.

geoalgo (commented):
Thanks LGTM, I only have a small remark.

aaronkl (approved):
LGTM. BTW: I already implemented a script to include it as a benchmark

blackbox_repository/conversion_scripts/scripts/fcnet_import.py
mseeger: Hi David, the original field 'runtime' is runtime for the whole training job. The script converts this to "elapsed time".

In the NB201 benchmark, metric_runtime is runtime for the given epoch, and you need to say that towards the back-end. I am afraid if we now call what is elapsed time (from start to end of epoch) 'runtime' here, this will be very confusing.

geoalgo: Ok thanks, I agree.

PR #68, opened 2021-12-07
Clean up matching of configurations for (approximate) equality. Avoid…

mseeger:
…s errors with categorical HPs with float values

*Issue #, if available:*

*Description of changes:*
Fixes the issues we had (David, Aaron) with duplicate configs being suggested, but only detected later. This PR is also making sure we can use str, int, float as value type for categorical everywhere.

### DIFF
--- a/syne_tune/search_space.py
+++ b/syne_tune/search_space.py
@@ -90,6 +90,18 @@ class Domain:
         """
         raise NotImplementedError
 
+    def match_string(self, value) -> str:
+        """
+        Returns string representation of `value` (which must be of domain type)
+        which is to match configurations for (approximate) equality.
+        For discrete types (e.g., `Integer`, `Categorical`), this matches for
+        exact equality.
+
+        :param value: Value of domain type (use `cast` to be safe)
+        :return: String representation useful for matching
+        """
+        raise NotImplementedError
+
 
 class Sampler:
     def sample(self,
@@ -275,6 +287,9 @@ class Float(Domain):
         else:
             return 1
 
+    def match_string(self, value) -> str:
+        return f"{value:.7e}"
+
 
 class Integer(Domain):
     class _Uniform(Uniform):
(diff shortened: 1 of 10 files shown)


syne_tune/search_space.py
```diff
@@ -796,6 +841,24 @@ def search_space_size(config_space: Dict, upper_limit: int = 2 ** 20) -> Optiona
     return size
 
 
+def config_to_match_string(config: Dict, config_space: Dict, keys: List[str]) -> str:
```
geoalgo, line 844: Is there a need to go through a string? My understanding is that if we cast all values, we will obtain the same behavior but we would not need to use string in-between, e.g. could we do directly something like this?

```
def cast_config(...):
    res  = {}
    for key in keys:
        domain = config_space[key]
        value = config[key]
        if isinstance(domain, Domain):
           res[key] = domain.cast(value)
        else:
           res[key] = value
```

(in any case we should probably handle the case where a config_space value is not a domain)

geoalgo (commented):
Thanks for addressing this, I have a small comment regarding the need of using string.

syne_tune/search_space.py
mseeger: Hi David, almost all use cases of this go into some hash function, meaning they are used as keys in dict and set.

It is also a little tricky. The main problem were these choice variables with float values. I made this bombproof now by putting the index into the match_string instead of the value. So if you have choice([0.1, 0.01, 0.001]) and the value is 0.01, then I use 1 (index of 0.01) in the match_string.

The other one is matching float values. There, I output a fixed accuracy string representation. This means that values very close by are mapped to the same value. This is a solution a bit like in AmperCorePython (where they have this DuplicateDetector concept), but it is simpler.

Also note that we also have HyperparameterRanges.config_to_tuple, which is used just to map between config and a tuple representation. So the new method here is only used for matching. You never need to go back from a string to a config (you could not, in fact, exactly).

geoalgo: Ok thanks.

geoalgo (approved):

PR #73, opened 2021-12-14
Nashpobench

aaronkl:
Add NAS-HPO-Bench Benchmark. We use a surrogate model, trained on a subset of the data, to project hyperparameters from integer to categorical values.


benchmarks/definitions/nashpobench.py
```diff
@@ -0,0 +1,103 @@
...
+        'max_resource_level': 100,
+        'grace_period': 1,
+        'reduction_factor': 3,
+        'instance_type': 'ml.m5.large',
+        'num_workers': 4,
+        'framework': 'PyTorch',
+        'framework_version': '1.6',
+        'dataset_name': 'protein_structure',
+        'dont_sleep': dont_sleep,
+        'cost_model_type': 'linear',
```
mseeger, line 73: This probably does not apply. The cost model for NASBench201 is very specific. We can spend extra work to define a fitting cost model for this benchmark here as well, but I think it is not there yet.

benchmarks/definitions/nashpobench.py
```diff
@@ -0,0 +1,103 @@
...
+
+from sklearn.ensemble import RandomForestRegressor
+
+from syne_tune.search_space import choice, randint, uniform, loguniform
+
+from blackbox_repository.conversion_scripts.scripts.fcnet_import import \
+    METRIC_ELAPSED_TIME, METRIC_VALID_LOSS, RESOURCE_ATTR, BLACKBOX_NAME
+
+
+class SubsamplingSurrogate(object):
```
mseeger, line 25: Can you add some comment on how this works?

benchmarks/definitions/nashpobench.py
```diff
@@ -0,0 +1,103 @@
...
+
+    """
+    config_space = dict(
+        _config_space,
+        epochs=params['max_resource_level'],
+        dataset_name=params['dataset_name'],
+        dont_sleep=params['dont_sleep'],
+        blackbox_repo_s3_root=params.get('blackbox_repo_s3_root'))
+    return {
+        'script': Path(__file__).parent.parent.parent / "examples" / "training_scripts" / "nashpobench" / "nashpobench.py",
```
mseeger, line 90: As discussed with David, we may not want to have this for every tabulated blackbox.

The plus is we can then use the blackbox also with other back-ends, not just simulated. But it is not sure whether we want that.

blackbox_repository/blackbox_surrogate.py
```diff
@@ -79,14 +79,20 @@ def fit_surrogate(self, surrogate=KNeighborsRegressor(n_neighbors=1)) -> Blackbo
...
+            hp = self.configuration_space[hp_name]
             if isinstance(hp, sp.Categorical):
-                categorical.append(hp)
+                categorical.append(hp_name)
             else:
-                numeric.append(hp)
+                numeric.append(hp_name)
+
+        if self.fidelity_space is not None:
+            for fideltiy_name in self.fidelity_space:
```
mseeger, line 90: Typo

blackbox_repository/blackbox_surrogate.py
```diff
@@ -79,14 +79,20 @@ def fit_surrogate(self, surrogate=KNeighborsRegressor(n_neighbors=1)) -> Blackbo
         # gets hyperparameters types, categorical for CategoricalHyperparameter, numeric for everything else
         numeric = []
         categorical = []
-        features = list(self.configuration_space.keys())
-        if self.fidelity_space is not None:
-            features += self.fidelity_space.keys()
-        for hp in features:
+        for hp_name in self.configuration_space:
```
mseeger, line 82: For hp_name, hp in self.configuration_space.items()

examples/training_scripts/nashpobench/nashpobench.py
```diff
@@ -0,0 +1,121 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mseeger, line 1: OK, you probably need this for comparisons, right? Say, to verify the simulator back-end by comparing against local back-end with sleeping.

At some point, we should probably do this with a single wrapper (as David suggested), so we don't have to do that for every blackbox.

Note that the simulator back-end does not need such a script at all by now, because it uses the blackbox directly.

benchmarks/definitions/nashpobench.py
aaronkl: ok can I just pass `None` ?

mseeger: Just remove the line.

examples/training_scripts/nashpobench/nashpobench.py
aaronkl: I agree but it's probably better to do this in a separate PR

benchmarks/definitions/nashpobench.py
aaronkl: I agree, I left it in there for future experiments regarding the simulator backend paper

examples/training_scripts/nashpobench/nashpobench.py
mseeger: Sure.

benchmarks/definitions/nashpobench.py
mseeger: That makes sense, because that we need to do for a paper.

mseeger (approved):

benchmarks/definitions/nashpobench.py
geoalgo: If something is only needed for a paper, can we just have it in a branch? (the cost of maintaining code is much higher than the writing cost)

blackbox_repository/blackbox_surrogate.py
```diff
@@ -79,14 +79,19 @@ def fit_surrogate(self, surrogate=KNeighborsRegressor(n_neighbors=1)) -> Blackbo
         # gets hyperparameters types, categorical for CategoricalHyperparameter, numeric for everything else
         numeric = []
         categorical = []
-        features = list(self.configuration_space.keys())
-        if self.fidelity_space is not None:
-            features += self.fidelity_space.keys()
-        for hp in features:
+        for hp_name, hp in self.configuration_space.items():
```
geoalgo, line 82: Can we avoid repeating the loop twice? 

```python
surrogate_hps = [self.configurations_space.keys()]
if self.fidelity_space is not None:
  surrogate_hps.append(self.fidelity_space)
for hp in surrogate_hps:
  ...
```

benchmarks/definitions/nashpobench.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    Fits a random forest to project the discrete space back to the original space with continuous and integer
+    hyperparameters. The entire dataset is quite large (for each config we have 4 seeds, 100 epochs, 4 metrics). To
+    reduce building time, we subsample the original dataset first before feeding it to the Random Forest.
+
+    """
+    def __init__(self, ratio=0.01):
+        self.ratio = ratio
+
+    def fit(self, X, y):
```
geoalgo, line 35: Could we just use `max_samples` argument of RandomForestRegressor? It is not clear why we want to have a class.

geoalgo (changes requested):

benchmarks/definitions/nashpobench.py
aaronkl: fair point, I will drop the file and add it to a separated branch

examples/training_scripts/nashpobench/requirements.txt
```diff
@@ -0,0 +1 @@
+filelock
```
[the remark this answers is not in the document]
aaronkl: This file will also go to another branch

benchmarks/definitions/nashpobench.py
aaronkl: good catch, didn't know this argument

benchmarks/definitions/nashpobench.py
mseeger: If you remove the script in examples/training_scripts, you can set 'script' to None here, that should work. But please test it, I am not sure.

mseeger: Howeever, we should keep the nasbench201 script in examples/training_scripts, if only as a reminder that we'd like to have a general wrapper script.

geoalgo (dismissed):

blackbox_repository/blackbox_surrogate.py
```diff
@@ -79,14 +79,19 @@ def fit_surrogate(self, surrogate=KNeighborsRegressor(n_neighbors=1)) -> Blackbo
         # gets hyperparameters types, categorical for CategoricalHyperparameter, numeric for everything else
         numeric = []
         categorical = []
-        features = list(self.configuration_space.keys())
+
         if self.fidelity_space is not None:
-            features += self.fidelity_space.keys()
-        for hp in features:
+            surrogate_hps = dict()
+            surrogate_hps.update(self.configuration_space.items())
```
geoalgo, line 85: no need for the items here, e.g. `surrogate_hps.update(self.configuration_space)` works.

Also I think those 3 lines are more straightforward to read:

```suggestion
            surrogate_hps = self.configuration_space
            if self.fidelity_space is not None:
              surrogate_hps.update(self.fidelity_space)
```

blackbox_repository/blackbox_surrogate.py
```diff
@@ -106,6 +111,7 @@ def fit_surrogate(self, surrogate=KNeighborsRegressor(n_neighbors=1)) -> Blackbo
             ('model', surrogate)
         ])
 
+        print(self.X.shape)
```
geoalgo, line 114: This is probably a left-over, can you remove it?

geoalgo (commented):
LGTM, only two small comments.

benchmarks/definitions/nashpobench.py
```diff
@@ -0,0 +1,77 @@
...
+
+_config_space = {
+    "hp_activation_fn_1": choice(["tanh", "relu"]),
+    "hp_activation_fn_2": choice(["tanh", "relu"]),
+    "hp_batch_size": randint(8, 64),
+    "hp_dropout_1": uniform(0.0, 0.6),
+    "hp_dropout_2": uniform(0.0, 0.6),
+    "hp_init_lr": loguniform(0.0005, 0.1),
+    'hp_lr_schedule': choice(["cosine", "const"]),
+    'hp_n_units_1': randint(16, 512),
```
mseeger, line 30: Just a question: Could this not be lograndint as well? I think the grid has logarithmic spacing, no?

benchmarks/definitions/nashpobench.py
```diff
@@ -0,0 +1,77 @@
...
+        'metric': METRIC_VALID_LOSS,
+        'mode': 'min',
+        'resource_attr': RESOURCE_ATTR,
+        'elapsed_time_attr': METRIC_ELAPSED_TIME,
+        'max_resource_attr': 'epochs',
+        'config_space': config_space,
+        'cost_model': None,
+        'supports_simulated': True,
+        'blackbox_name': BLACKBOX_NAME,
+        'surrogate': RandomForestRegressor(max_samples=0.005, bootstrap=True),
```
mseeger, line 74: General question: For inputs which we encoded logarithmically, is this done by random forest as well? Would we not want to log-transform them before passing them to RFRegressor?

mseeger (approved):

PR #76, opened 2021-12-20
benchmark for loop example

geoalgo:
*Issue #, if available:*

*Description of changes:*


benchmarks/benchmark_loop/benchmark_definitions.py
```diff
@@ -0,0 +1,48 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class BenchmarkDefinition:
```
mseeger, line 5: Do you intend this to be limited to blackbox benchmarks? If so, we should call this something like BlackboxBenchmarkDefinition.

Otherwise, we can attempt to unify the meta-data specification for all our benchmarks. This can be done later, but we want to avoid to have to specify this 2x.

benchmarks/benchmark_loop/benchmark_main.py
```diff
@@ -0,0 +1,86 @@
...
+    combinations = list(itertools.product(method_names, range(num_seeds), benchmark_names))
+
+    print(combinations)
+    for method, seed, benchmark_name in tqdm(combinations):
+        np.random.seed(seed)
+        benchmark = benchmark_definitions[benchmark_name]
+
+        print(f"Starting experiment ({method}/{benchmark_name}/{seed}) of {experiment_tag}")
+
+        if benchmark.time_this_resource_attr is None:
```
mseeger, line 42: You can just pass benchmark.time_this_resource_attr for time_this_resource_attr, the default is None anyway

benchmarks/benchmark_loop/benchmark_main.py
```diff
@@ -0,0 +1,86 @@
...
+from benchmarks.benchmark_loop.baselines import methods
+from benchmarks.benchmark_loop.benchmark_definitions import benchmark_definitions
+
+from syne_tune.backend.simulator_backend.simulator_callback import SimulatorCallback
+from syne_tune.stopping_criterion import StoppingCriterion
+from syne_tune.tuner import Tuner
+from coolname import generate_slug
+
+
+if __name__ == '__main__':
```
mseeger, line 17: This is leaving a lot of options at the defaults (for example, rung_levels, grace_period, etc, for HyperbandScheduler).

I suppose this is intended, to keep things simple. I see.

Moving forward, and in another PR, we should maybe add several benchmark_main.py examples for the different most important setups we have, and then expose the most important args.

In the end, the use case of running everything on everything is not so useful, most users will rather want to run one method, but then maybe explore its options.

benchmarks/benchmark_loop/launch_remote.py
```diff
@@ -0,0 +1,33 @@
...
+from coolname import generate_slug
+from sagemaker.pytorch import PyTorch
+
+from syne_tune.backend.sagemaker_backend.sagemaker_utils import get_execution_role
+import syne_tune
+import blackbox_repository
+import benchmarks
+from syne_tune.util import s3_experiment_path
+
+if __name__ == '__main__':
```
mseeger, line 13: Are you planning to make this launcher more generic, so we use it to launch all sorts of experiments, for example selecting the SM framework?

Or do you rather think the user should also copy and edit this one?

I'd for sure think users may want to copy and edit benchmark_main.py, but maybe they can also make use of the same launcher?

mseeger (approved):

benchmarks/benchmark_loop/benchmark_definitions.py
geoalgo: It should be any list the user wants (possibly real/non simulated benchmark), I think it makes to have a concise standalone definition but I agree we can discuss to see whether there are some parts that would benefit from being refactored.

benchmarks/benchmark_loop/benchmark_main.py
geoalgo: Thanks.

benchmarks/benchmark_loop/benchmark_main.py
geoalgo: Yes, my view is that most of the time to default of an HPO method will not be tuned and we want to evaluate the method with its default.

> In the end, the use case of running everything on everything is not so useful, most users will rather want to run one method, but then maybe explore its options.
Perhaps yes, this is intended to be an example of a simple run everything on everything (although you can also run specific combinations but the idea is to have a simple example).

benchmarks/benchmark_loop/launch_remote.py
geoalgo: > Are you planning to make this launcher more generic, so we use it to launch all sorts of experiments, for example selecting the SM framework?
> 
> Or do you rather think the user should also copy and edit this one?

I think copy pasting is the best option for now as it lets all options open for the user with minimum amount of code. We could factor a launcher once we have more clarity but I think it is better to leave this one for later as I do not think we have enough clarity on all use-cases yet.

Note: the use-case is slightly different than for remote-launcher that allows to run **one** tuning in the background easily which is a more frequent and delimited task. I still think we should support selecting the framework in the remote-launcher, I can take a look on this one after the holidays.

geoalgo:
Updated with your comment, could you approve again?

mseeger (approved):

PR #77, opened 2021-12-21
Small fix

mseeger:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/blackbox_repository/tabulated_benchmark.py
+++ b/blackbox_repository/tabulated_benchmark.py
@@ -88,16 +88,6 @@ class _BlackboxSimulatorBackend(SimulatorBackend):
         self._seed = seed
         self._seed_for_trial = dict()
 
-    def __getstate__(self):
-        return {
-            'elapsed_time_attr': self.elapsed_time_attr,
-            'time_this_resource_attr': self._time_this_resource_attr,
-            'max_resource_attr': self._max_resource_attr,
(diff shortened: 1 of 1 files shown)


blackbox_repository/tabulated_benchmark.py
```diff
@@ -204,23 +223,22 @@ def blackbox(self) -> Blackbox:
     def __getstate__(self):
         # we serialize only required metadata information since the blackbox data is contained in the repository and
         # its raw data does not need to be saved.
-        return {
```
geoalgo, line 207: Unless there is a particular reason, avoiding to define a custom `__getstate__` in the parent is better (we would also need to write a `__setstate__` then).

So I would suggest to do only:
```
    return {
            'seed': self.seed,
            'seed_for_trial': self._seed_for_trial,
            'blackbox_name': self.blackbox_name,
            'dataset': self.dataset,
            'surrogate': self._surrogate,
            'elapsed_time_attr': self.elapsed_time_attr,
            'time_this_resource_attr': self._time_this_resource_attr,
            'max_resource_attr': self._max_resource_attr,
            'simulatorbackend_kwargs': self.simulatorbackend_kwargs,
        }
```

mseeger: Done

geoalgo (approved):

PR #78, opened 2021-12-21
Refactor TuningJobState and BaseSearcher to use trial_id instead of c…

mseeger:
…onfig

*Issue #, if available:*

*Description of changes:*
This is cleaning up searcher code to use trial_id, instead of identifying trials by their configs. The latter lead to all sorts of problems. With this change, we could also allow two trials to have the same config.


geoalgo:
Hi Matthias,

Thanks for proposing those changes. The PR is very long and will be hard to review as it is.
It seems to be doing also several things at once that are independent (trial-id, warmstart, blackbox, support of batch suggest). 

Can we split it and send one PR per topic? This would make the review process more efficient. 

Thanks

mseeger:
I'll reopen this in 3 parts

PR #88, opened 2022-01-04
Make sure that synchronous schedulers are treated correctly in Tuner

mseeger:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/syne_tune/tuner.py
+++ b/syne_tune/tuner.py
@@ -24,6 +24,7 @@ from syne_tune.backend.sagemaker_backend.sagemaker_backend import SagemakerBacke
 from syne_tune.backend.trial_status import Status, Trial
 from syne_tune.constants import ST_TUNER_CREATION_TIMESTAMP, ST_TUNER_START_TIMESTAMP
 from syne_tune.optimizer.scheduler import SchedulerDecision, TrialScheduler
+from syne_tune.optimizer.schedulers.synchronous.hyperband import SynchronousScheduler
 from syne_tune.tuner_callback import TunerCallback, StoreResultsCallback
 from syne_tune.tuning_status import TuningStatus, print_best_metric_found
 from syne_tune.util import RegularCallback, experiment_path, name_from_base, check_valid_sagemaker_name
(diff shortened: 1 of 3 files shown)


syne_tune/tuner.py
```diff
@@ -46,7 +47,7 @@ def __init__(
             print_update_interval: float = 30.0,
             max_failures: int = 1,
             tuner_name: Optional[str] = None,
-            asynchronous_scheduling: bool = True,
+            asynchronous_scheduling: Optional[bool] = None,
```
geoalgo, line 50: Why changing to Optional?

syne_tune/tuner.py
```diff
@@ -99,11 +101,28 @@ def __init__(
 
         # inform the backend to the folder of the Tuner. This allows the local backend
         # to store the logs and tuner results in the same folder.
-        self.backend.set_path(results_root=self.tuner_path)
+        self.backend.set_path(results_root=str(self.tuner_path))
         self.callbacks = callbacks if callbacks is not None else [self._default_callback()]
 
         self.tuning_status = None
 
+    def _check_synchronous_scheduler(
```
geoalgo, line 109: Can we move this logic outside of the Tuner? (for instance in the caller or in a factory)

It seems to me that the caller or a factory could make this assert and type switch case and that it would be a better place since this logic is highly specific to one case.

syne_tune/tuner.py
mseeger: The idea is that the default now depends on the scheduler. In most cases, it is True as before. But for schedulers which require synchronous, it would be False.

mseeger: I wrote this PR because both Aaron and me ran into this problem that if you want to run sync Hyperband and not set asynchronous_scheduling=False, you get pretty non-sense errors.

syne_tune/tuner.py
mseeger: Well, factories already take this into account, but then we tell people to write launcher scripts, in which scheduler and Tuner are created "by hand" ...

mseeger: OK, the alternative here is to display a different error message in SynchronousHyperband

geoalgo (approved):

syne_tune/tuner.py
geoalgo: Makes sense, thanks for moving this check in the caller. 

The important thing that I am trying to preserve is to not add any special case in the tuner regarding scheduler or backend as it would couple it to extra logic and make it hard to maintain.

PR #89, opened 2022-01-05
add KDE searcher

aaronkl:
- adds a new KDE searcher 
- adds multi-fidelity KDE model for Hyperband
- unit test for both searchers
- allow to specify KDE searcher for the FIFO scheduler 


setup.py
```diff
@@ -29,6 +29,7 @@ def read_version():
 required_gpsearchers = load_requirements('requirements-gpsearchers.txt')
 required_bore = load_requirements('requirements-bore.txt')
 required_blackbox_repository = load_requirements('blackbox_repository/requirements.txt')
+required_kde = load_requirements('requirements-kde.txt')
```
mseeger, line 32: This is almost getting a bit fine-grained. One could also lump this together with gpsearchers

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,275 @@
...
+        List of configurations to be evaluated initially (in that order).
+        Each config in the list can be partially specified, or even be an
+        empty dict. For each hyperparameter not specified, the default value
+        is determined using a midpoint heuristic.
+        If None (default), this is mapped to [dict()], a single default config
+        determined by the midpoint heuristic. If [] (empty list), no initial
+        configurations are specified.
+    """
+
+    def __init__(
```
mseeger, line 79: I suppose this searcher does not work together with multi-fidelity schedulers, right?

Could you implement configure_scheduler as in gp_fifo_searcher, checking that the scheduler is FIFOScheduler? The error message would tell people to use MultiFidelityKernelDensityEstimator instead.

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,275 @@
...
+            configspace: Dict,
+            metric: str,
+            mode: str = "min",
+            num_min_data_points: int = None,
+            top_n_percent: int = 15,
+            min_bandwidth: float = 1e-3,
+            num_candidates: int = 64,
+            bandwidth_factor: int = 3,
+            random_fraction: float = .33,
+            points_to_evaluate: Optional[List[Dict]] = None,
```
mseeger, line 90: It would be good to have a random_seed argument here, and to manage random draws as in RandomSearcher.

Maybe I should do a subclass of BaseSearcher, where this is done ...

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,275 @@
...
+    ):
+        super().__init__(configspace=configspace, metric=metric, points_to_evaluate=points_to_evaluate)
+        self.mode = mode
+        self.metric_name = metric
+        self.num_evaluations = 0
+        self.min_bandwidth = min_bandwidth
+        self.random_fraction = random_fraction
+        self.num_candidates = num_candidates
+        self.bandwidth_factor = bandwidth_factor
+        self.points_to_evaluate = points_to_evaluate
```
mseeger, line 101: Not needed.

mseeger: In fact, this would be wrong here, because there is imputation going on

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,275 @@
...
+            min_bandwidth: float = 1e-3,
+            num_candidates: int = 64,
+            bandwidth_factor: int = 3,
+            random_fraction: float = .33,
+            points_to_evaluate: Optional[List[Dict]] = None,
+            **kwargs
+    ):
+        super().__init__(configspace=configspace, metric=metric, points_to_evaluate=points_to_evaluate)
+        self.mode = mode
+        self.metric_name = metric
```
mseeger, line 95: Can you use self._metric ?

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,275 @@
...
+        }
+        self.inv_categorical_maps = {
+            hp: dict(zip(map.values(), map.keys())) for hp, map in self.categorical_maps.items()
+        }
+
+        self.good_kde = None
+        self.bad_kde = None
+
+        self.vartypes = []
+        for hname in self.configspace:
```
mseeger, line 118: for name, hp in self.configspace.items()

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,275 @@
...
+            if isinstance(hp, sp.Integer):
+                self.vartypes.append(('o', (hp.lower, hp.upper)))
+            if isinstance(hp, sp.Float):
+                self.vartypes.append(('c', 0))
+
+        self.num_min_data_points = len(self.vartypes) if num_min_data_points is None else num_min_data_points
+        assert self.num_min_data_points >= len(self.vartypes)
+
+    @staticmethod
+    def to_feature(configspace, config, categorical_maps):
```
mseeger, line 131: It would be nicer to use HyperparameterRanges here. But I see this would be a small step away from the original code.

One thing: For int type, lower is always mapped to 0, upper to 1. Is that what you want? The default encoding of int a:b to real [0, 1] is rather mapping a to 1/(2 n), where n = b - a + 1. This makes sure that when you sample from [0, 1], then round to nearest int, this is the same as sampling from int a:b directly.

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,275 @@
...
+                return domain
+            else:
+                if isinstance(domain, sp.Categorical):
+                    index = int(values * len(domain))
+                    return categorical_map[index]
+                else:
+                    if isinstance(domain, sp.Float):
+                        return values * (domain.upper - domain.lower) + domain.lower
+                    else:
+                        return int(values * (domain.upper - domain.lower) + domain.lower)
```
mseeger, line 159: Maybe round here?

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,275 @@
...
+
+        res = {}
+        curr_pos = 0
+        for k, domain in configspace.items():
+            if hasattr(domain, "sample"):
```
mseeger, line 164: Better: isinstance(domain, sp.Domain)

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,275 @@
...
+                return [(value - domain.lower) / (domain.upper - domain.lower)]
+
+        return np.hstack([
+            numerize(value=config[k], domain=v, categorical_map=categorical_maps.get(k, {}))
+            for k, v in configspace.items()
+            if isinstance(v, sp.Domain)
+        ])
+
+    @staticmethod
+    def from_feature(configspace, feature_vector, inv_categorical_maps):
```
mseeger, line 146: Why are from_feature, to_feature static?

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,275 @@
...
+            return -result[self.metric_name]
+
+    def _update(self, trial_id: str, config: Dict, result: Dict):
+        self.X.append(self.to_feature(
+            config=config,
+            configspace=self.configspace,
+            categorical_maps=self.categorical_maps,
+        ))
+        self.y.append(self.to_objective(result))
+        self.train_kde(np.array(self.X), np.array(self.y))
```
mseeger, line 190: You should only do this in get_config, when you really need it.

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,275 @@
...
+    def get_config(self, **kwargs):
+        # if not done with points_to_evaluate, pick the next one
+        if self.points_to_evaluate is not None and self.num_evaluations < len(self.points_to_evaluate):
+            suggestion = self.points_to_evaluate[self.num_evaluations]
+        else:
+            if self.good_kde is None and self.bad_kde is None:
+                # if not enough suggestion made, sample randomly
+                suggestion = {k: v.sample()
+                if isinstance(v, sp.Domain) else v for k, v in self.configspace.items()}
+            elif np.random.rand() < self.random_fraction:
```
mseeger, line 201: Same code: Make this or in previous if.

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,275 @@
...
+                suggestion = {k: v.sample()
+                if isinstance(v, sp.Domain) else v for k, v in self.configspace.items()}
+            else:
+                l = self.good_kde.pdf
```
mseeger, line 205: One of these two could be None.

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,275 @@
...
+                        if vartype == 'c':
+                            # continuous parameter
+                            bw = self.bandwidth_factor * bw
+                            candidate.append(sps.truncnorm.rvs(-m / bw, (1 - m) / bw, loc=m, scale=bw))
+                        else:
+                            # categorical or integer parameter
+                            if np.random.rand() < (1 - bw):
+                                candidate.append(m)
+                            else:
+                                if vartype == 'o':
```
mseeger, line 230: Is this again a random sample from the domain? If so, why not use domain.sample() ?

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
mseeger: I'll do this, and then we change the code here later

syne_tune/optimizer/schedulers/searchers/multi_fidelity_kde_searcher.py
```diff
@@ -0,0 +1,118 @@
...
+        List of configurations to be evaluated initially (in that order).
+        Each config in the list can be partially specified, or even be an
+        empty dict. For each hyperparameter not specified, the default value
+        is determined using a midpoint heuristic.
+        If None (default), this is mapped to [dict()], a single default config
+        determined by the midpoint heuristic. If [] (empty list), no initial
+        configurations are specified.
+    """
+
+    def __init__(
```
mseeger, line 70: Also, please implement configure_scheduler to check the schedulers. This also deals with the resource_attr parameter

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,275 @@
...
+            metric: str,
+            mode: str = "min",
+            num_min_data_points: int = None,
+            top_n_percent: int = 15,
+            min_bandwidth: float = 1e-3,
+            num_candidates: int = 64,
+            bandwidth_factor: int = 3,
+            random_fraction: float = .33,
+            points_to_evaluate: Optional[List[Dict]] = None,
+            **kwargs
```
mseeger, line 91: kwargs is not used

syne_tune/optimizer/schedulers/searchers/multi_fidelity_kde_searcher.py
```diff
@@ -0,0 +1,118 @@
...
+            mode: str = "min",
+            num_min_data_points: int = None,
+            top_n_percent: int = 15,
+            min_bandwidth: float = 0.1,
+            num_candidates: int = 64,
+            bandwidth_factor: int = 3,
+            random_fraction: float = .33,
+            resource_attr: str = 'epochs',
+            points_to_evaluate: Optional[List[Dict]] = None,
+            **kwargs
```
mseeger, line 83: Do you need that?

syne_tune/optimizer/schedulers/searchers/multi_fidelity_kde_searcher.py
```diff
@@ -0,0 +1,118 @@
...
+        super().__init__(configspace, metric, mode, num_min_data_points,
+                         top_n_percent, min_bandwidth, num_candidates, bandwidth_factor, random_fraction,
+                         points_to_evaluate, **kwargs)
+
+        self.resource_attr = resource_attr
+
+        self.resource_levels = []
+
+    def _fit_kde_on_highest_resource_level(self, config, result):
+        resource_level = result[self.resource_attr]
```
mseeger, line 94: int(...) to make sure

syne_tune/optimizer/schedulers/searchers/multi_fidelity_kde_searcher.py
```diff
@@ -0,0 +1,118 @@
...
+        highest_resource_level = unique_resource_levels[idx[-1]]
+        indices = np.where(self.resource_levels == highest_resource_level)[0]
+
+        train_data = np.array([self.X[i] for i in indices])
+        train_targets = np.array([self.y[i] for i in indices])
+
+        self.train_kde(train_data, train_targets)
+
+    def _update(self, trial_id: str, config: Dict, result: Dict):
+        self._fit_kde_on_highest_resource_level(config, result)
```
mseeger, line 118: Only do this in get_config

syne_tune/optimizer/schedulers/searchers/multi_fidelity_kde_searcher.py
```diff
@@ -0,0 +1,118 @@
...
+import numpy as np
+
+from syne_tune.optimizer.schedulers.searchers.kde_searcher import KernelDensityEstimator
+
+__all__ = ['MultiFidelityKernelDensityEstimator']
+
+logger = logging.getLogger(__name__)
+
+
+class MultiFidelityKernelDensityEstimator(KernelDensityEstimator):
```
mseeger, line 24: Are you sure this also works with the async schedulers?

If so, shall we use this as another baseline in the JMLR paper, i.e. ASHA with model-based decisions by KDE?

tst/schedulers/test_multi_fidelity_kde.py
```diff
@@ -0,0 +1,33 @@
+import numpy as np
+
+from syne_tune.optimizer.schedulers.searchers.multi_fidelity_kde_searcher import MultiFidelityKernelDensityEstimator
```
geoalgo, line 3: Can you add a test in the general api tests? https://github.com/awslabs/syne-tune/blob/main/tst/schedulers/test_schedulers_api.py#L44

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
aaronkl: I am not sure what you mean. What is not needed? And what is imputed?

aaronkl: Not sure what you mean. What is not needed? And what is imputed?

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
aaronkl: you mean training the KDE? I agree

syne_tune/optimizer/schedulers/searchers/multi_fidelity_kde_searcher.py
aaronkl: Yes we could use it with asynchronous schedulers, the same way hyperopt does it, by ignoring pending evaluations.

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
mseeger: Instead of self.metric_name, use self._metric of the superclass. And (important) do not assign self.points_to_evaluate here, since this is handled by the superclass.

In general, anything you pass to superclass init, you don't have to worry here.

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
aaronkl: good catch

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
aaronkl: because they don't access any internal variables

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
aaronkl: Hmm shouldn't it be np.ceil, if I do the conversion that you described above?

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
aaronkl: This samples from the internal domain of the KDE model

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
mseeger: Yes, but you always pass configspace and inv_categorical_maps.

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
mseeger: You need to decide whether it is worth adoption HyperparameterRanges. If so, I'd just use that code. But it is some extra work.

syne_tune/optimizer/schedulers/searchers/multi_fidelity_kde_searcher.py
aaronkl: you mean **kwargs? Otherwise we get an exception since the scheduler_factor passes some additional arguments that we do not need.

setup.py
aaronkl: I would keep it like this for now, but in the future we could think of having something like 'requirements-models.txt' that bundles model-based optimizers together?

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
aaronkl: yes good point, I make it non-static

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
aaronkl: Not sure what you mean? What could be none, the model?

aaronkl: there are either both none or not

setup.py
mseeger: OK

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
```diff
@@ -0,0 +1,291 @@
...
+
+        Args:
+            scheduler: TaskScheduler
+                Scheduler the searcher is used with.
+
+        """
+        from syne_tune.optimizer.schedulers.fifo import FIFOScheduler
+
+        if not isinstance(scheduler, FIFOScheduler):
+            raise AssertionError("This searcher only works with FIFOScheduler. For multi-fidelity scheduler, such as "
```
mseeger, line 190: Actually, I realize now that HyperbandScheduler is a subclass of FIFOScheduler, so you can remove the 2nd part of the error message.

syne_tune/optimizer/schedulers/searchers/kde_searcher.py
mseeger: OK

mseeger (approved):

PR #94, opened 2022-01-10
WIP: restructure benchmarking

aaronkl:
This PR creates a new folder structure for benchmarking that accommodates:
- a list of curated benchmarks
- a nursery folder for benchmarks that will not be curated
- common utils for benchmarks


benchmarking/definitions/definition_resnet_cifar10.py
```diff
@@ -13,7 +13,7 @@
 from pathlib import Path
 
 from benchmarks.utils import get_cost_model_for_batch_size
-from examples.training_scripts.resnet_cifar10.resnet_cifar10 import \
+from benchmarking.resnet_cifar10.resnet_cifar10 import \
```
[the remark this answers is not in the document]
aaronkl: Good point. Yes that makes sense

benchmarks/definitions/nasbench201.py
```diff
@@ -13,7 +13,7 @@
 from pathlib import Path
```
mseeger, line 13: Would this file not move to benchmarking/definitions/ ?

examples/training_scripts/lstm_wikitext2/lstm_wikitext2.py
```diff
@@ -22,7 +22,7 @@
 from syne_tune.report import Reporter
```
mseeger, line 22: Would this not move to benchmarking/training_scripts/ ?

Maybe the PR tool is showing this in a funny way ...

benchmarks/definitions/nasbench201.py
aaronkl: yes this would eventually move to benchmarking/definitions. To not clutter the PR too much I just moved exemplary one benchmark

geoalgo (dismissed):
LGTM, thanks for doing this change. I anticipate that moving this many files around may cause sporadic issues in our benchmarking that we will need to address (but its worth it to improve our code structure).

PR #96, opened 2022-01-11
Support for GP bayesopt in synchronous Hyperband

mseeger:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/benchmarks/launch_utils.py
+++ b/benchmarks/launch_utils.py
@@ -254,7 +254,7 @@ def parse_args(allow_lists_as_values=True):
                         help='Relevant for synchronous Hyperband with bayesopt '
                              'searcher. If batch of size B is suggested, the '
                              'first suggest uses searcher_num_init_candidates, '
-                             'the K-1 subsequent suggests use this value',
+                             'the B-1 subsequent suggests use this value',
                         **allow_list)
     parser.add_argument('--searcher_use_old_code',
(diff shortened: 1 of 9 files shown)


benchmarks/launch_utils.py
```diff
@@ -250,6 +250,9 @@ def parse_args(allow_lists_as_values=True):
                         help='Normalize resource values to [0, 1] in '
                              'GP-expdecay surrogate model (only if '
                              'searcher_model = gp_expdecay)')
+    parser.add_argument('--searcher_num_init_candidates_for_batch', type=int,
```
aaronkl, line 253: Can we be more descriptive here? I am not entirely sure what does parameter does

aaronkl (dismissed):
LGTM just a small comment

benchmarks/launch_utils.py
mseeger: Yes. num_init_candidates is the number of configs used to seed the optimization of the acq function.
This can be rather large. Now, if you select a batch of size K, you call get_config K times, and to save time
you can say that num_init_candidates should be smaller for the K-1 entries other than the first.

This is something I implemented for the Amper use case, but it is good to have. You can ignore it, because
our num_init_candidates default is not large.

aaronkl (dismissed):

aaronkl (approved):

PR #102, opened 2022-01-12
Tutorial "Basics of Syne Tune"

mseeger:
*Issue #, if available:*

*Description of changes:*
This is using the material I used for the PB Tech summit tutorial


docs/tutorials/basics/basics_backend.md
```diff
@@ -0,0 +1,99 @@
...
+than that, as is often the case with `FIFOScheduler`, these over-heads may not
+matter much.
+
+However, for early stopping schedulers like `HyperbandScheduler`, most trials
+do not run for long, but are stopped or paused after one or a few epochs. For such
+algorithms start-up overheads can matter a lot more. On the other hand, they 
+also profit from increased parallelization.
+
+SageMaker training jobs are not only slow to start, but also have quite some
+overhead for getting stopped. Therefore, if you use ASHA or MOBSTER with the
```
geoalgo, line 87: Can you put more nuance to this sentence?

docs/tutorials/basics/basics_outlook.md
```diff
@@ -0,0 +1,50 @@
...
+not touch here. Some are established, but not basic, while others are still
+experimental. Here is an incomplete overview:
+
+* **Running many experiments in parallel**: We have stressed the importance of
+  running repetitions of experiments, as results carry quite some stochastic
+  variation. Also, there are higher-level decisions best done by trial-and-error,
+  which can be seen as "outer loop random search". Syne Tune offers facilities to
+  launch many tuning experiments in parallel, as SageMaker training jobs. See
+  [README.md](../../../README.md#launching-a-tuning-job) and
+  [launch_height_sagemaker_remotely.py](../../../examples/launch_height_sagemaker_remotely.py).
```
geoalgo, line 19: Can you add a link to the benchmark loop example as well? Something like
```suggestion
  launch many tuning experiments in parallel, as SageMaker training jobs. See
  [launch_height_sagemaker_remotely.py](../../../examples/launch_height_sagemaker_remotely.py) for an example to launch a tuning job remotely or 
  [launch_height_sagemaker_remotely.py](../../../benchmarks/benchmark_loop/README.md) for an example
  to launch a loop of different combinations remotely.
```

docs/tutorials/basics/basics_promotion.md
```diff
@@ -0,0 +1,130 @@
...
+the rung levels to the training script, but this is currently not done in
+Syne Tune.
+
+
+## Launcher Script for promotion-based ASHA and MOBSTER
+
+Once a training script supports checkpointing, the promotion variant of ASHA
+MOBSTER are obtained by selecting `type='promotion'`, see
+[launch_asha_promotion.py](scripts/launch_asha_promotion.py) and
+[launch_mobster_promotion/py](scripts/launch_mobster_promotion.py).
```
geoalgo, line 74: Having duplicate examples for all options will be hard to maintain (they will easily break with code changes). 

I wonder about options such as:
* 1 linking examples that already exist in `examples/`
* 2 linking example that are in a specific branch (this way they wont break)
* 3 giving only the sniplet to create the scheduler (given that the example only changes the scheduler option)

Creating a specific branch and linking from there (option 3) seems to me the easiest to make sure the examples still run.

docs/tutorials/basics/basics_backend.md
```diff
@@ -0,0 +1,99 @@
...
+SageMaker back-end, we recommend you use `type='promotion'` and specify
```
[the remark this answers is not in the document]
mseeger: OK, I remove this recommendation, and switch the example to use stopping, so it will work with the current code.

But I really do think that gap should be closed. If we could have a chat on how checkpointing for the SM backend works right now, I can also do the work.

As long as this is not there, we cannot even do comparisons in practice. Note that for the local back-end, the promotion variant often works quite well, and better than stopping towards later, and this is without any overheads.

And I have personally observed stopping an SM training job taking a very long time. That would bother the SM back-end only if such a job would still count as running, so no new trial can be started. Otherwise, it would still be an issue, because the experiment would run > n_workers training jobs at some point in time.

docs/tutorials/basics/basics_backend.md
mseeger: In general, if something is slow, I will tell users it is slow. If we want to have any chance of getting adoption from external people, we should agree that being honest about weaknesses is a good strategy. And we also need to be clear about the differences between the back-ends. The SM back-end has strenghts and weaknesses, and once there is back-end number 3, then that one will have strengths and weaknesses compared to the SM back-end as well.

docs/tutorials/basics/basics_outlook.md
mseeger: Good point. I am also planning to move the CL launcher to use that second option.

docs/tutorials/basics/basics_promotion.md
mseeger: I'd like to allow people to run these scripts, that is why they are complete.

We still don't have a solution for integration testing. Once we have one (with some engineering support hopefully), we could include the main tutorials by say running each launcher script until a time cutoff and check whether it does not fail.

As long as we don't have a go-to page for docs, I'd like to make sure the docs are part of the main branch.

It is just one of these things. Beyond all clever mechanisms, once in a while we need to do a sprint going over docs nd tutorials (the most important ones) and check whether they still work. We'll also get feedback from users. It is not terrible for an OS project that not every tutorial works all the time, given we fix things when told about.

mseeger: One possibility, which we could do in another PR, is to be more economical and create an "uber" launch script with options at the top.

But I am not sure. This would again be complicated, no?

docs/tutorials/basics/basics_backend.md
geoalgo: I am not saying that we should not mention that there is a delay, nor that we should mention backend pros and cons. 

I am saying that the formulation "SageMaker training jobs are not only slow to start, but also have quite some overhead for getting stopped" has a quite negative stance that can be avoided or precised. 

For instance, the formulation "quite some overhead for getting stopped" could be precised by measuring how long is the overhead so that users can decide whether it is problematic for their use-case or not.

docs/tutorials/basics/basics_promotion.md
geoalgo: I think that if we provide full snippets then we have to make sure that they work as it is a very bad experience if they don't (and having a sprint consisting of manually testing the tutorial is not a great option).

Can we put the tutorial in main but the snippets in a specific branch to ensure the examples always work?

docs/tutorials/basics/basics_backend.md
mseeger: Right now, users do not have a choice, because checkpointing is not implemented, so they need to use the stopping variant.

But the paragraph is removed.

We can continue this back and forth, but you really need to accept that I'll continue to push for improvements on the back-end side, because I think it is very important.

In order to make progress with the SM back-end, we need to indeed dive deep and consider what its properties really are. For example, I suggested to consider a variant for PBT where the different workers are run as different instances of the same SM training job. Since workers are started synchronously, this could save overhead.

As long as we do not acknowledge that there is a problem, we will end up not working on alternatives.

docs/tutorials/basics/basics_promotion.md
mseeger: No because then if the user clicks on the code, they will get an error there, because they don't have the right branch.

mseeger: To be clear: In the mid term, we need to find a better solution to docs / tutorials, such as a readthedocs.io page. There, we could try and build tutorials for a special branch, if that is possible.

As for now, we should make sure that whenever we release a new version to PyPI, all docs and tutorials are checked. Checking a tutorial's code is no different from an integration test.

mseeger: In fact, we should start to define a process for releases to PyPI.

mseeger: One idea would be to adopt readthedocs.io and build that docs page from the branch of the latest release, maybe with a "risky" option to build it from the current main.

Do you know anybody who could help with setting up readthedocs? I skimmed the docs, we need to also adopt something called sphinx, which I think is a specific way to annotate code.

geoalgo: > No because then if the user clicks on the code, they will get an error there, because they don't have the right branch.

We could very well link both the branch and explain clearly how to run the tutorial examples, eg to indicate that the snippets can be run by doing this:

```bash 
git checkout tutorial
python tutorial/example_foo.py
```

This will make sure that the tutorial doc is on main and that the code always run.

If this is not an option lets ship as it is, however I do not think we should put all tutorial examples in the integration tests as they duplicate a lot of examples already running and the CI would run significantly longer.

mseeger: OK, what I could do is add some instructions to the start of the tutorial, saying something like:
If you want to run the code examples linked in this tutorial, be sure to use this release (with instructions how to get it).

Note that the code being part of this tutorial is not meant to be run as part of the tutorial. But the Fair HPO tutorial would be a different example.

geoalgo: Ok, can you add such comment in all code examples too? (since people may find them without passing from the tutorial doc)

mseeger: OK, done that.
We'll get to a better solution, where people do not have to clone the repo.

geoalgo (approved):

PR #103, opened 2022-01-12
Launch example for constrained HPO

mseeger:
*Issue #, if available:*

*Description of changes:*
We were missing a launch example for constrained BO


mseeger:
We were missing a launch example for constrained BO, this is an important use case. I will at some point add the Fair HPO as tutorial (but markdown, not notebook), but people should also see an example how to use this.

geoalgo (approved):

PR #104, opened 2022-01-12
named classes for baselines

geoalgo:
*Description of changes:* add standalone classes corresponding to published method to allow user to find more easily random-search, BO etc. Also refactor baselines examples into a single example to avoid copy-pasting the same example a bunch of time.


examples/launch_height_baselines.py
```diff
@@ -0,0 +1,77 @@
...
+    }
+    entry_point = Path(__file__).parent / "training_scripts" / "height_example" / "train_height.py"
+    mode = "min"
+    metric = "mean_loss"
+
+
+    schedulers = [
+        RandomSearch(config_space, metric=metric, mode=mode),
+        BayesianOptimization(config_space, metric=metric, mode=mode),
+        ASHA(config_space, metric=metric, resource_attr='epoch', max_t=max_steps, mode=mode),
```
mseeger, line 45: Do we want to use all-capital names for classes? To me, all capital always sounds like constants.

examples/launch_height_baselines.py
```diff
@@ -0,0 +1,77 @@
...
+        MOBSTER(config_space, metric=metric, resource_attr='epoch', max_t=max_steps, mode=mode),
+        # commented as needs extra libraries or to save CI testing time
+        # PASHA(config_space, metric=metric, resource_attr='epoch', max_t=max_steps, mode=mode),
+        # BORE(config_space, metric=metric, mode=mode),
+        # SynchronousGeometricHyperbandScheduler(
```
mseeger, line 50: This one should be OK, it respects max_wallclock_time

examples/launch_height_baselines.py
```diff
@@ -0,0 +1,77 @@
...
+    metric = "mean_loss"
+
+
+    schedulers = [
+        RandomSearch(config_space, metric=metric, mode=mode),
+        BayesianOptimization(config_space, metric=metric, mode=mode),
+        ASHA(config_space, metric=metric, resource_attr='epoch', max_t=max_steps, mode=mode),
+        MOBSTER(config_space, metric=metric, resource_attr='epoch', max_t=max_steps, mode=mode),
+        # commented as needs extra libraries or to save CI testing time
+        # PASHA(config_space, metric=metric, resource_attr='epoch', max_t=max_steps, mode=mode),
```
mseeger, line 48: This one should also be OK, no? Should not be different to ASHA

examples/launch_height_sagemaker.py
```diff
@@ -43,7 +43,7 @@
     metric = "mean_loss"
 
     # Random search without stopping
-    scheduler = FIFOScheduler(
+    scheduler = RandomSearch(
         config_space,
         searcher='random',
```
mseeger, line 48: You can remove the searcher arg then, right?

examples/launch_height_sagemaker_custom_image.py
```diff
@@ -44,7 +44,7 @@
     metric = "mean_loss"
 
     # Random search without stopping
-    scheduler = FIFOScheduler(
+    scheduler = RandomSearch(
         config_space,
         searcher='random',
```
mseeger, line 49: remove (and everywhere else as well)

examples/launch_height_baselines.py
```diff
@@ -0,0 +1,77 @@
...
+        "steps": max_steps,
+        "width": randint(0, 20),
+        "height": randint(-100, 100)
+    }
+    entry_point = Path(__file__).parent / "training_scripts" / "height_example" / "train_height.py"
+    mode = "min"
+    metric = "mean_loss"
+
+
+    schedulers = [
```
mseeger, line 42: In order to show people how they can set extra args, and also because we found that to work well, you could add

search_options = {'num_init_random': n_workers + 2}

and then pass that to BayesianOptimization and to MOBSTER. This is an example for using **kwargs there.

I think it is important people also understand how to configure the methods, if they want to.

syne_tune/optimizer/baselines.py
```diff
@@ -0,0 +1,96 @@
+from typing import Dict
+
+from syne_tune.optimizer.schedulers.fifo import FIFOScheduler
+from syne_tune.optimizer.schedulers.hyperband import HyperbandScheduler
+from syne_tune.optimizer.schedulers.multiobjective.moasha import MOASHA
+from syne_tune.optimizer.schedulers.pbt import PopulationBasedTraining
+
```
mseeger, line 7: Do we also want to have a baseline for sync Hyperband? I could write it afterwards.

syne_tune/optimizer/baselines.py
```diff
@@ -0,0 +1,96 @@
...
+    def __init__(self, config_space: Dict, metric: str, **kwargs):
+        super(BayesianOptimization, self).__init__(
+            config_space=config_space,
+            metric=metric,
+            searcher="bayesopt",
+            **kwargs,
+        )
+
+
+class ASHA(HyperbandScheduler):
```
mseeger, line 29: I'd also include type ('stopping' or 'promotion') in the list of named args. Of course, I can get it with **kwargs, but it is certainly important to have that selection.

An alternative is to have ASHAStopping, ASHAPromotion,

or AsyncSHStopping, AsyncSHPromotion

mseeger: One thing about max_t.

This is actually not a mandatory parameter. If you set max_resource_attr, this value is obtained from config_space, which is often safer than to pass it again.

mseeger: Oh, and finally, I am not sure we should not include grace_period (with default 1). After all, this is quite an important parameter.

It can also remain in **kwargs.

Maybe it all depends how good the docs are for these baselines. We should tell people about which parameters are potentially important.

syne_tune/optimizer/baselines.py
```diff
@@ -0,0 +1,96 @@
...
+            max_t: int,
+            **kwargs,
+    ):
+        super(PASHA, self).__init__(
+            config_space=config_space,
+            metric=metric,
+            searcher="random",
+            resource_attr=resource_attr,
+            max_t=max_t,
+            hyperband_type='pasha',
```
mseeger, line 68: hyperband_type -> type

examples/launch_height_baselines.py
geoalgo: PASHA was broken in an earlier commit so I did not include it.

I left a comment:

```python
# TODO fix me, assert is thrown refusing to take PASHA arguments as valid
# PASHA(config_space=config_space, metric=metric1, resource_attr=resource_attr, max_t=max_t),
```

examples/launch_height_baselines.py
geoalgo: As indicated in the comment, I chose to only run 4 since it costs time in the CI, ultimately we can not run many many examples there as it is sort of expensive.

geoalgo: I added another comment to insist on this point.

examples/launch_height_sagemaker.py
geoalgo: Good catch, thanks.

syne_tune/optimizer/baselines.py
geoalgo: Good question. I would think that we do not because the method is strongly outperformed by async counterpart. 

I guess people will use it when they want to compare against async methods, in this sense those are probably researchers experts who know what they are doing and can configure Fifo/HB schedulers to be whatever they want (as opposed to users who just want ASHA to tune their algo).

syne_tune/optimizer/baselines.py
geoalgo: Indeed thanks.

syne_tune/optimizer/baselines.py
geoalgo: The rationale is to include only required parameters here. While `max_t` is not always required, it is required sometimes so it also went into this category.

examples/launch_height_baselines.py
geoalgo: I guess using all-capital is fine when it follows method name, we have it already in quite a few places (BORE, MOSHA etc).

syne_tune/optimizer/baselines.py
geoalgo: I propose as a first step to have only non-default parameters. We can certainly revise that if we want to put up more parameters (I guess the same question may apply to kwargs when using our schedulers).

aaronkl (approved):
LGTM. Nitpick: should we add docstrings for each baseline that reference to the original paper?

geoalgo:
> LGTM. Nitpick: should we add docstrings for each baseline that reference to the original paper?

We certainly can, lets do it in another PR.

PR #112, opened 2022-01-18
Sets the tuner_name field to be set inside Backend object.

talesa:
*Description of changes:*
Sets the tuner_name field to be set inside the `Backend` object.  
This was necessary for me to be able to rerun `docs/tutorials/basics/scripts/launch_sagemaker_backend.py` multiple times, because of SM Training instance non-unique name clashes.

### DIFF
--- a/syne_tune/tuner.py
+++ b/syne_tune/tuner.py
@@ -99,7 +99,7 @@ class Tuner:
 
         # inform the backend to the folder of the Tuner. This allows the local backend
         # to store the logs and tuner results in the same folder.
-        self.backend.set_path(results_root=self.tuner_path)
+        self.backend.set_path(results_root=self.tuner_path, tuner_name=self.name)
         self.callbacks = callbacks if callbacks is not None else [self._default_callback()]
 
         self.tuning_status = None


talesa:
Yeah, it might not be the right place to fix it, I don't have a good overview of the codebase yet!
My reasoning for fixing it this way is the following:
in https://github.com/awslabs/syne-tune/blob/main/syne_tune/backend/sagemaker_backend/sagemaker_backend.py,
L88: `self.tuner_name = None` in `__init__(...)` of `SagemakerBackend`,
then L162: `job_name = f"{self.tuner_name}-{trial_id}"`,
so we need make sure `self.tuner_name` is set to something else than `None`.
The change I'm suggesting seemed most like what was the originally intended behavior.

geoalgo (commented):
There is indeed an issue that seems to have been introduced in #65 when using both Sagemaker backend and a specific tuner name. 

The solution proposed does not work in all cases (for instance it reverts #65). 
We have a couple of different cases to support: running on your machine, running on Sagemaker, launching on your machine and running remotely, launching on your machine and running on Sagemaker. Ideally, this would be covered by our pipeline but we cannot test anything on Sagemaker yet (have to configure credentials in our pipeline to do so).

Can you close the PR and open an issue? (that SM backend and specific tuner name currently fail)

talesa:
Thank you, closing this PR, created #115!

PR #114, opened 2022-01-19
Fix bug in script of basic demo

mseeger:
*Issue #, if available:*

*Description of changes:*
Fixes issue in script of basic demo. Since "benchmarks" is not an automatic dependency, it has to be added.


syne_tune/util.py
```diff
@@ -132,13 +132,17 @@ def random_string(length: int) -> str:
     return ''.join(random.choice(pool) for _ in range(length))
 
 
-def script_checkpoint_example_path():
+def repository_root_path() -> Path:
```
geoalgo, line 135: Can you add a comment? (given that it returns the folder that contains `syne_tune` and not `syne_tune` folder)

```suggestion
def repository_root_path() -> Path:
  # returns the path that contains `syne_tune` folder which can be useful to get path of examples or utils
```

geoalgo (dismissed):

geoalgo (approved):

Issue #115, opened 2022-01-19
Duplicate SM training job names with SagemakerBackend


talesa:
Running `python docs/tutorials/basics/scripts/launch_sagemaker_backend.py` produces SM training jobs named None-0, None-1, etc which do not depend on `tuner_name`.
Rerunning the example leads to duplicate SM training job names and hence failure of the script.
This is because `tuner_name` inside the SagemakerBackend object is only ever set to None in the constructor.

Cross-ref: #112 and points by mseeger in #113.

PR #116, opened 2022-01-19
Fix path in sagemaker backend

geoalgo:
*Issue #115*

*Description of changes:* fix the issue of not setting sagemaker jobs uniquely. 
It works as intended on sagemaker backend. The small difference with #112 is that it also avoid having a path `{syne_tune_path}/{tuner_name}/{tuner_name}` when running with the local backend (e.g. it does not change how path are stored when running with local backend locally or remotely).

### DIFF
--- a/syne_tune/backend/backend.py
+++ b/syne_tune/backend/backend.py
@@ -196,12 +196,12 @@ class Backend(object):
 
     def set_path(self, results_root: Optional[str] = None, tuner_name: Optional[str] = None):
         """
-        :param results_root: the parent that should contains the results of the tuning experiment.
+        :param results_root: the local folder that should contains the results of the tuning experiment.
         Used by Tuner to indicate a desired path where the results should be written to. This is used
          to unify the location of backend files and Tuner results when possible (in the local backend).
          By default, the backend does not do anything since not all backends may be able to unify their files
          locations.
-        :param tuner_name: name of the tuner
+        :param tuner_name: name of the tuner can be used for instance to save checkpoints on remote storage.
         """
         pass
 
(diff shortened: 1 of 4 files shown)


geoalgo:
Thanks to Adam for finding this issue. 

@talesa : Sorry I could not suggest this simple change from #112 just looking at the code. The path logic is a bit complex and it is tedious that we cannot test it automatically (since we dont have access to SM in our CI pipeline).

syne_tune/backend/backend.py
```diff
@@ -196,7 +196,7 @@ def stop_all(self):
 
     def set_path(self, results_root: Optional[str] = None, tuner_name: Optional[str] = None):
         """
-        :param results_root: the parent that should contains the results of all experiments
+        :param results_root: the parent that should contains the results of the tuning experiment.
```
mseeger, line 199: Could you add a comment on what tuner_name is doing? For example, for the SM backend, it determines how SM jobs are called.

mseeger: I also do not understand one thing: Is the SageMakerBackend using results_root when set_path is called? If not, why not?

mseeger (dismissed):

syne_tune/backend/backend.py
geoalgo: > I also do not understand one thing: Is the SageMakerBackend using results_root when set_path is called? If not, why not?

No it does not since the SM backend does not store local files (as opposed to local backend). 



> Could you add a comment on what tuner_name is doing? For example, for the SM backend, it determines how SM jobs are called.

There is a comment "we use the tuner-name to set the checkpoint directory" in `SageMaker Backend`, I also added another comment in `Backend` on `set_path`.

Can you approve again?

mseeger (approved):

PR #125, opened 2022-01-25
add median stopping rule

geoalgo:
*Issue #, if available:*

*Description of changes:* add median stopping rule baseline from Golovin.

### DIFF
--- a/syne_tune/optimizer/schedulers/median_stopping_rule.py
+++ b/syne_tune/optimizer/schedulers/median_stopping_rule.py
@@ -1,0 +1,98 @@
+"""
+Example showing how to implement a new Scheduler.
+"""
+import logging
+from collections import defaultdict
+from typing import Optional, Dict, List
+
+import numpy as np
+
+from syne_tune.backend.trial_status import Trial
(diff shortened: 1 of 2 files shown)


syne_tune/optimizer/schedulers/median_stopping_rule.py
```diff
@@ -0,0 +1,98 @@
...
+        super(MedianStoppingRule, self).__init__(config_space=scheduler.config_space)
+        self.metric = scheduler.metric if metric is None else metric
+        self.sorted_results = defaultdict(list)
+        self.scheduler = scheduler
+        self.resource_attr = resource_attr
+        self.rank_cutoff = rank_cutoff
+        self.grace_time = grace_time
+        self.min_samples_required = grace_population
+        self.running_average = running_average
+        if running_average:
```
aaronkl, line 53: Wouldn't it make sense to use an exponential moving average instead?

aaronkl (approved):

syne_tune/optimizer/schedulers/median_stopping_rule.py
geoalgo: It could but it would be a step from the paper and also introduce another parameter.

PR #126, opened 2022-01-25
Equality comparison (useful for unit tests)

mseeger:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/syne_tune/search_space.py
+++ b/syne_tune/search_space.py
@@ -102,6 +102,12 @@ class Domain:
         """
         raise NotImplementedError
 
+    def __eq__(self, other) -> bool:
+        if self.sampler is None:
+            return other.sampler is None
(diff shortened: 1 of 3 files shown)


syne_tune/search_space.py
```diff
@@ -456,6 +489,11 @@ def cast(self, value):
     def match_string(self, value) -> str:
         return str(self.categories.index(value))
 
+    def __eq__(self, other) -> bool:
+        return isinstance(other, Categorical) \
+               and super(Categorical, self).__eq__(other) \
+               and self.categories == other.categories
```
hfurkanbozkurt, line 495: Do we care about the order of categories? We might want to use set equality instead.

hfurkanbozkurt (approved):

PR #127, opened 2022-01-25
adds regularized evolution

aaronkl:
*Issue #, if available:*

*Description of changes:*


syne_tune/optimizer/schedulers/regularized_evolution.py
```diff
@@ -0,0 +1,132 @@
+import numpy as np
```
mseeger, line 1: General comment: This is in fact a searcher, feeding into FIFOScheduler.

Doing it that way supports points_to_evaluate, does random seed management in the way the other schedulers do, and is easy to link into the CLI. Also, our users understand this is just another way of FIFO scheduling with a clever selection of the config.

mseeger: A simple example for how a searcher looks like, is RandomSearcher. If you really want to have very little work, you could create an internal instance of RandomSearcher to do the "points_to_evaluate" and random decisions for you. But you can also look at how RandomSearcher does it. This includes important things, like checking whether the search space is exhausted. Features we do not want to lose.

syne_tune/optimizer/schedulers/regularized_evolution.py
```diff
@@ -0,0 +1,132 @@
...
+            config_space,
+            metric: str,
+            mode: str,
+            population_size: int = 100,
+            sample_size: int = 10,
+            random_seed: int = None
+    ):
+        """
+        Implements the regularized evolution algorithm proposed by Real et al. The original implementation only
+        considers categorical hyperparameters. For integer and float parameters we sample a new value uniformly
```
geoalgo, line 31: Why not rather sampling over the distribution provided in the config space? (it may be log-uniform for instance).

syne_tune/optimizer/schedulers/regularized_evolution.py
```diff
@@ -0,0 +1,132 @@
...
+
+        # pick random hyperparameter and mutate it
+        hypers = []
+        for k, v in self.config_space.items():
+            if isinstance(v, Domain):
+                hypers.append(k)
+        name = self._random_state.choice(hypers)
+        hyperparameter = self.config_space[name]
+
+        if isinstance(hyperparameter, Categorical):
```
geoalgo, line 80: Can we assert that one of those case occur?

syne_tune/optimizer/schedulers/regularized_evolution.py
```diff
@@ -0,0 +1,132 @@
...
+            # drop current values from potential choices to not sample the same value again
+            choices = copy.deepcopy(hyperparameter.categories)
+            choices.remove(config[name])
+
+            new_value = self._random_state.choice(choices)
+
+        elif isinstance(hyperparameter, Float):
+            new_value = Float.sample(random_state=self._random_state)
```
geoalgo, line 88: Why not calling sample from the search space? The space may be logarithmic for instance.

syne_tune/optimizer/schedulers/regularized_evolution.py
```diff
@@ -0,0 +1,132 @@
...
+
+        elif isinstance(hyperparameter, Integer):
+            new_value = Integer.sample(random_state=self._random_state)
+
+        child_config[name] = new_value
+
+        return child_config
+
+    def sample_random_config(self):
+        return {k: v.sample(random_state=self._random_state) if isinstance(v, Domain) else v for k, v in
```
geoalgo, line 98: The formatting could be improved to increase readability, here is a suggestion:

```suggestion
        return {  
          k: v.sample(random_state=self._random_state) if isinstance(v, Domain) else v 
          for k, v in self.config_space.items()
        }
```

syne_tune/optimizer/schedulers/regularized_evolution.py
```diff
@@ -0,0 +1,132 @@
...
+                self.config_space.items()}
+
+    def _suggest(self, trial_id: int) -> Optional[TrialSuggestion]:
+
+        if len(self.population) < self.population_size:
+            config = self.sample_random_config()
+        else:
+            candidates = self._random_state.choice(list(self.population), size=self.sample_size)
+            parent = max(candidates, key=lambda i: i.score)
```
geoalgo, line 107: Since `"min"` is the default mode, it seems to me that it is probably more intuitive to use min there and revert the score by -1 only when the mode is max. What do you think?

syne_tune/optimizer/schedulers/regularized_evolution.py
```diff
@@ -0,0 +1,132 @@
...
+        self.metric = metric
+        self.mode = mode
+        self.population_size = population_size
+        self.sample_size = sample_size
+        self.population = deque()
+        if random_seed is None:
+            random_seed = np.random.randint(0, 2 ** 32)
+        self._random_state = np.random.RandomState(random_seed)
+
+    def mutate_arch(self, config):
```
geoalgo, line 68: Also `mutate_config` would be more consistent with the overall class naming.

tst/schedulers/test_schedulers_api.py
```diff
@@ -65,6 +66,7 @@ def make_ray_skopt():
     SimpleScheduler(config_space=config_space, metric=metric1),
     RandomSearch(config_space=config_space, metric=metric1),
     BayesianOptimization(config_space=config_space, metric=metric1),
+    RegularizedEvolution(config_space=config_space, metric=metric1, mode='min'),
```
geoalgo, line 69: Also why specifying min only in this case? For consistency with the surrounding code, I would keep the default mode.

syne_tune/optimizer/schedulers/regularized_evolution.py
mseeger: You can also check KernelDensityEstimator, which inherits from SearcherWithRandomSeed, so the random seed stuff is all there.

syne_tune/optimizer/schedulers/regularized_evolution.py
```diff
@@ -0,0 +1,132 @@
...
+
+        if len(self.population) < self.population_size:
+            config = self.sample_random_config()
+        else:
+            candidates = self._random_state.choice(list(self.population), size=self.sample_size)
+            parent = max(candidates, key=lambda i: i.score)
+
+            config = self.mutate_arch(parent.config)
+
+        return TrialSuggestion.start_suggestion(config=config)
```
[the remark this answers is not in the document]
aaronkl: In the original paper they used it for NAS without weight sharing. I am not sure how that would work with checkpointing?

syne_tune/optimizer/schedulers/regularized_evolution.py
```diff
@@ -0,0 +1,132 @@
...
+        hypers = []
+        for k, v in self.config_space.items():
+            if isinstance(v, Domain):
+                hypers.append(k)
+        name = self._random_state.choice(hypers)
+        hyperparameter = self.config_space[name]
+
+        if isinstance(hyperparameter, Categorical):
+            # drop current values from potential choices to not sample the same value again
+            choices = copy.deepcopy(hyperparameter.categories)
```
geoalgo, line 82: I think a list comprehension may also be readable (and would avoid using deepcopy):

```suggestion
            choices = [cat for cat in hyperparameter.categories if cat != config[name]]
```

syne_tune/optimizer/schedulers/regularized_evolution.py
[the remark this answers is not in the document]
geoalgo: To be clear, it is just a point of curiosity and not something that we would necessarily need for this PR.

We could support an option that also resume from the checkpoint of the mutated trial to save compute time (as in PBT). Of-course, it is only possible for some config-space (learning-rate or batch-size for instance not number of layers or cells).

```suggestion
        if self.resume_from_checkpoint:
          return TrialSuggestion.start_suggestion(config=config, checkpoint_trial_id=config_trial_id)
        else:
          return TrialSuggestion.start_suggestion(config=config)
```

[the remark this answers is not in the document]
mseeger: Another reason this should be a searcher. Then you can use it with PBT scheduling as well.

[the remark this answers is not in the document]
geoalgo: I do not think it is equivalent, using REA as a searcher in PBT gives a different algorithms than just adding checkpointing support in REA.

In any case, I think Aaron should decide what is the best approach since he contributed the method. I think they are trade-offs. On one hand we can reuse REA as a searcher in other cases if implemented as a searcher, but it also makes the method example more complex to grasp (a few base classes and edge case to be taken care of).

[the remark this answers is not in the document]
aaronkl: It's a tricky case. I will make REA a searcher for consistency. However, it's not a perfect solution. For example it might be confusing to users to be able to pass points_to_evaluate since the algorithm always starts from a random population. We might have to rethink our interfaces. I also agree with David, that REA combined with PBT is a new algorithm compared to just adding checkpoint support.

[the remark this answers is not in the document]
mseeger: Thanks, and as I said, I can help.
We should eat our own dog food, the BaseSearcher API was commonly agreed upon, it does not help if only I make use of it.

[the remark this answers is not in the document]
mseeger: points_to_evaluate is taken from Ray Tune, it just makes our code more modular, because you can specify the first config(s) to try, which I think can make sense for any scheduler.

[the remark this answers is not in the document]
mseeger: I stop here, but there are shades of "difficult to grasp". You can do anything from scratch every time you code a scheduler, and then maybe the single isolated code may be easier to grasp, because you never need to learn about common concepts. But the overall codebase will be a mess of duplicated code, and things like controlling the random seed or using user-specified default configs will be very hard.

As we otherwise put a big emphasis on clean code structure, we should follow it here as well.

syne_tune/optimizer/schedulers/searchers/regularized_evolution.py
```diff
@@ -0,0 +1,138 @@
...
+from syne_tune.search_space import Domain, Categorical, Float, Integer
+
+
+@dataclass
+class PopulationElement:
+    score: int = 0
+    config: dict = None
+
+
+class RegularizedEvolution(SearcherWithRandomSeed):
```
mseeger, line 17: Thanks!

syne_tune/optimizer/schedulers/searchers/regularized_evolution.py
```diff
@@ -0,0 +1,138 @@
...
+    def __init__(
+            self,
+            configspace,
+            metric: str,
+            mode: str = 'min',
+            population_size: int = 100,
+            sample_size: int = 10,
+            points_to_evaluate: Optional[List[Dict]] = None,
+            random_seed: int = None
```
mseeger, line 26: This is not fully correct. I'd put **kwargs here, and pass that to the superclass.

The full story is that there is a random_seed_generator which is used to generate seeds from a master seed.

syne_tune/optimizer/schedulers/searchers/regularized_evolution.py
```diff
@@ -0,0 +1,138 @@
...
+                hypers.append(k)
+        name = self.random_state.choice(hypers)
+        hyperparameter = self.configspace[name]
+
+        if isinstance(hyperparameter, Categorical):
+            # drop current values from potential choices to not sample the same value again
+            choices = [cat for cat in hyperparameter.categories if cat != config[name]]
+            new_value = self.random_state.choice(choices)
+
+        elif isinstance(hyperparameter, Float) or isinstance(hyperparameter, Integer):
```
mseeger, line 82: Could we not just say "else:" here?
For example, there are also FiniteRange. In the end, since RandomSearcher works for all Domain instances (which has the sample method), why would this one not?

syne_tune/optimizer/schedulers/searchers/regularized_evolution.py
```diff
@@ -0,0 +1,138 @@
...
+        else:
+            raise AssertionError(f"Hyperparameter {name} must be of type Float, Categorical or Integer")
+
+        child_config[name] = new_value
+
+        return child_config
+
+    def sample_random_config(self) -> Dict:
+        return {
+          k: v.sample(random_state=self.random_state) if isinstance(v, Domain) else v
```
mseeger, line 94: OK, but a searcher can in fact ignore the constant parts of configspace, they are appended back by default in TrialScheduler.

syne_tune/optimizer/schedulers/searchers/regularized_evolution.py
```diff
@@ -0,0 +1,138 @@
...
+    config: dict = None
+
+
+class RegularizedEvolution(SearcherWithRandomSeed):
+    def __init__(
+            self,
+            configspace,
+            metric: str,
+            mode: str = 'min',
+            population_size: int = 100,
```
mseeger, line 23: Is it correct to say that this one needs 100 trials to be finished to the end before it is different from random search? Wow.

syne_tune/optimizer/schedulers/searchers/regularized_evolution.py
mseeger: Do you also want to link it into searcher_factory, scheduler_factory, or do this latewr?

syne_tune/optimizer/baselines.py
```diff
@@ -83,6 +83,17 @@ def __init__(self, config_space: Dict, metric: str, mode: str, **kwargs):
         )
 
 
+class REA(FIFOScheduler):
+    def __init__(self, config_space: Dict, metric: str, **kwargs):
+        from syne_tune.optimizer.schedulers.searchers.regularized_evolution import RegularizedEvolution
```
geoalgo, line 88: Any reason for why we would do an internal import? (It seems that all the imports you are using are from `core`)

syne_tune/optimizer/schedulers/searchers/regularized_evolution.py
```diff
@@ -0,0 +1,138 @@
...
+        if initial_config is not None:
+            return initial_config
+
+        if len(self.population) < self.population_size:
+            config = self.sample_random_config()
+        else:
+            candidates = self.random_state.choice(list(self.population), size=self.sample_size)
+            parent = min(candidates, key=lambda i: i.score)
+
+            config = self.mutate_arch(parent.config)
```
geoalgo, line 111: This one is undefined now (It would make sense to diminish the population size to have this branch this tested)

syne_tune/optimizer/schedulers/searchers/regularized_evolution.py
aaronkl: let's do this in a separate PR

PR #131, opened 2022-01-26
Fixes related to benchmarking refactoring

mseeger:
*Issue #, if available:*

*Description of changes:*
Not unexpected with such a huge change, there were a number of fixes to be made.

### DIFF
--- a/benchmarking/cli/launch_hpo.py
+++ b/benchmarking/cli/launch_hpo.py
@@ -25,14 +25,13 @@ from syne_tune.backend.simulator_backend.simulator_callback import SimulatorCall
 from syne_tune.stopping_criterion import StoppingCriterion
 from syne_tune.tuner import Tuner
 from syne_tune.remote.remote_launcher import RemoteLauncher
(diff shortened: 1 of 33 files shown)


benchmarking/cli/launch_hpo.py
```diff
@@ -396,10 +396,10 @@
                 estimator_kwargs['max_run'] = int(1.01 * scheduler_timeout)
             log_level = logging.DEBUG if params['debug_log_level'] \
                 else logging.INFO
-            root_path = Path(__file__).parent.parent
+            root_path = repository_root_path()
             dependencies = [
                 str(root_path / module)
-                for module in ("benchmarks", "blackbox_repository")]
+                for module in ("benchmarking",)]
```
mseeger, line 402: Question to David: Since blackbox_repository is now under benchmarking, and so is utils, it is sufficient to have benchmarking here?

We could also use "benchmarking/blackbox_repository", "benchmarking/utils".

aaronkl (approved):

PR #132, opened 2022-01-27
fix paths to training scripts

aaronkl:
*Issue #, if available:*

*Description of changes:*


mseeger (approved):
Darn, these moving things around are always tough.

PR #138, opened 2022-01-31
add REA

aaronkl:
replaces #127 

### DIFF
--- a/syne_tune/optimizer/baselines.py
+++ b/syne_tune/optimizer/baselines.py
@@ -85,11 +85,13 @@ class BORE(FIFOScheduler):
 
 
 class REA(FIFOScheduler):
-    def __init__(self, config_space: Dict, metric: str, **kwargs):
+    def __init__(self, config_space: Dict, metric: str, population_size: int = 100, sample_size: int = 10, **kwargs):
         super(REA, self).__init__(
             config_space=config_space,
             metric=metric,
-            searcher=RegularizedEvolution(configspace=config_space, metric=metric, **kwargs),
+            searcher=RegularizedEvolution(configspace=config_space, metric=metric,
+                                          population_size=population_size, sample_size=sample_size,
+                                          **kwargs),
             **kwargs,
         )
 
--- a/syne_tune/optimizer/baselines.py
+++ b/syne_tune/optimizer/baselines.py
@@ -83,6 +83,17 @@ class BORE(FIFOScheduler):
         )
 
 
+class REA(FIFOScheduler):
+    def __init__(self, config_space: Dict, metric: str, **kwargs):
+        from syne_tune.optimizer.schedulers.searchers.regularized_evolution import RegularizedEvolution
+        super(REA, self).__init__(
+            config_space=config_space,
+            metric=metric,
+            searcher=RegularizedEvolution(configspace=config_space, metric=metric, **kwargs),
+            **kwargs,
+        )
+
+
 # dictionary that allows to also list baselines who don't need a wrapper class such as PBT.
 baselines = {
     'Random Search': RandomSearch,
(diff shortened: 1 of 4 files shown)


syne_tune/optimizer/baselines.py
```diff
@@ -83,6 +84,18 @@ def __init__(self, config_space: Dict, metric: str, mode: str, **kwargs):
         )
 
 
+class REA(FIFOScheduler):
+    def __init__(self, config_space: Dict, metric: str, population_size: int = 100, sample_size: int = 10, **kwargs):
+        super(REA, self).__init__(
+            config_space=config_space,
+            metric=metric,
+            searcher=RegularizedEvolution(configspace=config_space, metric=metric,
```
mseeger, line 92: Please don't create the searcher by hand, this is not something we want to encourage people to do. Instead:

[...], searcher='kde', **kwargs)

The specific parameters can then be written into 'search_options' in kwargs.

syne_tune/optimizer/schedulers/searchers/regularized_evolution.py
```diff
@@ -0,0 +1,133 @@
...
+        metric : str
+            Name of metric to optimize, key in result's obtained via
+            `on_trial_result`
+        mode : str
+            Mode to use for the metric given, can be 'min' or 'max', default to 'min'.
+        population_size : int
+            Size of the population.
+        sample_size : int
+            Size of the candidate set to obtain a parent for the mutation.
+        random_seed : int
```
mseeger, line 54: Remove this comment, or otherwise add comments about random_seed_generator, random_seed in other searchers.

tst/schedulers/test_schedulers_api.py
```diff
@@ -65,6 +65,7 @@ def make_ray_skopt():
     SimpleScheduler(config_space=config_space, metric=metric1),
     RandomSearch(config_space=config_space, metric=metric1),
     BayesianOptimization(config_space=config_space, metric=metric1),
+    REA(config_space=config_space, metric=metric1, population_size=1, sample_size=2),
```
mseeger, line 68: Here, you'd then do:

REA(config_space, metric=metric1, search_options={'population_size': 1, 'sample_size': 2})

syne_tune/optimizer/baselines.py
aaronkl: Can you elaborate why this is not something we want people to do? Otherwise it would cause additional friction to adapt the scheduler factory if people add a new searcher

geoalgo (approved):

syne_tune/optimizer/baselines.py
mseeger: If you look into the constructor of FIFOScheduler, then if searcher is str, all sorts of required information is passed to the searcher_factory. All of this you have to remember to pass when creating your searcher on your own, and if you do not, something may break somewhere.

On top, if you go around searcher_factory, you may combine searchers with schedulers which do not go together. Again, this will not be checked then, and you get some very weird error somewhere.

Searcher is an internal concept. it is not en par with TrialScheduler, which is an external API.

And all other entries in baselines, except for Bore, go this safe way over searcher_factory.

mseeger: I don't understand why this creates friction. If I (for example) implement a new operator for PyTorch, I also need to link it into the structures where operators need to be linked in. Otherwise, things do not even work. So, the answer would be: if you write a new searcher, link it into searcher_factory.

mseeger: To explain this better: FIFOScheduler (or TrialScheduler) has certain arguments which need to also be passed to the searcher: metric (easy), points_to_evaluate (NOT easy, because of imputation), random_seed_generator / random_seed.

The searcher_factory plus mechanism in FIFOScheduler does all of that for you.

If you create BaseSearcher on your own, you have to do all of that.

aaronkl: let's take this discussion offline

PR #142, opened 2022-02-01
Change FiniteRange so that upper, lower can be changed after creation

mseeger:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/syne_tune/search_space.py
+++ b/syne_tune/search_space.py
@@ -604,14 +604,21 @@ class FiniteRange(Domain):
         self.lower = lower
         self.upper = upper
         self.log_scale = log_scale
-        if not log_scale:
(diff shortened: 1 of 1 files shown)


geoalgo:
It seems that we would not need it as we are not changing upper/lower in the TL anymore right?

mseeger:
OK, so you changed the code? Great, I think that is safer.
But I am also OK with saying that for such simple classes, it is OK to change attributes (I usually think of dataclass that way).
Your call.

geoalgo:
I agree that setting attributes can be dangerous and is not desired so I would propose to close the PR.
In the future, if we have the need to change ranges (lower, upper) I guess we could expose `update_range` method on numeric domains.

PR #145, opened 2022-02-02
Adding g5 Sagemaker instances

talesa:


geoalgo (approved):
Thanks for the great initiative, I saw the announcement but did not add them as I did not find them in our main docs:
https://aws.amazon.com/sagemaker/pricing/

Would be great to use boto service to get those automatically at some point.

geoalgo:
> Thanks for the great initiative, I saw the announcement but did not add them as I did not find them in our main docs: https://aws.amazon.com/sagemaker/pricing/
> 
> Would be great to use boto service to get those automatically at some point.

Actually I double checked, they are g5 instances but not in all regions...

talesa:
I added it because I wasn't able to start jobs on the `ml.g5` instances (even in the region where they are available) because the Reporter tries to evaluate the cost of a non-existent instance type in
https://github.com/awslabs/syne-tune/blob/71f4c22f24446f6164c7913a3c537e2f4d30e3fa/syne_tune/report.py#L58
and then fails in line
https://github.com/awslabs/syne-tune/blob/567f822c51b1da887d3984d73cfdb4e5223c1de3/syne_tune/backend/sagemaker_backend/instance_info.py#L43
because no rows were returned.

Of course would be solved by
>Would be great to use boto service to get those automatically at some point.

geoalgo:
Ok, in the meantime we should not fail if the instance cost is not found. I sent a fix for this: #146 .

PR #148, opened 2022-02-04
Install the most recent python dependecies in the docker image

talesa:
*Description of changes:*
Genesis: I had troubles running a SM job that was successfully started when run from my local environment, but wouldn't get started through the RemoteLauncher. The reason was out-of-date version of `botocore` in the docker image.
Simply rebuilding the image didn't solve an issues, a `pip install --upgrade` flag had to be added. (This is because this out-of-date version of `botocore` comes with the base image, so without `--upgrade` other dependencies in the `requirements.txt` were downgraded from the most recent versions, rather than existing dependencies being updated.)

Additionally, the changes to `build_syne_tune_container.sh` allow to rebuild an image easily by simply rerunning the script (no need to first delete an image via `docker image rm`).

`README.md` explains the need to rebuild the image, in case they face issues, to the users.

### DIFF
--- a/container/build_syne_tune_container.sh
+++ b/container/build_syne_tune_container.sh
@@ -55,7 +55,9 @@ fi
 docker build -t ${image} . \
              --build-arg REGION=${region} \
              --build-arg DLAMI_REGISTRY_ID=${DLAMI_REGISTRY_ID} \
-             --build-arg CONTEXT=${CONTEXT}
+             --build-arg CONTEXT=${CONTEXT} \
+             --pull \
+             --no-cache
 
 # tag and push to ECR
 docker tag ${image} ${ECR_URL}/${image}
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -54,4 +54,4 @@ WORKDIR /opt/app
 ADD \
   source/requirements.txt \
   /tmp/packages/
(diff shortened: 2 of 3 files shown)


container/build_syne_tune_container.sh
```diff
@@ -55,8 +55,10 @@ fi
 docker build -t ${image} . \
              --build-arg REGION=${region} \
              --build-arg DLAMI_REGISTRY_ID=${DLAMI_REGISTRY_ID} \
-             --build-arg CONTEXT=${CONTEXT}
+             --build-arg CONTEXT=${CONTEXT} \
+             --pull \
+             --no-cache
```
geoalgo, line 60: If I am not mistaken, this would build the image every time we call the remote launcher right?

talesa: It seems to me the `build_syne_tune_container.sh` is only triggered when the image is not available in the ECR:
https://github.com/awslabs/syne-tune/blob/cc8b7b39192e15d95f2252c77b6d2294775b31ea/syne_tune/remote/remote_launcher.py#L243-L256

geoalgo: Indeed you are right. 
So, it wont matter for users as the docker image wont be built/uploaded at each call and if they run the command themselves, then the docker image is updated. 
Thanks for proposing this improvement.

geoalgo (approved):

PR #149, opened 2022-02-04
RUSHScheduler

wistuba:
*Description of changes:*
Added RUSHScheduler


syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -0,0 +1,151 @@
...
+            config_space: Dict,
+            transfer_learning_evaluations: Dict[str, TransferLearningTaskEvaluations],
+            metric: str,
+            **kwargs
+    ) -> None:
+        """
+        A transfer learning variation of Hyperband which uses previously well-performing hyperparameter configurations
+        as an initialization. Additionally, the performance of these initial configuration at the different rung levels
+        serve as a hurdle for other sampled configurations. No candidate configuration which is performing worse
+        than the best initial candidate, will be promoted.
```
geoalgo, line 36: best initial candidate**s**?

syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -0,0 +1,151 @@
...
+        Simple baseline that computes a bounding-box of the best candidate found in previous tasks to restrict the
```
geoalgo, line 37: This paragraph can be removed.

syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -0,0 +1,151 @@
...
+        points_to_evaluate = RUSHScheduler._determine_baseline_configurations(
+            config_space=config_space,
+            transfer_learning_evaluations=transfer_learning_evaluations,
+            metric=metric,
+            mode=kwargs.get('mode', 'min')
+        )
+        if 'points_to_evaluate' in kwargs:
+            points_to_evaluate += kwargs['points_to_evaluate']
+            points_to_evaluate = [dict(s) for s in set(frozenset(p.items()) for p in points_to_evaluate)]
+        kwargs['points_to_evaluate'] = points_to_evaluate
```
geoalgo, line 60: Given that you do something special with this parameter, perhaps we could add it in the constructor and document the behavior since it is reasonable but somehow non obvious? (something like `"when points_to_evaluate is not None, the provided configurations are evaluated first in addition to top performing configurations from other tasks and also serve to preemptively cut underperforming configurations"`)

syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -0,0 +1,151 @@
...
+
+        self._hyperband_scheduler = HyperbandScheduler(config_space, metric=metric, **kwargs)
```
geoalgo, line 62: Perhaps you could rename to `hyperband_kwargs` to be more explicit?

syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -0,0 +1,151 @@
...
+        self._num_init_configs = len(points_to_evaluate)
+        self._thresholds = dict()  # thresholds at different resource levels that must be met
+
+    @staticmethod
+    def _determine_baseline_configurations(config_space: Dict,
+                                           transfer_learning_evaluations: Dict[str, TransferLearningTaskEvaluations],
+                                           metric: str,
+                                           mode: str) -> List[Dict]:
+        argbest, best = (np.argmin, np.min) if mode == 'min' else (np.argmax, np.max)
```
geoalgo, line 71: Could you add a comment such as `# compute the best hyperparameter on each task`? (Or possibly we could also rename the method to something like `_compute_best_hyperparameter_per_task`)

geoalgo (commented):
Thanks for sending this, the code is quite clean. I only have small comments.

syne_tune/optimizer/schedulers/transfer_learning/rush.py
610v4nn1, line 36: Best initial candidate is ambiguous, also it would be good to specify that if the 'best initial candidate' is below the threshold normally set by SH, it will be pruned as well.

syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -0,0 +1,151 @@
...
+        record['num_init_configs'] = pickle.dumps(self._num_init_configs)
+        record['thresholds'] = pickle.dumps(self._thresholds)
+        return record
+
+    def load_state_dict(self, state_dict):
+        self._num_init_configs = pickle.loads(state_dict['num_init_configs'])
+        del state_dict['num_init_configs']
+        self._thresholds = pickle.loads(state_dict['thresholds'])
+        del state_dict['thresholds']
+        self._hyperband_scheduler(state_dict)
```
mseeger, line 108: Must be load_state_dict.

But also feel just to not implement those two methods

syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -0,0 +1,151 @@
...
+        trial_decision = self._hyperband_scheduler.on_trial_result(trial, result)
+        return self._on_trial_result(trial_decision, trial, result)
+
+    def on_trial_remove(self, trial: Trial) -> None:
+        self._hyperband_scheduler.on_trial_remove(trial)
+
+    def on_trial_complete(self, trial: Trial, result: Dict) -> None:
+        self._hyperband_scheduler.on_trial_complete(trial, result)
+
+    def suggest(self, trial_id: int) -> Optional[TrialSuggestion]:
```
mseeger, line 94: Is RUSH restricted to type == 'stopping'? If not, then for promotion you probably also want to influence the resume decisions here. I mean: suggest may decide to resume a paused trial, but then RUSH may want to say: no, stay paused.

mseeger: This is getting quite complex then. Is there any chance you could implement RUSH by just implementing a new RungSystem? We currently have StoppingRungSystem, PromotionRungSystem, PASHARungSystem, CostPromotionRungSystem, so you could just do another one, no?

I could help, and if you like to refactor something there, please feel totally free.

syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def load_state_dict(self, state_dict):
+        self._num_init_configs = pickle.loads(state_dict['num_init_configs'])
+        del state_dict['num_init_configs']
+        self._thresholds = pickle.loads(state_dict['thresholds'])
+        del state_dict['thresholds']
+        self._hyperband_scheduler(state_dict)
+
+    def _on_trial_result(self, trial_decision: str, trial: Trial, result: Dict) -> str:
+        if trial_decision != SchedulerDecision.CONTINUE:
```
mseeger, line 111: Comment above. For type == 'promotion', you will never see CONTINUE, so your modification would not do anything. The promotion type is implement in suggest.

It would, I think, be simpler to implement another RungSystem

syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -0,0 +1,151 @@
...
+        if 'points_to_evaluate' in kwargs:
+            points_to_evaluate += kwargs['points_to_evaluate']
+            points_to_evaluate = [dict(s) for s in set(frozenset(p.items()) for p in points_to_evaluate)]
+        kwargs['points_to_evaluate'] = points_to_evaluate
+
+        self._hyperband_scheduler = HyperbandScheduler(config_space, metric=metric, **kwargs)
+        self._num_init_configs = len(points_to_evaluate)
+        self._thresholds = dict()  # thresholds at different resource levels that must be met
+
+    @staticmethod
```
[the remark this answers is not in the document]
wistuba: It does not require `self` and I reuse the function for unit testing. Wouldn't be a problem to change but is there a benefit?

syne_tune/optimizer/schedulers/transfer_learning/rush.py
wistuba: It removes duplicates and preserves ordering.

syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -0,0 +1,151 @@
...
+        self._hyperband_scheduler.on_trial_remove(trial)
+
+    def on_trial_complete(self, trial: Trial, result: Dict) -> None:
+        self._hyperband_scheduler.on_trial_complete(trial, result)
+
+    def suggest(self, trial_id: int) -> Optional[TrialSuggestion]:
+        return self._hyperband_scheduler.suggest(trial_id)
+
+    def state_dict(self) -> Dict:
+        record = self._hyperband_scheduler.state_dict()
```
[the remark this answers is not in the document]
wistuba: Let's remove them if they are not used. I spotted them in HyperbandScheduler and thought they are important.

syne_tune/optimizer/schedulers/transfer_learning/rush.py
wistuba: I will have a look into the RungSystem classes to get an idea what this is about.

geoalgo: If we can implement a `RungSystem`, then its great. I would also be fine shipping the current version (asserting the promotion type is stop) and adding the rung system in another PR. @mseeger what do you think?

mseeger: That is also a possibility, it depends on Martin. I'd say the code would also be more clean if the RungSystem was implemented. Also, P-ASHA, the other contribution from Giovanni's team, did that.

geoalgo: Ok awesome. I will then give a ship-it then and Martin can decide if he prefers to merge this PR (and open a new one after checking if the rung system can be extended) or to do it all at once with this PR.

geoalgo (approved):

syne_tune/optimizer/schedulers/transfer_learning/rush.py
wistuba: How do we access `transfer_learning_evaluations` in case we want to implement a new RungSystem? It would require us to add a new argument to HyperbandScheduler, right? @mseeger @geoalgo

mseeger: This is how P-ASHA was done, they used another value for the type argument, and then added some extra arguments. But you already have the info of transfer_learning_evaluations via points_to_evaluate, so the additional args can be simple (for example: how many of the points_to_evaluate come from transfer).

If you think this is not clean, subclassing HyperbandScheduler is another option. But I am in general not such a huge fan of making a subclass, if all I am doing is change a component of a class. Essentially, RUSH is just making a small change to the stop/continue rule, everything else stays the same.

syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -112,7 +112,8 @@ def __setstate__(self, state):
         self._thresholds = pickle.loads(state['thresholds'])
 
     def _on_trial_result(self, trial_decision: str, trial: Trial, result: Dict) -> str:
-        if trial_decision == SchedulerDecision.STOP:
+        if trial_decision == SchedulerDecision.STOP or \
+                trial_decision == SchedulerDecision.PAUSE and self._hyperband_scheduler.scheduler_type == 'stopping':
```
mseeger, line 116: This does not fix the issues.

In type='promotion', this happens:
- A trial is always paused at the next milestone (if rung levels are 1, 3, 9, ...), a trial promoted from 1 has milestone 3, so the next rung level)
- The interesting decision is which trials get promoted, and that happens in _suggest

I bet that RUSH needs that decision to be influenced.

geoalgo: @wistuba : Perhaps we can only support "stop" type for now?

wistuba: > This does not fix the issues.
> 
> In type='promotion', this happens:
> 
>     * A trial is always paused at the next milestone (if rung levels are 1, 3, 9, ...), a trial promoted from 1 has milestone 3, so the next rung level)
> 
>     * The interesting decision is which trials get promoted, and that happens in _suggest
> 
> 
> I bet that RUSH needs that decision to be influenced.

The new piece of code is actually L133. At this point we identify for scheduler type 'promotion' that a milestone is reached and decide whether to stop or not. Does that not work?

wistuba: test_given_trial_not_meeting_threshold_return_stop shows that we correctly chose STOP rather than PAUSE. Not sure whether this is sufficient to make sure the trial is never suggested again.

syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -0,0 +1,159 @@
...
+        return self._on_trial_result(trial_decision, trial, result)
+
+    def on_trial_remove(self, trial: Trial) -> None:
+        self._hyperband_scheduler.on_trial_remove(trial)
+
+    def on_trial_complete(self, trial: Trial, result: Dict) -> None:
+        self._hyperband_scheduler.on_trial_complete(trial, result)
+
+    def suggest(self, trial_id: int) -> Optional[TrialSuggestion]:
+        return self._hyperband_scheduler.suggest(trial_id)
```
mseeger, line 100: The fact that you simply call suggest here, shows that something is missing. In type='promotion', the decision is made in suggest (whether to promote a paused trial or not).

It would be really hard to hack this post-hoc. Say normal ASHA decides to promote a trial here, but then RUSH says: no. Then, you'd have to go back to ASHA and ask "would you want to promote another trial, but not that one here?". This is not supported.

And if RUSH would block all possible ASHA promotions, then you'd have to call the searcher to suggest a new config.

This is too complex. Better implement the RungSystem.

syne_tune/optimizer/schedulers/transfer_learning/rush.py
mseeger: It would be OK with me to only support stopping, but I am not sure this is so good practice in the long run. How to communicate this to a user? Also, I suppose the overall task of "RUSH in Syne Tune" would not be done.

geoalgo: I cant comment on the above question of Martin whether this behavior will be correct to support promotion.

On the other point, from my side it is really fine to support only stopping type. I would still mark it as done, the initial paper does not consider at all the async case, having one of those mode is fine in my view given that they dont make such a big impact in final results.

wistuba:
I will open a new PR with new rung systems instead.

PR #150, opened 2022-02-05
Rewrite sync hb

mseeger:
*Issue #, if available:*
Total rewrite of synchronous HB. The previous version was for batch decisions, while this here is an asynchronous scheduler, but the decision-making (which trials to promote) is synchronized. This works quite a bit better than the old version, and provides a much fairer comparison with (synchronous) BOHB. The code is also quite a bit simpler.

*Description of changes:*


benchmarking/cli/launch_hpo.py
```diff
@@ -372,7 +372,7 @@
             metadata=metadata,
             tuner_name=tuner_name,
             max_failures=params['max_failures'],
-            asynchronous_scheduling=not params['synchronous'],
+            asynchronous_scheduling=True,
```
aaronkl, line 375: Do we still need this argument?

aaronkl (approved):

benchmarking/cli/launch_hpo.py
mseeger: You are right, it is just the default

PR #151, opened 2022-02-05
Remove state_dict, load_state_dict from TrialScheduler

mseeger:
*Issue #, if available:*
Removing unused methods

*Description of changes:*

### DIFF
--- a/syne_tune/optimizer/schedulers/hyperband.py
+++ b/syne_tune/optimizer/schedulers/hyperband.py
@@ -11,7 +11,6 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 from typing import Dict, Optional, List
-import pickle
 import logging
 import numpy as np
 import os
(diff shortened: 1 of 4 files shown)


syne_tune/optimizer/schedulers/hyperband.py
```diff
@@ -832,28 +831,6 @@ def _get_paused_trials(self) -> Dict:
             k: {k2: v[k2] for k2 in rem_keys}
             for k, v in self._active_trials.items() if not v['running']}
 
-    def state_dict(self) -> Dict:
```
geoalgo, line 835: Great, thanks for doing this. Note that you could still have custom (de)-serializer by implementing `__getstate__` and `__setstate__` when needed (as we do in tabulated benchmark to avoid saving uneeded tabulated data https://github.com/awslabs/syne-tune/blob/main/benchmarking/blackbox_repository/tabulated_benchmark.py#L219).

geoalgo (approved):

PR #152, opened 2022-02-08
Small fix

mseeger:
*Issue #, if available:*

*Description of changes:*


hfurkanbozkurt (approved):

geoalgo:
Hi @mseeger , can you confirm that this would not fail by default? (since debug_log is switched on by default?)

mseeger:
Hello David,
you are right. But the thing is, right now all our schedulers do not use get_batch_configs, and the default debug_log=True is only when you create searchers via these schedulers.

But I can add a comment to the get_batch_configs method, or make the exception message clearer.

PR #154, opened 2022-02-09
Update icml2020_import.py

talesa:
*Description of changes:*
`sp.loguniform` no longer accepts parameter `base`


geoalgo (approved):

geoalgo:
Thanks!

mseeger:
Yeah, that's one of these things in Ray Tune. They have this base parameter, but do not realize that all loguniform distributions are exactly the same, no matter what base is. I did not remove it for compatibility with them.

talesa:
The code I edited in this PR uses `def loguniform(lower: float, upper: float)`
https://github.com/awslabs/syne-tune/blob/5f15abb8bd8cf6c7eeed725f6afc37dcfb3ac019/syne_tune/search_space.py#L703-L714

PR #157, opened 2022-02-10
Allow BlackboxOffline _objective_function to return all fidelities when fidelity=None

talesa:
*Description of changes:*
Allow BlackboxOffline _objective_function to return all fidelities when fidelity=None, since that's the API assumed by Blackbox.

### DIFF
--- a/benchmarking/blackbox_repository/blackbox_offline.py
+++ b/benchmarking/blackbox_repository/blackbox_offline.py
@@ -84,12 +84,9 @@ class BlackboxOffline(Blackbox):
         key_dict = configuration
         if self.seed_col is not None:
             key_dict[self.seed_col] = seed
-        if fidelity is not None:
-            if self.fidelity_space is not None:
-                key_dict.update(fidelity)
-            else:
-                raise ValueError("fidelity is not None and self.fidelity_space is None.")
-        if fidelity is None and self.fidelity_space is not None:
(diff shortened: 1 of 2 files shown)


benchmarking/blackbox_repository/blackbox_offline.py
```diff
@@ -82,18 +82,29 @@ def _objective_function(
...
             raise ValueError(
                 f"the hyperparameter {configuration} is not present in available evaluations. Use `add_surrogate(blackbox)` if"
                 f" you want to add interpolation or a surrogate model that support querying any configuration."
             )
+        if fidelity is not None:
+            return output.iloc[0].to_dict()
+        else:
+            # TODO select only the fidelity values in the self.fidelity_space, since it might be the case there are more
+            #  values in the dataframe. Then the output tensor has larger number of elements than expected num_fidelities.
+            return output
```
talesa, line 107: At the moment, for `fidelity=None` it's returning a Pandas DataFrame with shape (num_fidelities, num_objectives) rather than a numpy tensor. Should we change it to output a tensor?

talesa: Changed to output a numpy tensor.

geoalgo: Thanks, makes sense.

geoalgo (approved):

PR #159, opened 2022-02-11
Make sure BB surrogates work with serialization, and change config_sp…

mseeger:
…ace of nashpobench (log spacing)

*Issue #, if available:*
Ensures that surrogates in BBs are not serialized as fitted estimators, but either just as names or as unfitted estimators. This makes sure this works with RemoteLauncher (I had issues with different scikit-learn versions), and the dill files stored at the end are of reasonable size (they were so large before that we could not run many experiments).

I also changed the type for some HPs from randint to lograndint. The reason is that their grid is logarithmic, so I feel this points to logarithmic scaling as well: lograndint is closer to choice over the grid than randint.

*Description of changes:*

### DIFF
--- a/benchmarking/blackbox_repository/blackbox_surrogate.py
+++ b/benchmarking/blackbox_repository/blackbox_surrogate.py
@@ -30,7 +30,7 @@ class BlackboxSurrogate(Blackbox):
             configuration_space: Dict,
             fidelity_space: Optional[Dict] = None,
             fidelity_values: Optional[np.array] = None,
-            surrogate=KNeighborsRegressor(n_neighbors=1),
+            surrogate=None,
             name: Optional[str] = None,
     ):
         """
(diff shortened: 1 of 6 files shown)


benchmarking/blackbox_repository/blackbox_surrogate.py
```diff
@@ -58,22 +80,27 @@ def __init__(
         # todo other types of assert with configuration_space, objective_names, ...
         self.X = X
         self.y = y
-        self.surrogate = surrogate
-        self.fit_surrogate(surrogate)
+        if surrogate is None:
+            surrogate = 'nn-1'
+        self.surrogate = blackbox_surrogate_factory(surrogate)
```
geoalgo, line 85: This will break user code who use a surrogate not covered by the factory (I currently have a few cases that are not supported by the factory). It will also prevent the use from custom surrogate without touching the factory. 

I would propose to rather save only surrogate options if the size is an issue (I can make that change). It seems that the original issue comes from a sklearn version also, we can have a look together to fix it in the meantime.

geoalgo (changes requested):

benchmarking/blackbox_repository/blackbox_surrogate.py
mseeger: Hello David, this does not break previous code. If surrogate is not a string, it does the same as before. So you can keep doing what you did before. It just may fail.

To be honest (and without disrespect), but I feel I am wasting time with these serialization of complex objects issues. I told you before that I am sceptical for exactly this reason. It is also not good practice to configre options by complex code and just serialize it, and hope for the best. It just calls for trouble every time there is any change anywhere.

mseeger: And the solution with clone is doing what it's supposed to do, without having to dig into which estimators have what parameters.

mseeger: To be specific: If surrogate is not a string, the factory returns surrogate (it does not fail). So, if you pass your favourite sklearn estimator, or anything obeying its API, you are fine.

mseeger:
Continued in #215

PR #165, opened 2022-02-11
RushScheduler

wistuba:
*Description of changes:*
Adds RUSHScheduler for the stopping rule only. Seeking feedback on new implementation before starting with the promotion rung system.

Two ways this currently works:
1) Using HyperbandScheduler with type='rush_stopping': if points_to_evaluate is given, use those. otherwise it still works but gives an info to the user that it is not really using RUSH
2) Using RUSHScheduler: this will work with transfer_learning_evaluations. In this case, points_to_evaluate are automatically estimated and forwarded to HyperbandScheduler. Then all other logic happens in the rung system.


syne_tune/optimizer/schedulers/transfer_learning/__init__.py
```diff
@@ -46,13 +69,15 @@ def __init__(
...
             metric_names: List[str],
+            **kwargs
     ):
         """
         A mixin that adds basic functionality for using offline evaluations.
         :param config_space: configuration space to be sampled from
         :param transfer_learning_evaluations: dictionary from task name to offline evaluations.
         :param metric_names: name of the metric to be optimized.
         """
+        super().__init__(config_space=config_space, **kwargs)
```
mseeger, line 80: What is this calling? There is no superclass.

tst/schedulers/transfer_learning/test_rush.py
```diff
@@ -0,0 +1,197 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mseeger, line 1: Nice tests!

syne_tune/optimizer/schedulers/hyperband_rush.py
```diff
@@ -0,0 +1,65 @@
...
+        self._num_points_to_evaluate = num_points_to_evaluate
+        self._thresholds = dict()  # thresholds at different resource levels that must be met
+
+    def task_continues(self, trial_id: str, mode: str, cutoff: float, metric_value: float, resource: int) -> bool:
+        task_continues = super().task_continues(trial_id=trial_id, mode=mode, cutoff=cutoff, metric_value=metric_value,
+                                                resource=resource)
+        return self._task_continues(task_continues, trial_id, metric_value, resource)
+
+    def _task_continues(self, task_continues: bool, trial_id: str, metric_value: float, resource: int) -> bool:
+        if not task_continues:
```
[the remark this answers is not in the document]
wistuba: based on the paper pseudo code, it should work like this but I can check with giovanni if he had this in mind.

syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -0,0 +1,65 @@
...
+from syne_tune.optimizer.schedulers.transfer_learning import TransferLearningTaskEvaluations, TransferLearningMixin
+
+
+class RUSHScheduler(TransferLearningMixin, HyperbandScheduler):
+    def __init__(
+            self,
+            config_space: Dict,
+            transfer_learning_evaluations: Dict[str, TransferLearningTaskEvaluations],
+            metric: str,
+            points_to_evaluate: Optional[List[Dict]] = None,
```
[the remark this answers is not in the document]
wistuba: yes, if RUSHScheduler is called directly, points_to_evaluate considered are points_to_evaluate passed to RUSH + top k configs. However, if someone calls HyperbandScheduler with type='rush_stopping', only points_to_evaluate passed to HyperbandScheduler are considered.

syne_tune/optimizer/schedulers/transfer_learning/__init__.py
wistuba: It will call the __init__ of any other extended class. In case of RUSHScheduler, this is the line that calls __init__ of HyperbandScheduler.

syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -0,0 +1,65 @@
...
+        :param config_space: configuration space for trial evaluation function.
+        :param metric: objective name to optimize, must be present in transfer learning evaluations.
+        :param transfer_learning_evaluations: dictionary from task name to offline evaluations.
+        :param points_to_evaluate: when points_to_evaluate is not None, the provided configurations are evaluated first
+        in addition to top performing configurations from other tasks and also serve to preemptively prune
+        underperforming configurations
+        :param num_hyperparameters_per_task: the number of top hyperparameter configurations to consider per task.
+        """
+        self._metric_names = [metric]
+        kwargs['type'] = 'rush_stopping'
```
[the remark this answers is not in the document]
wistuba: Yes, eventually it will have a type that allows for switching between those types. At the moment, there is only one so I've fixed it. I've planned to change as soon as the other one is supported. It's a bit ugly atm, I agree.

tst/schedulers/transfer_learning/test_rush.py
```diff
@@ -0,0 +1,197 @@
...
+        'b': randint(-100, 100)
+    }
+
+
+@pytest.fixture()
+def scheduler(metadata, config_space, request):
+    return RUSHScheduler(config_space=config_space,
+                         metric='loss',
+                         max_t=10,
+                         type=request.param,
```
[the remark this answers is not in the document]
wistuba: Yes, I've added this param as well as tests for 'promotion' already. I wanted to avoid rewriting all tests later given that I already know that these tests will come. But those tests don't test for promotion yet.

syne_tune/optimizer/schedulers/hyperband.py
```diff
@@ -373,6 +376,7 @@ def __init__(self, config_space, **kwargs):
         self.terminator = HyperbandBracketManager(
             scheduler_type, self._resource_attr, self.metric, self.mode,
             self.max_t, rung_levels, brackets, rung_system_per_bracket,
+            num_points_to_evaluate=len(kwargs.get('points_to_evaluate', list())),
```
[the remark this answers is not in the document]
wistuba: Are you suggesting that we need to support the case that someone wants to provide initial configs which aren't be used as threshold candidates for RUSH? I thought it might be very convenient to simply pass custom points_to_eval instead of using any metadata.

syne_tune/optimizer/schedulers/transfer_learning/__init__.py
mseeger: Maybe you can explain this to me in our next meeting. I am coming from C++, where this multiple inheritance is always quite complicated. I think I need this mixin technique as well.

syne_tune/optimizer/schedulers/hyperband_promotion.py
```diff
@@ -86,20 +85,18 @@ def _find_promotable_config(self, recorded, prom_quant, trial_id=None):
...
+            # Best id among trials paused at this rung (i.e., not yet promoted)
+            id, val = max(
+                ((k, v[0]) for k, v in recorded.items() if self._is_feasible_config(k, v[0], not v[1], resource)),
+                key=lambda x: sign * x[1],
+                default=(None, 0.0))
             if id is not None and sign * (val - cutoff) >= 0:
                 ret_id = id
         return ret_id
 
+    def _is_feasible_config(self, trial_id, metric_val, is_paused, resource):
```
mseeger, line 97: Elegant! Maybe I'd call this `_is_promotable_config`

syne_tune/optimizer/schedulers/transfer_learning/rush.py
```diff
@@ -0,0 +1,79 @@
...
+class RUSHScheduler(TransferLearningMixin, HyperbandScheduler):
+    def __init__(
+            self,
+            config_space: Dict,
+            transfer_learning_evaluations: Dict[str, TransferLearningTaskEvaluations],
+            metric: str,
+            type: str = 'stopping',
+            points_to_evaluate: Optional[List[Dict]] = None,
+            custom_rush_points: Optional[List[Dict]] = None,
+            num_hyperparameters_per_task: Optional[int] = 1,
```
mseeger, line 28: Is this really Optional[int] if the default is 1?

mseeger (approved):

mseeger (approved):

PR #168, opened 2022-02-14
Sample from active space for warmstart for random search

iaroslav-ai:
*Description of changes:*

In case of warmstart (active search space specified) candidates from active search space should be sampled for random search.

This also adds few instructions on how to develop this library.

PR #169, opened 2022-02-14
Fix bug pointed out by Iaroslav (which affects transfer HPO)

mseeger:
*Issue #, if available:*
#168 

*Description of changes:*
Fixes the issue that random configs are sampled from the wrong config space


syne_tune/optimizer/schedulers/searchers/gp_fifo_searcher.py
```diff
@@ -380,7 +380,7 @@ def _assign_random_searcher(self, random_seed=None):
             if random_seed is None:
                 random_seed = self.random_state.randint(0, 2 ** 32)
             self._random_searcher = RandomSearcher(
-                self.hp_ranges.config_space,
+                self.hp_ranges.config_space_for_sampling,
```
iaroslav-ai, line 383: Could you also import the unit test I added in the other PR - to make sure that this will work as expected going forward?

PR #170, opened 2022-02-14
Fix configuration space for nashpobench benchmark so that interpolati…

mseeger:
…on (surrogate) is not required

*Issue #, if available:*
The code in main right now does not work. It uses a surrogate, which needs fixes as proposed in #159 

*Description of changes:*
This is a small part of #159 . It sets the config space for nashpobench such that a surrogate is not needed. This is still not the best for BO, because one parameter remains categorical. But it is better than having all of them categorical.

### DIFF
--- a/benchmarking/definitions/definition_nashpobench.py
+++ b/benchmarking/definitions/definition_nashpobench.py
@@ -10,31 +10,30 @@
 # on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
+from syne_tune.search_space import choice, logfinrange, finrange
 
-from sklearn.ensemble import RandomForestRegressor
-
-from syne_tune.search_space import choice, randint, uniform, loguniform
-
-from benchmarking.blackbox_repository.conversion_scripts.scripts.fcnet_import import \
-    METRIC_ELAPSED_TIME, METRIC_VALID_LOSS, RESOURCE_ATTR, BLACKBOX_NAME
+from benchmarking.blackbox_repository.conversion_scripts.scripts.fcnet_import \
+    import METRIC_ELAPSED_TIME, METRIC_VALID_LOSS, RESOURCE_ATTR, BLACKBOX_NAME
 
 
+# This configuration space allows to use the tabulated blackbox without any
+# interpolation (surrogate), meaning that all numerical HPs have finite ranges.
+# Note that `hp_init_lr` remains categorical and will be 1-hot encoded. It
+# would be better for BO to use a surrogate.
 _config_space = {
     "hp_activation_fn_1": choice(["tanh", "relu"]),
     "hp_activation_fn_2": choice(["tanh", "relu"]),
-    "hp_batch_size": randint(8, 64),
-    "hp_dropout_1": uniform(0.0, 0.6),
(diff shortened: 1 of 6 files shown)


syne_tune/search_space.py
```diff
@@ -595,10 +595,11 @@ class FiniteRange(Domain):
     """
     Represents a finite range `[lower, ..., upper]` with `size` values
     equally spaced in linear or log domain.
+    If `is_int`, the value type is int (rounding after the transform).
 
     """
     def __init__(self, lower: float, upper: float, size: int,
-                 log_scale: bool = False):
+                 log_scale: bool = False, is_int: bool = False):
```
geoalgo, line 602: `cast_int` would be a better name since we cast ultimately to differences that may not be too small.

benchmarking/definitions/definition_nashpobench.py
```diff
@@ -71,7 +67,6 @@ def nashpobench_benchmark(params):
         'cost_model': None,
         'supports_simulated': True,
         'blackbox_name': BLACKBOX_NAME,
-        'surrogate': RandomForestRegressor(max_samples=0.005, bootstrap=True),
+        'surrogate': None,
```
geoalgo, line 70: Could you revert this line? (it will not work currently)

benchmarking/definitions/definition_nashpobench.py
```diff
@@ -10,31 +10,30 @@
...
+# Note that `hp_init_lr` remains categorical and will be 1-hot encoded. It
+# would be better for BO to use a surrogate.
 _config_space = {
     "hp_activation_fn_1": choice(["tanh", "relu"]),
     "hp_activation_fn_2": choice(["tanh", "relu"]),
-    "hp_batch_size": randint(8, 64),
-    "hp_dropout_1": uniform(0.0, 0.6),
-    "hp_dropout_2": uniform(0.0, 0.6),
-    "hp_init_lr": loguniform(0.0005, 0.1),
+    "hp_batch_size": logfinrange(8, 64, 4, is_int=True),
```
geoalgo, line 26: If we update this one, we should also update `fcnet_recipe.py` for consistency with the same values.

geoalgo (commented):
LGTM, I only have small comments.

benchmarking/definitions/definition_nashpobench.py
mseeger: What do you mean? This idea here is not to have a surrogate. I removed the whole line.

benchmarking/definitions/definition_nashpobench.py
mseeger: Good idea

benchmarking/definitions/definition_nashpobench.py
mseeger: BTW: When I am accessing an entry in a blackbox table, where a HP has float as type, do I need to hit the float value in the table exactly? In principle, finrange and logfinrange with value type float could not hit the intended values exactly. The cast_int is a mechanism to keep things safe for int values, which is good.

geoalgo: > What do you mean? This idea here is not to have a surrogate. I removed the whole line.

Indeed you are right.

Regarding your question, yes the value must be exact otherwise a surrogate model must be used.

geoalgo (approved):

PR #174, opened 2022-02-17
API Renaming

geoalgo:
*Description of changes:* api renaming and put core classes in root init to simplify imports.

* `Backend` -> `TrialBackend`
* `backend` parameter of tuner is now called `trial_backend` to reflect this
* `SagemakerBackend` -> `SageMakerBackend`
* also fix some examples (hence a larger diff for them) where imports were not updated after our move to utils in benchmarking/

I did not rename `LocalBackend` to `LocalTrialBackend` as the naming in the tuner should make it now more clear that those are trials backend.


docs/schedulers.md
```diff
@@ -41,43 +42,43 @@ from syne_tune.optimizer.schedulers.fifo import FIFOScheduler
...
-        metric=benchmark['metric'])
-
-    tuner = Tuner(
-        backend=backend,
-        scheduler=scheduler,
-        stop_criterion=StoppingCriterion(max_wallclock_time=120),
-        n_workers=n_workers)
-
-    tuner.run()
+  logging.getLogger().setLevel(logging.DEBUG)
```
aaronkl, line 50: is there a reason why you change the indent?

aaronkl (approved):

docs/schedulers.md
geoalgo: space versus tab in IDE I guess, we should use black to avoid this.

PR #176, opened 2022-02-17
Add additional baselines

mseeger:
*Issue #, if available:*

*Description of changes:*
These short names would be useful for the paper, since I want to make a table for the schedulers


syne_tune/optimizer/baselines.py
```diff
@@ -28,50 +31,147 @@ def __init__(self, config_space: Dict, metric: str, **kwargs):
 
 
 class ASHA(HyperbandScheduler):
-    def __init__(self, config_space: Dict, metric: str, resource_attr: str, max_t: int, **kwargs):
+    def __init__(
+            self, config_space: Dict, metric: str, resource_attr: str,
+            **kwargs):
+        """
+        One of `max_t`, `max_resource_attr` needs to be in `kwargs`. For
```
mseeger, line 38: If max_resource_attr is passed, the method cannot only infer max_t (equal to config_space[max_resource_attr]), but it can also change this value when starting a trial. This is useful, because the trial will then stop on its own, instead of having to be stopped.

aaronkl (approved):

PR #180, opened 2022-02-22
Reporter in root __init__.py to ease user imports

geoalgo:
*Description of changes:* This changes adds `Reporter` to root init allow easier user import.


mseeger (approved):
Just to confirm: The old way of import would also still work, right?

Should also update the docs

geoalgo:
Yes, it would.

PR #182, opened 2022-02-23
Return training job name from remote launcher

hfurkanbozkurt:
*Issue #, if available:*

*Description of changes:* This is useful to programmatically query the status of remote tuning jobs

### DIFF
--- a/syne_tune/remote/remote_launcher.py
+++ b/syne_tune/remote/remote_launcher.py
@@ -102,7 +102,7 @@ class RemoteLauncher:
     def run(
             self,
             wait: bool = True,
-    ):
+    ) -> str:
         """
         :param wait: Whether the call should wait until the job completes (default: True). If False the call returns
         once the tuning job is scheduled on Sagemaker.
@@ -114,7 +114,7 @@ class RemoteLauncher:
             # launching in this is needed to send a default configuration on the tuning loop running on Sagemaker
             # todo restore the env variable if present to avoid a side effect
             os.environ['AWS_DEFAULT_REGION'] = 'us-west-2'
-        self.launch_tuning_job_on_sagemaker(wait=wait)
+        return self.launch_tuning_job_on_sagemaker(wait=wait)
 
     def prepare_upload(self):
         """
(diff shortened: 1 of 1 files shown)


geoalgo (commented):
Given that we can set the training job name as parameter, is there a particular reason to return as output?

talesa:
The name passed as an argument doesn't include the unique datetime stamp part of the SM Training Job name, so it is useful to return it, I think.

geoalgo:
> The name passed as an argument doesn't include the unique datetime stamp part of the SM Training Job name, so it is useful to return it, I think.

The name is only generated if you do not pass a name as input (by passing `job_name` to fit). 

If this is really needed, we can add that.

But the general thinking is to try to keep the API of the remote launcher as close as possible from a SM Estimator since every gap can be a potential surprise for a SM user and its also extra logic to maintain (and SM estimator dont return training job name https://github.com/aws/sagemaker-python-sdk/blob/dev/src/sagemaker/estimator.py#L897)

talesa:
> The name is only generated if you do not pass a name as input (by passing job_name to fit).

I don't think that's true: for `RemoteLauncher(tuner_name='speed-bs-it-nw-new', ...)` I get 
image
where the bottom one is the name of the RemoteLauncher job.

@hfurkanbozkurt could you remove this line while you're at it? `self.base_job_name` is not used anywhere.
https://github.com/awslabs/syne-tune/blob/main/syne_tune/remote/remote_launcher.py#L75

geoalgo:
Your example is probably incomplete (you probably meant to pass tuner_name in Tuner?)

If you do the following, then the training-job will be exactly the tuner name.

```
    tuner = Tuner(
        trial_backend=trial_backend,
        scheduler=scheduler,
        n_workers=n_workers,
        tuner_name="height-tuning",
        stop_criterion=StoppingCriterion(max_wallclock_time=600),
    )
    tuner = RemoteLauncher(
        tuner=tuner,
        instance_type='ml.m5.large',
    )
    tuner.run(wait=False)
```

For instance, when running this example the tuner name was "height-tuning-2022-02-24-14-30-24-698" and the training job name of the remote is then also exactly "height-tuning-2022-02-24-14-30-24-698".

The remote launcher always use the tuner name for the training job name as can be seen in this line: 
https://github.com/awslabs/syne-tune/blob/main/syne_tune/remote/remote_launcher.py#L235

As it is I do not think we should return this information since it is already available. 

On your other comment, I agree that the line you refered to is obsolete and should be removed.

geoalgo:
> So you're saying this is unnecessary because in your example we could  access it via tuner.tuner.name?

Yes exactly, using tuner for remote launcher is a bit ugly, if it was named `remote_launcher` then you would have it by just calling `tuner.name`.

PR #186, opened 2022-02-24
add pasha demo

ondrejbohdal:
*Description of changes:*
* Add a Jupyter notebook with NASBench201 demo for PASHA algorithm - within the examples directory
* Enable combining PASHA with Bayesian Optimization searcher
* Fix typos in hyperband.py and baselines.py files that relate to PASHA

I've tested the changes locally by running the Jupyter notebook.

### DIFF
--- a/syne_tune/optimizer/baselines.py
+++ b/syne_tune/optimizer/baselines.py
@@ -77,7 +77,7 @@ class MOBSTER(HyperbandScheduler):
 class PASHA(HyperbandScheduler):
     def __init__(
             self, config_space: Dict, metric: str, resource_attr: str,
-            **kwargs):
+            rung_system_kwargs: Dict, **kwargs):
         """
         One of `max_t`, `max_resource_attr` needs to be in `kwargs`. The
         latter is more useful, see also :class:`HyperbandScheduler`.
@@ -92,7 +92,7 @@ class PASHA(HyperbandScheduler):
             searcher="random",
             resource_attr=resource_attr,
             type='pasha',
-            ranking_criterion='soft_ranking',
+            rung_system_kwargs=rung_system_kwargs,
             **kwargs,
         )
 
--- a/examples/pasha_nasbench201_demo.ipynb
+++ b/examples/pasha_nasbench201_demo.ipynb
@@ -1,0 +1,635 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "e400615e-3e2b-4410-9f56-03461c6900bc",
+   "metadata": {},
+   "source": [
+    "# PASHA: Efficient HPO with Progressive Resource Allocation"
+   ]
+  },
+  {
+   "cell_type": "markdown",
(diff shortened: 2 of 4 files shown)


geoalgo:
Thanks a lot for the fixes, that we can merge easily. 

For the notebook, we do not have any notebook currently because we would need to set extra tools to support them (they dont play with with versioning and are harder to test). Would it be possible to put the notebook in a branch? (You can still refer it in the doc if you want this way)

610v4nn1:
Would it help to convert the notebook in an example script as the ones available in the /examples folder?

geoalgo:
Yes certainly, script are fine.

ondrejbohdal:
I've now converted the Jupyter notebook into a script and removed the Jupyter notebook. I've tested the script too.

syne_tune/optimizer/baselines.py
```diff
@@ -77,7 +77,7 @@ def __init__(
 class PASHA(HyperbandScheduler):
     def __init__(
             self, config_space: Dict, metric: str, resource_attr: str,
-            **kwargs):
+            rung_system_kwargs: Dict, **kwargs):
```
mseeger, line 80: The idea here was to only expose really important args

examples/launch_pasha_nasbench201_demo.py
```diff
@@ -0,0 +1,424 @@
+# PASHA: Efficient HPO with Progressive Resource Allocation
```
mseeger, line 1: Hello Ondre,

thanks a lot for this great example.

My feeling is it is a bit too long for what we normally have in the examples folder. Also, it would be a bit wasted there, because it is almost more like a tutorial.

If others agree, and you'd be up for it, one idea would be to have this in docs/tutorials/pasha.

It would include this complete script here, and then a markdown with code snippets, where the user can learn about PASHA.

What do you think?

geoalgo: I was thinking to suggest this also as it is indeed a bit long for examples. Either way works for me.

ondrejbohdal: I think it is a good idea to create a tutorial out of it - originally it was a Jupyter notebook, which has a similar style as a tutorial.

syne_tune/optimizer/baselines.py
ondrejbohdal: That makes sense - I've now changed it so that rung_system_kwargs is not exposed here - it can be accessed via kwargs, so it does not need to be there.

geoalgo (dismissed):

examples/launch_pasha_nasbench201_demo.py
mseeger: Hi Ondre, could you please rather move it to docs/tutorials/pasha/? The folder basics is for a specific tutorial

ondrejbohdal: sure, I've just moved it there

mseeger (approved):

PR #187, opened 2022-02-25
Config space

mseeger:
*Issue #, if available:*

*Description of changes:*
Maps 'configspace' to 'config_space', and 'search_space' to 'config_space'.
The latter is a breaking API change, so I kept search_space common functions in, which log warnings and call config_space functions.

### DIFF
--- a/syne_tune/search_space.py
+++ b/syne_tune/search_space.py
@@ -1,0 +1,79 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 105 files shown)


syne_tune/search_space.py
```diff
@@ -14,956 +14,66 @@
...
-
-def sample_from(func: Callable[[Dict], Any]):
-    """Specify that tune should sample configuration values from this function.
-
-    Arguments:
-        func: An callable function to draw a sample from.
-    """
-    return Function(func)
+def _deprecated_warning(name: str):
+    logger.warning(
```
geoalgo, line 37: When will we remove this deprecation? We should include when support will be removed also I think, it could be in the next version for instance.

geoalgo (commented):
LGTM thanks for doing this. One small comment I believe we should be explicit and indicate when the support for deprecated function will be removed.

aaronkl (dismissed):

geoalgo (approved):

PR #189, opened 2022-02-26
Simplify imports for backend and schedulers

mseeger:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/examples/launch_height_standalone_scheduler.py
+++ b/examples/launch_height_standalone_scheduler.py
@@ -23,9 +23,8 @@ from syne_tune.backend import LocalBackend
 from syne_tune.backend.trial_status import Trial
 from syne_tune.optimizer.scheduler import TrialScheduler, \
     SchedulerDecision, TrialSuggestion
-from syne_tune import Tuner
+from syne_tune import Tuner, StoppingCriterion
 from syne_tune.config_space import randint
-from syne_tune import StoppingCriterion
 
 
 class SimpleScheduler(TrialScheduler):
--- a/syne_tune/backend/__init__.py
+++ b/syne_tune/backend/__init__.py
@@ -10,8 +10,10 @@
 # on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
-__all__ = ['LocalBackend', 'SageMakerBackend']
+
+__all__ = ['LocalBackend',
+           'SageMakerBackend']
 
 from syne_tune.backend.local_backend import LocalBackend
-from syne_tune.backend.sagemaker_backend.sagemaker_backend import SageMakerBackend
-
+from syne_tune.backend.sagemaker_backend.sagemaker_backend import \
+    SageMakerBackend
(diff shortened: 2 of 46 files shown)


syne_tune/backend/__init__.py
```diff
@@ -10,8 +10,10 @@
 # on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
-__all__ = ['LocalBackend', 'SageMakerBackend']
 
-from syne_tune.backend.local_backend import LocalBackend
-from syne_tune.backend.sagemaker_backend.sagemaker_backend import SageMakerBackend
+__all__ = ['LocalBackend',
```
geoalgo, line 14: This is only changing code formatting? I am asking because the PR titles mentions "simplify imports for backend" but I do not understand what would be the change.

syne_tune/optimizer/schedulers/synchronous/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+__all__ = ['SynchronousHyperbandScheduler',
```
geoalgo, line 14: I am not sure we want to maintain a list of imports in the root of every sub folder for schedulers (async, sync, multiobj, etc). We have a list of method definitions in baselines, should we keep this one as the main entry point?

tst/test_constrained_bo.py
```diff
@@ -15,9 +15,9 @@
 
 from syne_tune.backend import LocalBackend
 from syne_tune.optimizer.schedulers.fifo import FIFOScheduler
-from syne_tune import Tuner
+from syne_tune.tuner import Tuner
 from syne_tune.config_space import uniform
-from syne_tune import StoppingCriterion
+from syne_tune.stopping_criterion import StoppingCriterion
```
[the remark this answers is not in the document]
mseeger: Uups, I'll revert this back

examples/launch_height_standalone_scheduler.py
```diff
@@ -23,9 +23,9 @@
 from syne_tune.backend.trial_status import Trial
 from syne_tune.optimizer.scheduler import TrialScheduler, \
     SchedulerDecision, TrialSuggestion
-from syne_tune import Tuner
+from syne_tune.tuner import Tuner
+from syne_tune.stopping_criterion import StoppingCriterion
```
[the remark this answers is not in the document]
mseeger: Will revert back

syne_tune/backend/__init__.py
mseeger: Nothing changes for local and sagemaker, but for simulated there is a simplification.

syne_tune/optimizer/schedulers/synchronous/__init__.py
mseeger: I'd prefer to have this as well. It is optional in a way for whoever adds something there.
But quite some users will still want to create things not using the baselines. And in fact, the baseline definitions should also be easy to understand, so they should use the simplified imports as well.

mseeger: If the backends get easier imports, the schedulers should get them as well, both are key concepts that can become many.
Some people will just not like to use baselines, and some scheduler combinations may not even be in there.

geoalgo (approved):

PR #192, opened 2022-03-01
Scalable Gaussian Process Transfer

wistuba:
Added another transfer learning baseline: Scalable Gaussian Process Transfer

Problems:
- Init points by default are 3, should be 0.
- If we can set it to 0, requires special treatment of that case since the standard GP without data won't work
- Not sure, standard scaling actually works too well for this method. I saw the GPs for the target task predict very extreme values. Originally, it used normalizing targets to [0, 1].


geoalgo (approved):
Really cool stuff, thanks a lot! Looking forward having it in mainline as well :-)

PR #197, opened 2022-03-02
Fix upper bound of benchmark

wistuba:
Upper bound of max units is 1024 (see: https://github.com/automl/LCBench)


wistuba:
```
from benchmarking.nursery.benchmark_kdd.benchmark_main import get_transfer_learning_evaluations
metadata = get_transfer_learning_evaluations('lcbench', 'APSFailure')
print(metadata['Amazon_employee_access'].hyperparameters['max_units'].max())
```
1024

PR #200, opened 2022-03-02
Blackbox surrogate dictionary, small fixes

geoalgo:
*Description of changes:* backport changes on blackboxes from kdd branch (I will open separate PR for CTS and additional tooling for experiment).
* add option to fit surrogate on subset
* standardize output before fitting surrogates
* make sure cumulative time is increasing in simulations
* change BlackboxRepositoryBackend surrogate parameters to string/dictionary and only fit estimator in a lazy fashion (as proposed in Matthias PR)

### DIFF
--- a/benchmarking/blackbox_repository/blackbox_surrogate.py
+++ b/benchmarking/blackbox_repository/blackbox_surrogate.py
@@ -30,7 +30,7 @@ class BlackboxSurrogate(Blackbox):
             configuration_space: Dict,
             fidelity_space: Optional[Dict] = None,
             fidelity_values: Optional[np.array] = None,
-            surrogate=KNeighborsRegressor(n_neighbors=1),
+            surrogate=None,
             max_fit_samples: Optional[int] = None,
             name: Optional[str] = None,
     ):
@@ -43,11 +43,14 @@ class BlackboxSurrogate(Blackbox):
         :param y: dataframe containing objectives values
         :param configuration_space:
         :param fidelity_space:
-        :param surrogate: the model that is fitted to predict objectives given any configuration.
+        :param surrogate: the model that is fitted to predict objectives given any configuration, default to
+        KNeighborsRegressor(n_neighbors=1).
         Possible examples: KNeighborsRegressor(n_neighbors=1), MLPRegressor() or any estimator obeying Scikit-learn API.
         The model is fit on top of pipeline that applies basic feature-processing to convert rows in X to vectors.
         We use the configuration_space hyperparameters types to deduce the types of columns in X (for instance
         CategoricalHyperparameter are one-hot encoded).
+        :param max_fit_samples: maximum number of samples to be fed to the surrogate estimator, if the more data points
+        than this number are passed, then they are subsampled without replacement.
         :param name:
         """
         super(BlackboxSurrogate, self).__init__(
@@ -57,7 +60,7 @@ class BlackboxSurrogate(Blackbox):
         )
         assert len(X) == len(y)
         # todo other types of assert with configuration_space, objective_names, ...
-        self.surrogate = surrogate
+        self.surrogate = surrogate if surrogate is not None else KNeighborsRegressor(n_neighbors=1)
         self.max_fit_samples = max_fit_samples
         self.fit_surrogate(X=X, y=y, surrogate=surrogate, max_samples=self.max_fit_samples)
         self.name = name
@@ -101,14 +104,14 @@ class BlackboxSurrogate(Blackbox):
             ('model', model)
         ])
 
-    def fit_surrogate(self, X, y, surrogate=KNeighborsRegressor(n_neighbors=1), max_samples: Optional[int] = None) -> Blackbox:
+    def fit_surrogate(self, X, y, surrogate=None, max_samples: Optional[int] = None) -> Blackbox:
         """
         Fits a surrogate model to a blackbox.
         :param surrogate: fits the model and apply the model transformation when evaluating a
         blackbox configuration. Possible example: KNeighborsRegressor(n_neighbors=1), MLPRegressor() or any estimator
         obeying Scikit-learn API.
         """
-        self.surrogate = surrogate
+        self.surrogate = surrogate if surrogate is not None else KNeighborsRegressor(n_neighbors=1)
 
         self.surrogate_pipeline = self.make_model_pipeline(
             configuration_space=self.configuration_space,
--- a/benchmarking/blackbox_repository/blackbox_surrogate.py
+++ b/benchmarking/blackbox_repository/blackbox_surrogate.py
@@ -31,6 +31,7 @@ class BlackboxSurrogate(Blackbox):
             fidelity_space: Optional[Dict] = None,
             fidelity_values: Optional[np.array] = None,
             surrogate=KNeighborsRegressor(n_neighbors=1),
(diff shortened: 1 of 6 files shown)


benchmarking/blackbox_repository/blackbox_surrogate.py
```diff
@@ -31,6 +31,7 @@ def __init__(
             fidelity_space: Optional[Dict] = None,
             fidelity_values: Optional[np.array] = None,
             surrogate=KNeighborsRegressor(n_neighbors=1),
+            max_fit_samples: Optional[int] = None,
```
mseeger, line 34: Can you document this?

BTW: I'd not downsample the BB data, these nonparametric methods can handle it, esp. KNN.

If this is about reducing the size of the dill file, let us do this in another way. I have a PR open, and maybe you have another idea.

But I view this BB data as really valuable, and the goal is interpolation, so you really do not want to subsample. One can subsample for things like classification, but interpolating a complex function like this, you really want to use all the data you have.

benchmarking/blackbox_repository/blackbox_surrogate.py
```diff
@@ -31,6 +31,7 @@ def __init__(
             fidelity_space: Optional[Dict] = None,
             fidelity_values: Optional[np.array] = None,
             surrogate=KNeighborsRegressor(n_neighbors=1),
```
mseeger, line 33: I'd prefer =None here, and then set the default inside. It is a better pattern.

benchmarking/blackbox_repository/blackbox_surrogate.py
```diff
@@ -106,15 +95,38 @@ def fit_surrogate(self, surrogate=KNeighborsRegressor(n_neighbors=1)) -> Blackbo
...
             ("features", FeatureUnion(features_union)),
-            ('model', surrogate)
+            ('standard scaler', StandardScaler(with_mean=False)),
+            ('model', model)
         ])
 
-        self.surrogate_pipeline.fit(
-            X=self.X,
-            y=self.y
+    def fit_surrogate(self, X, y, surrogate=KNeighborsRegressor(n_neighbors=1), max_samples: Optional[int] = None) -> Blackbox:
```
mseeger, line 104: Again, would prefer surrogate=None

benchmarking/blackbox_repository/blackbox_surrogate.py
```diff
@@ -106,15 +95,38 @@ def fit_surrogate(self, surrogate=KNeighborsRegressor(n_neighbors=1)) -> Blackbo
...
+        """
+        Fits a surrogate model to a blackbox.
+        :param surrogate: fits the model and apply the model transformation when evaluating a
+        blackbox configuration. Possible example: KNeighborsRegressor(n_neighbors=1), MLPRegressor() or any estimator
+        obeying Scikit-learn API.
+        """
+        self.surrogate = surrogate
```
mseeger, line 111: I'd expect something like if surrogate is None, use self.surrogate ?

benchmarking/blackbox_repository/blackbox_tabular.py
```diff
@@ -81,14 +81,19 @@ def __init__(
 
     def _objective_function(
             self,
-            configuration: Dict,
+            configuration: Union[Dict, int],
```
mseeger, line 84: This seems odd. Did you change this in the baseclass as well?

benchmarking/blackbox_repository/simulated_tabular_backend.py
```diff
@@ -156,9 +156,46 @@ def _run_job_and_collect_results(
                 result[self.elapsed_time_attr] -= elapsed_time_offset
         else:
             results = all_results
+
+        # Makes sure that time is monotonically increasing which may not be the case due to numerical errors or due to
+        # the use of a surrogate
+        for i in range(1, len(results)):
```
mseeger, line 162: Did you encounter this? I'd be more happy if we failed here, ideally returning something that would allow us to figure out what is going on.

benchmarking/blackbox_repository/simulated_tabular_backend.py
```diff
@@ -156,9 +156,46 @@ def _run_job_and_collect_results(
...
+            results[i][self.elapsed_time_attr] = max(
+                results[i][self.elapsed_time_attr],
+                results[i - 1][self.elapsed_time_attr] + 0.001
+            )
+
         return status, results
 
 
+def make_surrogate(surrogate: Optional[str] = None, surrogate_kwargs: Optional[Dict] = None):
```
mseeger, line 171: Ah OK, so this would get us around these issues I had?

benchmarking/blackbox_repository/simulated_tabular_backend.py
```diff
@@ -184,10 +222,13 @@ def __init__(
         :param max_resource_attr:
         :param dataset: Selects different versions of the blackbox
         :param surrogate: optionally, a model that is fitted to predict objectives given any configuration.
-        Possible examples: KNeighborsRegressor(n_neighbors=1), MLPRegressor() or any estimator obeying Scikit-learn API.
+        Possible examples: "KNeighborsRegressor" or "MLPRegressor" or "XGBRegressor" which would enable using
+        the corresponding Scikit-learn estimator.
         The model is fit on top of pipeline that applies basic feature-processing to convert hyperparameters
         rows in X to vectors. The configuration_space hyperparameters types are used to deduce the types of columns in
          X (for instance CategoricalHyperparameter are one-hot encoded).
+        :param surrogate_kwargs: arguments for the scikit-learn estimator, for instance {"n_neighbors": 1} can be used
```
mseeger, line 230: I like this a lot more than passing a sckikt.learn object

benchmarking/blackbox_repository/blackbox_surrogate.py
geoalgo: Added a comment. This is not about reducing the file of the dill file, we are only saving the surrogate name and its arguments, it is to reduce the potential very long runtime that happens if you have a large number of evaluations.

benchmarking/blackbox_repository/blackbox_surrogate.py
geoalgo: Updated.

benchmarking/blackbox_repository/blackbox_tabular.py
geoalgo: This is to be able to support a comparison with LCbench which access configuration by integer, not by dictionary.

benchmarking/blackbox_repository/simulated_tabular_backend.py
geoalgo: Yes in some cases, the runtime recorded is 0 which causes this error. If we would fail, we should probably fail rather at the creation of the tabular blackbox. I propose to postpone this to another PR given that it seems optional and it would be hard for me to merge otherwise.

benchmarking/blackbox_repository/simulated_tabular_backend.py
geoalgo: Yes.

benchmarking/blackbox_repository/blackbox_surrogate.py
geoalgo: That would not work with how the method is called.

mseeger (approved):

PR #204, opened 2022-03-04
Included definition for LCBench and improve handling of datasets

mseeger:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/benchmarking/definitions/definition_lcbench.py
+++ b/benchmarking/definitions/definition_lcbench.py
@@ -1,0 +1,105 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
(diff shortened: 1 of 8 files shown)


benchmarking/definitions/definition_lcbench.py
```diff
@@ -0,0 +1,105 @@
...
+    "shuttle",
+    "sylvine",
+    "vehicle",
+    "volkert",
+]
+
+
+# This configuration space requires a surrogate to be used. This is needed
+# anyway, since the blackbox is not densely evaluated on a grid
+_config_space = {
```
geoalgo, line 60: Could we import the config-space defined in lcbench rather to avoid the duplication?

geoalgo (commented):
LGTM, just have a small comment.

benchmarking/definitions/definition_lcbench.py
mseeger: OK.

mseeger: OK, done. I also added a fix

geoalgo (approved):

PR #205, opened 2022-03-04
Fix to domain for bounding-box (support having only a single element …

geoalgo:
…and provide support for restricting finite search spaces).

*Issue #, if available:*

*Description of changes:* pull changes from automl experiment branch that affects common components (this is the last one the other PR will be just new schedulers and experiment scripts)

* support finite range with only single elements
* support restricting finite range to smaller intervals
* support not loading tuner when accessing results (much faster)
* use metadata instead of inspecting the tuner for mode and metrics
* allow to load metadata only of experiments and filter by name (much faster than loading all experiments results)

### DIFF
--- a/syne_tune/config_space.py
+++ b/syne_tune/config_space.py
@@ -619,6 +619,10 @@ class FiniteRange(Domain):
             self._step_internal = (upper_internal - self._lower_internal) / (size - 1) if size > 1 else 0
         self._values = [self._map_from_int(x) for x in range(self.size)]
 
+    @property
+    def values(self):
+        return self._values
+
     def _map_from_int(self, x: int) -> Union[float, int]:
         y = x * self._step_internal + self._lower_internal
         if self.log_scale:
@@ -991,7 +995,7 @@ def restrict_domain(numerical_domain: Domain, lower: float, upper: float) -> Dom
         new_domain_dict['domain_kwargs']['upper'] = upper
         return from_dict(new_domain_dict)
     else:
-        values = numerical_domain._values
+        values = numerical_domain.values
         assert lower < max(numerical_domain._values)
         assert upper > min(numerical_domain._values)
         i = 0
--- a/syne_tune/config_space.py
+++ b/syne_tune/config_space.py
@@ -600,8 +600,8 @@ class FiniteRange(Domain):
     """
     def __init__(self, lower: float, upper: float, size: int,
                  log_scale: bool = False, cast_int: bool = False):
-        assert lower < upper
-        assert size >= 2
+        assert lower <= upper
+        assert size >= 1
         if log_scale:
             assert lower > 0.0
         self._uniform_int = randint(0, size - 1)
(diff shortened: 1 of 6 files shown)


syne_tune/config_space.py
```diff
@@ -967,3 +973,38 @@ def from_dict(d: Dict) -> Domain:
...
+    """
+    Restricts a numerical domain to be in the range [lower, upper]
+    :return:
+    """
+    assert hasattr(numerical_domain, "lower") and hasattr(numerical_domain, "upper")
+    assert lower <= upper
+    if not isinstance(numerical_domain, FiniteRange):
+        # domain is numerical, set new lower and upper ranges with bounding-box values
+        new_domain_dict = to_dict(numerical_domain)
+        new_domain_dict['domain_kwargs']['lower'] = lower
```
mseeger, line 988: Cast upper, lower to Domain type?

syne_tune/config_space.py
```diff
@@ -967,3 +973,38 @@ def from_dict(d: Dict) -> Domain:
...
+        new_domain_dict['domain_kwargs']['upper'] = upper
+        return from_dict(new_domain_dict)
+    else:
+        values = numerical_domain._values
+        assert lower < max(numerical_domain._values)
```
mseeger, line 993: Again, first cast upper, lower to domain type. Important for int.

syne_tune/config_space.py
```diff
@@ -967,3 +973,38 @@ def from_dict(d: Dict) -> Domain:
...
+        new_lower = values[i]
+
+        j = len(values) - 1
+        while upper < values[j] and i < j:
+            j -= 1
+        new_upper = values[j]
+        return FiniteRange(
+            lower=new_lower,
+            upper=new_upper,
+            size=len(values) - i - (len(values) - j - 1),
```
mseeger, line 1007: j - i + 1

syne_tune/config_space.py
```diff
@@ -612,38 +612,44 @@ def __init__(self, lower: float, upper: float, size: int,
...
             self._lower_internal = lower
-            self._step_internal = (upper - lower) / (size - 1)
+            self._step_internal = (upper - lower) / (size - 1) if size > 1 else 0
         else:
             self._lower_internal = np.log(lower)
             upper_internal = np.log(upper)
-            self._step_internal = \
-                (upper_internal - self._lower_internal) / (size - 1)
+            self._step_internal = (upper_internal - self._lower_internal) / (size - 1) if size > 1 else 0
+        self._values = [self._map_from_int(x) for x in range(self.size)]
```
mseeger, line 620: Could you make this a property, say self.values, and then use it in the function below?

syne_tune/experiments.py
```diff
@@ -144,13 +146,15 @@ def load_experiment(
...
         results = None
-    try:
-        tuner = Tuner.load(path)
-    except FileNotFoundError:
-        tuner = None
-    except Exception:
+    if load_tuner:
+        try:
+            tuner = Tuner.load(path)
+        except FileNotFoundError:
```
mseeger, line 152: Is this needed?

mseeger (commented):
I am missing hp_ranges_impl, it had the same bug as config_space.

syne_tune/config_space.py
geoalgo: Done.

syne_tune/config_space.py
geoalgo: Done.

syne_tune/config_space.py
geoalgo: I chose this way as it allows to cache the values when sampling.

syne_tune/experiments.py
geoalgo: There may be cases where you cannot load the tuner but still want the results.

geoalgo:
> I am missing hp_ranges_impl, it had the same bug as config_space.

Sorry, it is not clear what you mean by that.

syne_tune/config_space.py
mseeger: Yes, I like it, but maybe we should have a property, otherwise you access something internal below

mseeger:
What I mean: There is also .../bayesopt/datatypes/hp_ranges_impl.py, where you have HyperparameterRangeFiniteRange, and I think you want to add your change there as well, so it also supports finrange of size 1.

mseeger:
Right now, this would not work for BO methods, because they use hp_ranges_impl and would fail there.

syne_tune/config_space.py
geoalgo: I am not sure I follow, `_values` is accessed only by the class right?

geoalgo: Oh I see, you mean in the restrict domain method, I will add a property as well.

mseeger (approved):

PR #210, opened 2022-03-15
Tutorial for PASHA

ondrejbohdal:
A tutorial explaining PASHA and a new concise script that shows how to use PASHA.


docs/tutorials/pasha/scripts/launch_pasha_nasbench201.py
```diff
@@ -0,0 +1,206 @@
...
+from syne_tune.optimizer.schedulers.hyperband import HyperbandScheduler
+from syne_tune.backend.simulator_backend.simulator_callback import SimulatorCallback
+from benchmarking.definitions.definition_nasbench201 import nasbench201_benchmark, nasbench201_default_params
+from benchmarking.blackbox_repository.simulated_tabular_backend import BlackboxRepositoryBackend
+from benchmarking.blackbox_repository import load
+import pandas as pd
+import logging
+
+
+def load_nb201(dataset_name, nb201_random_seed):
```
mseeger, line 16: Why is this function needed?

ondrejbohdal: The idea is to put the data loading and processing part into a separate function so that the script becomes more readable (especially the part at the bottom). It should also make it slightly easier to reuse the code. However, it is not necessary for it to be within a function.

mseeger (approved):
I'd still prefer if you removed code that is not needed, but let us bring it in like this.

docs/tutorials/pasha/scripts/launch_pasha_nasbench201.py
ondrejbohdal: I see now why you asked about that part. The returned dataframe is actually used within the `analyse_experiment` function, but it was not explicitly passed as an argument to the function, which caused the confusion. I've now improved the code to make it clear. Thanks a lot for asking about it!

mseeger (approved):

geoalgo:
Hi @ondrejbohdal , 
I would like to merge this change but there is a conflict that needs to be resolved which prevents the merge right now, could you take a look?
Thanks for your contribution!

ondrejbohdal:
I've just looked at it and resolved the conflict. Thanks a lot!

PR #215, opened 2022-03-18
Allows blackboxes with a finite configuration space (such as nashpobe…

mseeger:
…nch) to use an infinite one with a surrogate

*Issue #, if available:*

*Description of changes:*
These are the remaining changes from #159 , which I will close.

### DIFF
--- a/benchmarking/definitions/definition_nashpobench.py
+++ b/benchmarking/definitions/definition_nashpobench.py
@@ -26,19 +26,26 @@ DATASET_NAMES = [
 ]
 
 
-_config_space = {
-    True: dict(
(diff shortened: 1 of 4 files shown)


benchmarking/definitions/definition_nashpobench.py
```diff
@@ -24,6 +26,21 @@
 ]
 
 
+_config_space = {
```
geoalgo, line 29: Could you add a comment there? What the boolean does may not be obvious for a reader who dont have all the context.

geoalgo (approved):

geoalgo (approved):

PR #216, opened 2022-03-28
Supporting deleting checkpoints once not needed anymore

mseeger:
*Issue #, if available:*

*Description of changes:*
This unblocks tuning of HuggingFace models without syncing checkpoints to S3. If checkpoints are written locally to ~/, the local disk fills up quickly.

Supporting this for SageMaker backend will be a forthcoming PR. Already implemented it, but a test is missing.

### DIFF
--- a/syne_tune/backend/local_backend.py
+++ b/syne_tune/backend/local_backend.py
@@ -205,14 +205,8 @@ class LocalBackend(TrialBackend):
             # (which allows to write a time-stamp when the process finishes) but it is probably OK if all_results
             # is called every few seconds.
             if os.path.exists(trial_path / "end"):
-                try:
-                    training_end_time = self._read_time_stamp(
-                        trial_id=trial_id, name="end")
-                except ValueError as exc:
-                    # File is present, but empty
-                    training_end_time = datetime.now()
-                    logger.warning(
-                        f"Timestamp file {str(trial_path / 'end')} exists, but is empty:\n{exc}")
+                training_end_time = self._read_time_stamp(
+                    trial_id=trial_id, name="end")
             else:
                 training_end_time = datetime.now()
 
--- a/syne_tune/backend/local_backend.py
+++ b/syne_tune/backend/local_backend.py
@@ -260,12 +260,8 @@ class LocalBackend(TrialBackend):
 
     def _write_time_stamp(self, trial_id: int, name: str):
         time_stamp_path = self._file_path(trial_id=trial_id, filename=name)
-        try:
-            with open(time_stamp_path, 'w') as f:
-                f.write(str(datetime.now().timestamp()))
-        except OSError as exc:
-            logger.warning(
-                f"Failed to write timestamp file {str(time_stamp_path)}:\n{exc}")
+        with open(time_stamp_path, 'w') as f:
+            f.write(str(datetime.now().timestamp()))
 
     def _read_time_stamp(self, trial_id: int, name: str):
         time_stamp_path = self._file_path(trial_id=trial_id, filename=name)
--- a/syne_tune/backend/local_backend.py
+++ b/syne_tune/backend/local_backend.py
@@ -43,6 +43,7 @@ class LocalBackend(TrialBackend):
             self,
             entry_point: str,
             rotate_gpus: bool = True,
+            delete_checkpoints: bool = False
     ):
         """
         A backend running locally by spawning sub-process concurrently.
@@ -54,9 +55,11 @@ class LocalBackend(TrialBackend):
             scheduled on a different GPU. A new trial is preferentially
             scheduled on a free GPU, and otherwise the GPU with least prior
             assignments is chosen
+        :param delete_checkpoints: If True, checkpoints of stopped or completed
+            trials are deleted
 
         """
-        super(LocalBackend, self).__init__()
+        super(LocalBackend, self).__init__(delete_checkpoints)
 
         assert Path(entry_point).exists(), f"the script provided to tune does not exist ({entry_point})"
         self.entry_point = entry_point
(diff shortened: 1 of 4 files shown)


syne_tune/backend/trial_backend.py
```diff
@@ -157,6 +179,8 @@ def fetch_status_results(self, trial_ids: List[int]) -> Tuple[Dict[int, Tuple[Tr
                     position_last_seen = self._last_metric_seen_index[trial_result.trial_id]
                     new_metrics = trial_result.metrics[position_last_seen:]
                     self._last_metric_seen_index[trial_result.trial_id] += len(new_metrics)
+                    if self.delete_checkpoints and trial_result.status == Status.completed:
```
geoalgo, line 182: This code is not needed since you call the deletion already in l137.

syne_tune/backend/local_backend.py
```diff
@@ -243,8 +260,12 @@ def _file_path(self, trial_id: int, filename: str):
 
     def _write_time_stamp(self, trial_id: int, name: str):
         time_stamp_path = self._file_path(trial_id=trial_id, filename=name)
-        with open(time_stamp_path, 'w') as f:
-            f.write(str(datetime.now().timestamp()))
+        try:
+            with open(time_stamp_path, 'w') as f:
+                f.write(str(datetime.now().timestamp()))
+        except OSError as exc:
```
geoalgo, line 266: Why is this one needed?

syne_tune/backend/local_backend.py
```diff
@@ -195,7 +205,14 @@ def _all_trial_results(self, trial_ids: List[int]) -> List[TrialResult]:
             # (which allows to write a time-stamp when the process finishes) but it is probably OK if all_results
             # is called every few seconds.
             if os.path.exists(trial_path / "end"):
-                training_end_time = self._read_time_stamp(trial_id=trial_id, name="end")
+                try:
+                    training_end_time = self._read_time_stamp(
+                        trial_id=trial_id, name="end")
+                except ValueError as exc:
```
geoalgo, line 211: Same question as bellow, why is this one needed?

syne_tune/backend/local_backend.py
```diff
@@ -74,14 +77,21 @@ def __init__(
...
+
     def copy_checkpoint(self, src_trial_id: int, tgt_trial_id: int):
-        src_checkpoint_path = self.trial_path(src_trial_id) / "checkpoints"
-        tgt_checkpoint_path = self.trial_path(tgt_trial_id) / "checkpoints"
+        src_checkpoint_path = self._checkpoint_trial_path(src_trial_id)
+        tgt_checkpoint_path = self._checkpoint_trial_path(tgt_trial_id)
         shutil.copytree(src_checkpoint_path, tgt_checkpoint_path)
 
+    def delete_checkpoint(self, trial_id: int):
+        checkpoint_path = self._checkpoint_trial_path(trial_id)
```
geoalgo, line 92: Logging a message with info or debug would be good here.

syne_tune/backend/local_backend.py
```diff
@@ -54,9 +55,11 @@ def __init__(
             scheduled on a different GPU. A new trial is preferentially
             scheduled on a free GPU, and otherwise the GPU with least prior
             assignments is chosen
+        :param delete_checkpoints: If True, checkpoints of stopped or completed
```
geoalgo, line 58: Currently with the modification proposed, the checkpoints are stopped when we stop, not just when they are completed. This is a good approach (cleaning them whenever stop happens) so I would be keen on keeping this behavior, we just need to update the doc to "If True, checkpoints folders of stopped trial are deleted."

geoalgo (commented):
LGTM, thanks for doing this, just have small comments.

syne_tune/backend/trial_backend.py
mseeger: Hi David, the one in l137 is only triggered when the trial is stopped, not when it ends on its own. For example, with FIFOScheduler (RS, BO), trials are never stopped.

One important point to note: Calling delete_checkpoint for a trial which does not have a checkpoint, is not doing anything. So you can just call it multiple times for the same trial.

syne_tune/backend/local_backend.py
mseeger: I had this exception being thrown. But I can take this out again, because if the local disk fills up, things will break anyway.

syne_tune/backend/local_backend.py
mseeger: I had this exception being thrown. I can take this out, it was probably again due to disk filling up.

But could it not happen that the file is written, but empty or corrupted? The code here would then continue and treat it as if it is not there.

syne_tune/backend/trial_backend.py
geoalgo: Ok this makes sense. 

Can we deal with this case only at this place then? It seems that we also probably want to delete checkpoints in other termination cases (when the status becomes failed, stopped in addition to completed).

syne_tune/backend/local_backend.py
geoalgo: Ok can we remove it then? It seems that this should be an error then if the file cant be written.

syne_tune/backend/local_backend.py
mseeger: Maybe I am wrong, but I thought if a trial ends on its own, it gets status completed but not stopped, right?

But even if a trial that gets stopped, is later changed to completed, that is fine, because delete_checkpoint can be called several times. It just does not do anything later on.

syne_tune/backend/local_backend.py
geoalgo: I would rather remove this catch because it does not seem we can recover the tuning if files cannot be written.

aaronkl (approved):

PR #218, opened 2022-03-30
Copy checkpoint files on S3 between trials, and delete checkpoint fil…

mseeger:
…es, for SageMaker backend

*Issue #, if available:*

*Description of changes:*
Recursive copy and recursive delete on S3, plugs the gap we had in the SageMaker backend.
There is a test for copy, but it is skipped (takes too long, needs SageMaker). I also tested delete manually.

### DIFF
--- a/syne_tune/backend/sagemaker_backend/sagemaker_backend.py
+++ b/syne_tune/backend/sagemaker_backend/sagemaker_backend.py
@@ -18,7 +18,9 @@ from typing import Dict, List, Optional
 import boto3
 from botocore.exceptions import ClientError
 import numpy as np
+
 from sagemaker import LocalSession, Session
+from sagemaker.estimator import Framework
 
 from syne_tune.backend.trial_backend import TrialBackend
 from syne_tune.constants import ST_INSTANCE_TYPE, ST_INSTANCE_COUNT, ST_CHECKPOINT_DIR
@@ -26,8 +28,8 @@ from syne_tune.util import s3_experiment_path
 from syne_tune.backend.trial_status import TrialResult, Status
 from syne_tune.backend.sagemaker_backend.sagemaker_utils import \
     sagemaker_search, get_log, sagemaker_fit, metric_definitions_from_names, \
-    add_syne_tune_dependency, map_identifier_limited_length
-from sagemaker.estimator import Framework
+    add_syne_tune_dependency, map_identifier_limited_length, \
+    s3_copy_files_recursively, s3_delete_files_recursively
 
 
 logger = logging.getLogger(__name__)
@@ -87,7 +89,7 @@ class SageMakerBackend(TrialBackend):
         self.resumed_counter = dict()
         if s3_path is None:
             s3_path = s3_experiment_path()
-        self.s3_path = s3_path
+        self.s3_path = s3_path.rstrip('/')
         self.tuner_name = None
 
     @property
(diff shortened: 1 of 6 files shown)


syne_tune/backend/sagemaker_backend/sagemaker_backend.py
```diff
@@ -129,6 +131,12 @@ def np_encoder(object):
                 return object.item()
         return json.loads(json.dumps(dict, default=np_encoder))
 
+    def _checkpoint_s3_uri_for_trial(self, trial_id: int) -> str:
+        parts = [self.s3_path, str(trial_id), 'checkpoints', '']
```
geoalgo, line 135: I think it would be cleaner to use pathlib rather than string operations there:

syne_tune/backend/sagemaker_backend/sagemaker_utils.py
```diff
@@ -284,3 +285,144 @@ def map_identifier_limited_length(
...
+    more_objects = True
+    continuation_kwargs = dict()
+    list_objects_kwargs = dict(Bucket=bucket,
+                               Prefix=prefix,
+                               Delimiter='/')
+    all_next_prefixes = []
+    num_action_calls = 0
+    num_successful_action_calls = 0
+    first_error_message = None
+    while more_objects:
```
geoalgo, line 318: Optional but the code would probably be easier with paginator, something like:

```python  
paginator = s3.get_paginator('list_objects_v2')
pages = paginator.paginate(**list_objects_kwargs)
  
for page in pages:
    for object in page['Contents']:
        print(object['Key'])
        all_next_prefixes.append(...)
```

https://yellowdesert.consulting/2019/07/29/how-to-paginate-in-boto3-use-collections-instead/

Generally when you have this next token, boto provides those "paginator" that allows to get an iterator without having to bother about passing next token around.

syne_tune/backend/sagemaker_backend/sagemaker_utils.py
```diff
@@ -284,3 +285,144 @@ def map_identifier_limited_length(
...
+        copy_source = dict(Bucket=bucket, Key=object_key)
+        ret_msg = None
+        try:
+            s3_client.copy_object(
+                CopySource=copy_source,
+                Bucket=trg_bucket,
+                Key=target_key)
+            logger.debug(f"Copied s3://{bucket}/{object_key}   to   s3://{trg_bucket}/{target_key}")
+        except ClientError as ex:
+            ret_msg = str(ex)
```
geoalgo, line 388: If we cannot copy a checkpoint, something is probably wrong and we should probably not try to continue as it would be hard to recover such an error.

geoalgo (commented):
Really nice thanks for adding this, also the test is nice (although we cant run it in our CI, it is still great to have).

syne_tune/backend/sagemaker_backend/sagemaker_utils.py
mseeger: Thanks. Let's keep it as it is, since it works.

syne_tune/backend/sagemaker_backend/sagemaker_utils.py
mseeger: OK, but this would be decided at a higher level. I'll put that in.

mseeger: Actually, already done it, your previous comment.

For deleting checkpoints, I'll keep the warning, because not being able to delete a checkpoint should not be a reason to fail.

geoalgo (approved):

syne_tune/backend/sagemaker_backend/sagemaker_utils.py
geoalgo: I am fine if you prefer to just have a warning, but I believe an error would be more suited.
If the user you specifically asks for the checkpoint to be deleted and they cannot be, this should be an error since we cannot do the operation the user asked.

PR #219, opened 2022-04-01
Cleaned up and added comments to RungSystem and subclasses for Hyperb…

mseeger:
…andScheduler

*Issue #, if available:*

*Description of changes:*
Mainly comments and clean-up.

### DIFF
--- a/syne_tune/optimizer/schedulers/hyperband_promotion.py
+++ b/syne_tune/optimizer/schedulers/hyperband_promotion.py
@@ -10,11 +10,13 @@
 # on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
+from typing import List, Tuple, Optional
+
 from syne_tune.optimizer.schedulers.hyperband_stopping import RungEntry, \
-    quantile_cutoff
+    quantile_cutoff, RungSystem
 
 
-class PromotionRungSystem(object):
+class PromotionRungSystem(RungSystem):
     """
     Implements both the promotion and stopping logic for an asynchronous
     variant of Hyperband, known as ASHA:
@@ -25,13 +27,10 @@ class PromotionRungSystem(object):
     promoted, which means that a free task picks up their evaluation until
     the next milestone.
 
-    Different to `StoppingRungSystem`, reward data at rungs is associated
-    with configs, not with tasks. To avoid having to use config as key, we
-    maintain unique integers as config keys.
-
-    The stopping rule is simple: Per task_id, we record the config key and
-    the milestone the task should be stopped at (it may still continue there,
-    if it directly passes the promotion test).
+    The rule to decide whether a paused trial is promoted (or remains
+    paused) is the same as in :class:`StoppingRungSystem`, except that
+    `continues` becomes `gets_promoted`. If several paused trials in a
+    rung can be promoted, the one with the best metric value is chosen.
 
     Note: Say that an evaluation is resumed from level resume_from. If the
     train_fn does not implement pause & resume, it needs to start training from
@@ -46,28 +45,28 @@ class PromotionRungSystem(object):
     """
     def __init__(self, rung_levels, promote_quantiles, metric, mode,
                  resource_attr, max_t):
-        self.max_t = max_t
-        self._metric = metric
-        self._mode = mode
-        self._resource_attr = resource_attr
-        # The data entry in _rungs is a dict mapping config_key to
-        # (reward_value, was_promoted)
+        super().__init__(metric, mode, resource_attr)
         assert len(rung_levels) == len(promote_quantiles)
+        self.max_t = max_t
+        # The data entry in `_rungs` is a dict mapping trial_id to
+        # (reward_value, was_promoted)
         self._rungs = [
             RungEntry(level=x, prom_quant=y, data=dict())
             for x, y in reversed(list(zip(rung_levels, promote_quantiles)))]
-        # _running maps trial_id to dict(milestone, resume_from).
-        # The tasks runs trial trial_id until resource reaches milestone.
-        # The resume_from field can be None. If not, the task is running a
-        # trial which has been resumed from rung level resume_from. This info
-        # is required for on_result to properly report ignore_data.
+        # `_running` maps `trial_id `to `dict(milestone, resume_from)`.
+        # The tasks runs trial `trial_id` until resource reaches milestone.
+        # The `resume_from` field can be None. If not, the task is running a
+        # trial which has been resumed from rung level `resume_from.` This info
+        # is required for `on_task_report` to properly report `ignore_data`.
         self._running = dict()
 
-    def _cutoff(self, recorded, prom_quant):
+    def _cutoff(self, recorded: dict, prom_quant: float):
         values = [x[0] for x in recorded.values()]
         return quantile_cutoff(values, prom_quant, self._mode)
 
-    def _find_promotable_config(self, recorded, prom_quant, resource):
+    def _find_promotable_trial(
+            self, recorded: dict, prom_quant: float,
+            resource: int) -> Optional[str]:
         """
         Check whether any not yet promoted entry in `recorded` is
         promotable, i.e. its value is better or equal to the cutoff
@@ -76,7 +75,7 @@ class PromotionRungSystem(object):
 
         :param recorded: Dict to scan
         :param prom_quant: Quantile for promotion
-        :param resource: Amount of resources spent on the rung.
+        :param resource: Rung level
         :return: trial_id if found, otherwise None
         """
         ret_id = None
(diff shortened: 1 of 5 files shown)


syne_tune/optimizer/schedulers/hyperband_promotion.py
```diff
@@ -46,28 +45,28 @@ class PromotionRungSystem(object):
     """
     def __init__(self, rung_levels, promote_quantiles, metric, mode,
                  resource_attr, max_t):
+        super().__init__(metric, mode, resource_attr)
+        assert len(rung_levels) == len(promote_quantiles)
```
wistuba, line 49: Can this be done in the parent class (same for StoppingRungSystem)

syne_tune/optimizer/schedulers/hyperband_promotion.py
```diff
@@ -46,28 +45,28 @@ class PromotionRungSystem(object):
...
-        self._metric = metric
-        self._mode = mode
-        self._resource_attr = resource_attr
-        # The data entry in _rungs is a dict mapping config_key to
+        # The data entry in `_rungs` is a dict mapping trial_id to
         # (reward_value, was_promoted)
-        assert len(rung_levels) == len(promote_quantiles)
         self._rungs = [
             RungEntry(level=x, prom_quant=y, data=dict())
             for x, y in reversed(list(zip(rung_levels, promote_quantiles)))]
```
wistuba, line 55: Can't this be moved to the parent class? (same for stopping)

syne_tune/optimizer/schedulers/hyperband_promotion.py
```diff
@@ -76,7 +75,7 @@ def _find_promotable_config(self, recorded, prom_quant, resource):
 
         :param recorded: Dict to scan
         :param prom_quant: Quantile for promotion
-        :param resource: Amount of resources spent on the rung.
+        :param resource: Rung level
```
wistuba, line 78: I found the old comment more expressive. It was removed since it is actually the amount of resources spent minus the resources spent earlier?

syne_tune/optimizer/schedulers/hyperband_stopping.py
```diff
@@ -34,58 +34,154 @@ def quantile_cutoff(values, prom_quant, mode):
...
+        self._mode = mode
         self._resource_attr = resource_attr
-        self._rungs = [
-            RungEntry(level=x, prom_quant=y, data=dict())
-            for x, y in reversed(list(zip(rung_levels, promote_quantiles)))]
 
-    def on_task_schedule(self):
+    def on_task_schedule(self) -> dict:
+        """
+        Called when new task is to be scheduled. For a promotion-bqsed
```
wistuba, line 54: typo: bqsed

syne_tune/optimizer/schedulers/hyperband_stopping.py
```diff
@@ -34,58 +34,154 @@ def quantile_cutoff(values, prom_quant, mode):
...
+        rung system, check whether any trial can be promoted. If so,
+        return `dict(trial_id, resume_from, milestone)`, where
```
wistuba, line 56: Unclear description: this dict is mapping from what to what?

syne_tune/optimizer/schedulers/hyperband_stopping.py
```diff
@@ -34,58 +34,154 @@ def quantile_cutoff(values, prom_quant, mode):
...
+        """
         return dict()
 
-    def on_task_add(self, trial_id, skip_rungs, **kwargs):
+    def on_task_add(self, trial_id: str, skip_rungs: int, **kwargs):
+        """
+        Called when new task is started.
+
+        :param trial_id:
+        :param skip_rungs: This number of lowest rung levels are not
```
wistuba, line 70: It's a bit a odd that this is described here but this isn't happening here. Is it possible to move the corresponding logic into this function?

syne_tune/optimizer/schedulers/hyperband_stopping.py
```diff
@@ -34,58 +34,154 @@ def quantile_cutoff(values, prom_quant, mode):
...
-        milestone. If True, next_milestone is the next milestone after
-        resource, or None if there is none.
+        Called when task is removed.
 
         :param trial_id:
-        :param result:
+        """
+        pass
+
+    def get_first_milestone(self, skip_rungs: int) -> int:
```
wistuba, line 101: Looks like we can implement this function and reuse in promotion/stopping. Same for other non-implemented functions?

syne_tune/optimizer/schedulers/hyperband_stopping.py
```diff
@@ -34,58 +34,154 @@ def quantile_cutoff(values, prom_quant, mode):
...
+        if cutoff is None:
+            return True
+        return metric_value <= cutoff if self._mode == 'min' else \
+            metric_value >= cutoff
+
+    def _milestone_rungs(self, skip_rungs: int) -> List[RungEntry]:
+        if skip_rungs > 0:
+            return self._rungs[:(-skip_rungs)]
+        else:
+            return self._rungs
```
wistuba, line 175: Duplicate function. Move it from stopping/promotion to RungSystem?

wistuba (changes requested):
Just to make this clear for me as a reviewer, this is **only** commenting and refactoring, it does not change the logic of any algorithm?

syne_tune/optimizer/schedulers/hyperband_promotion.py
mseeger: OK, can bring it back.

syne_tune/optimizer/schedulers/hyperband_stopping.py
mseeger: Don't understand. These methods are more like callbacks, they are called when something happens

syne_tune/optimizer/schedulers/hyperband_promotion.py
```diff
@@ -162,12 +173,20 @@ def on_task_add(self, trial_id, skip_rungs, **kwargs):
...
         metric_value = result[self._metric]
         assert trial_id not in recorded  # Sanity check
         recorded[trial_id] = (metric_value, False)
 
-    def on_task_report(self, trial_id, result, skip_rungs):
+    def _milestone_rungs(self, skip_rungs: int) -> List[RungEntry]:
+        if skip_rungs > 0:
+            return self._rungs[:(-skip_rungs)]
+        else:
+            return self._rungs
```
wistuba, line 186: I think this can be removed. The same function exists in RungSystem.

syne_tune/optimizer/schedulers/hyperband_stopping.py
```diff
@@ -34,58 +34,165 @@ def quantile_cutoff(values, prom_quant, mode):
...
+
+    in case `mode == 'min'`. See `_task_continues`.
+
+    """
+    def __init__(
+            self, rung_levels, promote_quantiles, metric, mode, resource_attr):
+        super().__init__(
+            rung_levels, promote_quantiles, metric, mode, resource_attr)
+        # The data entry in `_rungs` is a dict mapping trial_id to
+        # metric value
```
wistuba, line 162: can be deleted

wistuba (changes requested):
I little bit more cleanup is required as far as I understand the code.

syne_tune/optimizer/schedulers/hyperband_stopping.py
```diff
@@ -34,58 +34,165 @@ def quantile_cutoff(values, prom_quant, mode):
...
+
+    def _cutoff(self, recorded, prom_quant):
```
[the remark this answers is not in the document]
mseeger: It is not exactly the same. Value type of recorded is tuple in promotion, just float in stopping. Sure, I could now define a class for the value type, which has a subclass for promotion, but honestly one can overdo things.

syne_tune/optimizer/schedulers/hyperband_stopping.py
mseeger: The rationale of skip_rungs comes into play when you do Hyperband, where (a) each bracket has its own rung level system, or (b) all brackets share one system, but have different values of skip_rungs. So I cannot store skip_rungs in the rung level system, because in case (a) I don't need it (skip_rungs=0), and in case (b) I only have one system, but different values of skip_rungs.

I also don't understand what the problem is. These methods are just like callbacks, they are called when something happens. When I implement a class for callbacks, I don't have to implement all methods. And even if I do, I may not need all arguments passed into the method.

wistuba: I agree. Maybe one resolve this by removing skip_rungs here? What's strange to me is that the abstract class has skip_rungs as a param which isn't needed by all its children. So one could remove here and have it covered by kwargs and only introduce it in PromotionRungSystem. I won't insist it though.

wistuba (approved):
There are still two functions which are dead code. If they are removed, I'm fine with the code.

PR #226, opened 2022-04-11
Zero Shot Transfer Learning Method

wistuba:
*Description of changes:*
Added zero-shot transfer learning method.


syne_tune/optimizer/schedulers/transfer_learning/zero_shot.py
```diff
@@ -0,0 +1,151 @@
...
+            config_space: Dict,
+            transfer_learning_evaluations: Dict[str, TransferLearningTaskEvaluations],
+            metric: str,
+            mode: str = 'min',
+            sort_transfer_learning_evaluations: bool = True,
+            use_surrogates: bool = False,
+            random_seed: Optional[int] = None
+    ) -> None:
+        """
+        A zero-shot transfer hyperparameter optimization method which selects configurations that minimize the average
```
geoalgo, line 40: I would propose the following change:
```suggestion
        A zero-shot transfer hyperparameter optimization method which jointly selects configurations that minimize the average
```
since it may feel otherwise that you are just greedily picking the next top one.

syne_tune/optimizer/schedulers/transfer_learning/zero_shot.py
```diff
@@ -0,0 +1,151 @@
...
+        Martin Wistuba, Nicolas Schilling, Lars Schmidt-Thieme.
+        IEEE International Conference on Data Mining (ICDM) 2015.
+
+        :param config_space: Configuration space for trial evaluation function.
+        :param transfer_learning_evaluations: Dictionary from task name to offline evaluations.
+        :param metric: Objective name to optimize, must be present in transfer learning evaluations.
+        :param mode: Whether to minimize (min) or maximize (max)
+        :param sort_transfer_learning_evaluations: Use False if the hyperparameters for each task in
+        transfer_learning_evaluations Are already in the same order. If set to True, hyperparameters are sorted.
+        :param use_surrogates: If the same configuration is not evaluated on all tasks, set this to true. This will
```
geoalgo, line 53: Couldnt this one be inferred automatically? (that configurations are not all the same in all tasks)

syne_tune/optimizer/schedulers/transfer_learning/zero_shot.py
```diff
@@ -0,0 +1,151 @@
...
+            )
+            X_train = task_data.hyperparameters
+            y_train = task_data.objective_values(metric).mean(axis=1)
+            if self._mode == 'max':
+                y_train = y_train.max(axis=1)
+            else:
+                y_train = y_train.min(axis=1)
+            estimator.fit(X_train, y_train)
+
+            num_candidates = 10000 if len(config_space) >= 6 else 5 ** len(config_space)
```
geoalgo, line 116: Nit comment: I would imagine that it is hard to guess the number of samples required just given the dimension (all HPs could be constant but one) and that a constant number of candidates is easier to reason with.

syne_tune/optimizer/schedulers/transfer_learning/zero_shot.py
```diff
@@ -0,0 +1,151 @@
...
+            )
+        return surrogate_transfer_learning_evaluations
+
+    def get_config(self, **kwargs) -> Optional[Dict]:
+        if self._ranks.shape[1] == 0:
+            return None
+        best_idx = self._ranks.mean(axis=0).idxmin()
+        self._ranks.clip(upper=self._ranks[best_idx], axis=0, inplace=True)
+        self._scores.drop(columns=best_idx, inplace=True)
+        best_config = self._hyperparameters.loc[best_idx]
```
geoalgo, line 132: Those lines would probably deserve a comment given that it is the hearth of the method but understanding what is happening is not straightforward just given the lines in pandas.

geoalgo (commented):
LGTM, thanks for adding this.

syne_tune/optimizer/schedulers/transfer_learning/zero_shot.py
geoalgo: In term of runtime complexity, if you have n configs and k datasets, it would require `O(k n log n)` which does not sound too bad right? Probably less than a few seconds in the largest of our datasets I guess.

I leave you the choice so it is not a blocking comment :) 
My only important points are on adding comments.

geoalgo (approved):

PR #228, opened 2022-04-12
Python backend

geoalgo:
*Issue #, if available:*

*Description of changes:* add a backend that allows to tune python function rather than python scripts.

### DIFF
--- a/examples/launch_height_python_backend.py
+++ b/examples/launch_height_python_backend.py
@@ -1,0 +1,56 @@
+"""
+An example showing to launch a tuning of a python function `train_height`.
+"""
+
+from syne_tune import Tuner, StoppingCriterion
+from syne_tune.backend.python_backend.python_backend import PythonBackend
+from syne_tune.config_space import randint
+from syne_tune.experiments import load_experiment
+from syne_tune.optimizer.baselines import ASHA
+
+
+def train_height(steps: int, width: float, height: float):
+    import logging
+    from syne_tune import Reporter
+    import time
+    root = logging.getLogger()
+    root.setLevel(logging.INFO)
+    reporter = Reporter()
+    for step in range(steps):
+        dummy_score = (0.1 + width * step / 100) ** (-1) + height * 0.1
+        # Feed the score back to Syne Tune.
+        reporter(step=step, mean_loss=dummy_score, epoch=step + 1)
+        time.sleep(0.1)
+
+
+if __name__ == '__main__':
+    import logging
+
+    root = logging.getLogger()
+    root.setLevel(logging.DEBUG)
+
+    max_steps = 100
+    n_workers = 4
+
+    config_space = {
+        "steps": max_steps,
+        "width": randint(0, 20),
+        "height": randint(-100, 100)
+    }
+    mode = "min"
+    metric = "mean_loss"
+
+    scheduler = ASHA(config_space, metric=metric, resource_attr='epoch', max_t=max_steps, mode=mode)
+
+    trial_backend = PythonBackend(tune_function=train_height, config_space=config_space)
+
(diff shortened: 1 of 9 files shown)


examples/launch_height_python_backend.py
```diff
@@ -0,0 +1,54 @@
...
+
+from syne_tune import Tuner, StoppingCriterion
+from syne_tune.backend.python_backend import PythonBackend
+from syne_tune.config_space import randint
+from syne_tune.experiments import load_experiment
+from syne_tune.optimizer.baselines import ASHA
+
+
+def train_height(steps: int, width: float, height: float):
+    import logging
```
aaronkl, line 13: maybe add a comment here that imports have to be in the objective function

syne_tune/backend/python_backend/python_backend.py
```diff
@@ -0,0 +1,101 @@
...
+        )
+        self.config_space = config_space
+        # save function without reference to global variables or modules
+        self.tune_function = types.FunctionType(tune_function.__code__, {})
+        self.tune_function_path = self.local_path / "tune_function"
+
+    def set_path(self, results_root: Optional[str] = None, tuner_name: Optional[str] = None):
+        super(PythonBackend, self).set_path(results_root=results_root, tuner_name=tuner_name)
+        if self.local_path.exists():
+            print(f"path {self.local_path} already exists, make sure you have a unique tuner name.")
```
aaronkl, line 82: Why do we print this error message twice?

syne_tune/backend/python_backend/python_backend.py
```diff
@@ -0,0 +1,101 @@
...
+        self.save_tune_function(self.tune_function)
+        config = config.copy()
+        config["tune_function_root"] = str(self.tune_function_path)
+        # to detect if the serialized function is the same as the one passed by the user, we pass the md5 to the
+        # endpoint script. The hash is checked before executing the function.
+        config["tune_function_hash"] = file_md5(self.tune_function_path / "tune_function.dill")
+        super(PythonBackend, self)._schedule(trial_id=trial_id, config=config)
+
+    def save_tune_function(self, tune_function):
+        print(self.tune_function_path)
```
aaronkl, line 96: drop this print statement

syne_tune/backend/python_backend/python_backend.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def set_path(self, results_root: Optional[str] = None, tuner_name: Optional[str] = None):
+        super(PythonBackend, self).set_path(results_root=results_root, tuner_name=tuner_name)
+        if self.local_path.exists():
+            print(f"path {self.local_path} already exists, make sure you have a unique tuner name.")
+            logging.error(f"path {self.local_path} already exists, make sure you have a unique tuner name.")
+        self.tune_function_path = self.local_path / "tune_function"
+
+    def _schedule(self, trial_id: int, config: Dict):
+        self.save_tune_function(self.tune_function)
```
aaronkl, line 87: Why do we need to save the tune function every time we call schedule? Wouldn't it be enough to save it once in the init?

aaronkl (dismissed):
LGTM just some small comments

syne_tune/backend/python_backend/python_backend.py
geoalgo: This is a left-over, thanks for the catch.

syne_tune/backend/python_backend/python_backend.py
geoalgo: I see your point, the reason of not doing it initially in the constructor is that the path can be changed after the creation of the backend, I added a check so that the operation is only performed once.

geoalgo:
@aaronkl Thanks for the approval, I adapted the code with your comments, can you approve again?

PR #229, opened 2022-04-13
Load Benchmarks without S3 Credentials

wistuba:
*Description of changes:*
Try to upload preprocessed files to S3 but don't fail if it doesn't work.


geoalgo (approved):
Awesome, thank for spotting and fixing this.

PR #232, opened 2022-04-22
Fix s3-path

RSNirwan:
There is bug in the s3-key convertion in sagemaker backend. S3-key contains a double slash. The conversion to python `Path` removes one of them. Consequently, the returned string is not a proper key anymore.
We should keep the s3 key as string and not convert it to `Path`. This pr fixes the issue.

```python
>>> from pathlib import Path
>>> str(Path("s3://xxx"))
's3:/xxx'
```

### DIFF
--- a/syne_tune/backend/sagemaker_backend/sagemaker_backend.py
+++ b/syne_tune/backend/sagemaker_backend/sagemaker_backend.py
@@ -132,11 +132,10 @@ class SageMakerBackend(TrialBackend):
         return json.loads(json.dumps(dict, default=np_encoder))
 
     def _checkpoint_s3_uri_for_trial(self, trial_id: int) -> str:
-        res_path = Path(self.s3_path)
+        res_path = self.s3_path
         if self.tuner_name is not None:
-            res_path = res_path / self.tuner_name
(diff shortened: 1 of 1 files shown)


syne_tune/backend/sagemaker_backend/sagemaker_backend.py
```diff
@@ -132,11 +132,10 @@ def np_encoder(object):
         return json.loads(json.dumps(dict, default=np_encoder))
 
     def _checkpoint_s3_uri_for_trial(self, trial_id: int) -> str:
-        res_path = Path(self.s3_path)
+        res_path = self.s3_path
```
mseeger, line 135: If the last character is '/', you'd have to remove it.

An alternative is to strip off the "s3://" prefix and then use Path

mseeger (approved):

syne_tune/backend/sagemaker_backend/sagemaker_backend.py
RSNirwan: the `self.s3_path.rstrip('/')`is already in the init of the class.

PR #233, opened 2022-04-22
Compute Statistics for Numeric Results Only

wistuba:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/syne_tune/tuning_status.py
+++ b/syne_tune/tuning_status.py
@@ -11,6 +11,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 import numbers
+import logging
 import time
 from collections import defaultdict, OrderedDict
 from typing import List, Dict, Tuple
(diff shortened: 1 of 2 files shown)


syne_tune/tuning_status.py
```diff
@@ -36,16 +36,12 @@ def __init__(self):
...
-                self.metric_names.append(name)
-                self.min_metrics[name] = np.inf
-                self.max_metrics[name] = -np.inf
-                self.sum_metrics[name] = 0
-
-        self.min_metrics = {m: min(metrics[m], current) for m, current in self.min_metrics.items()}
-        self.max_metrics = {m: max(metrics[m], current) for m, current in self.max_metrics.items()}
-        self.sum_metrics = {m: metrics[m] + current for m, current in self.sum_metrics.items()}
+        for metric_name, current_metric in metrics.items():
+            if isinstance(current_metric, numbers.Number):
```
geoalgo, line 40: This should fix the issue indeed, thanks! (I was going to ask about adding this case to the unit-test but you already did it since the first commit :-)

One bug that could happen is that one emits numerical metrics, and then a non numerical one (which I think could happen pretty often). 

Could you either:
* let a TODO comment to deal with this case later
* or store a boolean per metric indicating if they are numerical and throws an error if the type change?

geoalgo (dismissed):

wistuba:
> oh I see, the metric would be added and then removed... That seems indirect, wouldnt this be simpler?
> 
> ```python
>                 is_current_metric_numeric = isinstance(current_metric, numbers.Number)
>                 self.is_numeric[metric_name] = is_current_metric_numeric
>                 if is_current_metric_numeric:
>                     self.min_metrics[metric_name] = min(self.min_metrics.get(metric_name, np.inf), current_metric)
>                     self.max_metrics[metric_name] = max(self.max_metrics.get(metric_name, -np.inf), current_metric)
>                     self.sum_metrics[metric_name] = self.sum_metrics.get(metric_name, 0) + current_metric
>                 # no else / pop statements needed, we only add metrics that are numerical
> ```

we need this for scenarios where we observe first numeric values but then observe a string. in that case we want to change it to non-numeric but also remove the stats.

geoalgo (approved):

PR #235, opened 2022-04-25
Fix metadata remote

RSNirwan:
Metadata isn't read after downloading the file from s3, but is set to None. Downloading needs to happen before metadata is read. This PR fixes the issue.


geoalgo (approved):
Thanks!

PR #240, opened 2022-04-29
Add strictly_non_binary_choice_hyperparameter_keys to enforce non-bin…

hfurkanbozkurt:
…ary warm-starting task range parameter

*Issue #, if available:*

*Description of changes:* Introduces a new parameter `strictly_non_binary_choice_hyperparameter_keys` to make sure task attr parameter is non-binary.

### DIFF
--- a/syne_tune/optimizer/schedulers/searchers/bayesopt/datatypes/hp_ranges.py
+++ b/syne_tune/optimizer/schedulers/searchers/bayesopt/datatypes/hp_ranges.py
@@ -32,7 +32,9 @@ class HyperparameterRanges(object):
             self, config_space: Dict, name_last_pos: Optional[str] = None,
             value_for_last_pos=None,
             active_config_space: Optional[Dict] = None,
-            prefix_keys: Optional[List[str]] = None):
+            prefix_keys: Optional[List[str]] = None,
+            strictly_non_binary_choice_hyperparameter_keys: Optional[List[str]] = None
+    ):
         """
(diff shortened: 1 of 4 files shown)


syne_tune/optimizer/schedulers/searchers/bayesopt/datatypes/hp_ranges.py
```diff
@@ -65,12 +67,15 @@ def __init__(
         :param prefix_keys: If given, these keys into `config_space` come first
             in the internal ordering, which determines the internal
             encoding
+        :param strictly_non_binary_choice_hyperparameter_keys: If given, these keys are considered to be strictly
```
mseeger, line 70: I don't understand the rationale for this. The binary vs non-binary encoding is an internal property of the child class. Nothing in warmstarting should depend on how categorical HPs are internally encoded.

hfurkanbozkurt: Here is the part where we depend on the dimension of each hyperparameter: https://github.com/awslabs/syne-tune/blob/main/syne_tune/optimizer/schedulers/searchers/utils/warmstarting.py#L105-L108

Since binary hyperparameters have dimensionality of 1, we calculate size of this kernel wrong.

mseeger: Better solution: #241

PR #247, opened 2022-05-18
implements interface to use BORE as model for Hyperband

aaronkl:
*Issue #, if available:*

*Description of changes:* Implements an interface such that we can use BORE as model for Hyperband. This PR also adds some random sampling without replacement to BORE

### DIFF
--- a/syne_tune/optimizer/schedulers/searchers/bore/bore.py
+++ b/syne_tune/optimizer/schedulers/searchers/bore/bore.py
@@ -35,12 +35,12 @@ class Bore(SearcherWithRandomSeed):
 
     def __init__(
             self, config_space: dict, metric: str,
-            points_to_evaluate=None, random_seed_generator=None,
+            points_to_evaluate=None,
             random_seed=None, mode: str = 'max', gamma: float = 0.25,
             calibrate: bool = False, classifier: str = 'xgboost',
             acq_optimizer: str = 'rs', feval_acq: int = 500,
             random_prob: float = 0.0, init_random: int = 6,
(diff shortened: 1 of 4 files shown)


syne_tune/optimizer/schedulers/searchers/bore/multi_fidelity_bore.py
```diff
@@ -0,0 +1,117 @@
...
+import numpy as np
+
+from syne_tune.optimizer.schedulers.searchers.bore import Bore
+
+__all__ = ['MultiFidelityBore']
+
+logger = logging.getLogger(__name__)
+
+
+class MultiFidelityBore(Bore):
```
geoalgo, line 24: Can you add it to scheduler unit-tests?

syne_tune/optimizer/schedulers/searchers/bore/bore.py
```diff
@@ -140,38 +139,56 @@ def get_config(self, **kwargs):
...
+                        xi = self._hp_ranges.random_config(self.random_state)
+                        if xi not in X:
+                            X.append(xi)
+                            values.append(self.loss(self._hp_ranges.to_ndarray(xi))[0])
+                            counter = 0
+                        else:
+                            logging.warning("Re-sampled the same configuration. Retry...")
+                            counter += 1  # we stop sampling if after 10 retires we are not able to find a new config
+
+                    ind = np.array(values).argmin()
```
geoalgo, line 190: Can you add a warning in the case where less than feval_acq samples are found?

PR #257, opened 2022-06-01
order methods in table by average rank

geoalgo:
*Description of changes:* print result table with methods sorted by average rank


geoalgo:
> Order isn't according to average rank, ASHA-BB < ZS. This is already addressed in #256.

AFAIK in the new results, the average rank of RUSH is 0.36 and ZS 0.3 so ZS should appear after.

PR #267, opened 2022-06-16
Code for replaying events to create HyperbandScheduler state (needed …

mseeger:
…for a stateless API)

*Issue #, if available:*

*Description of changes:*
This code is used to allow for a stateless API for HyperbandScheduler, which is needed by our partners. I also include quite some code to test this, in particular you can compare the inner states between running an experiment normally and replaying it.


mseeger:
We don't need this, to be honest, and it is kind of ugly.

PR #270, opened 2022-06-20
Add standalone Bayesian optimization workflow

jjaeyeon:
*Issue #, if available:*

*Description of changes:*
Added standalone bayesian optimization workflow to operate upon the core subset of dependencies.
For this change, additional changes in parts of the code have been made:
- conditional import in backend
- comment out metrics_df (this requires pandas, but after checking that this function is not yet used in other parts, commented out for now)


syne_tune/backend/trial_status.py
```diff
@@ -81,8 +81,8 @@ def cost(self):
...
-    def metrics_df(self) -> pd.DataFrame:
-        if len(self.metrics) == 0:
-            return pd.DataFrame()
-        else:
-            return pd.DataFrame(data=self.metrics)
+    # def metrics_df(self) -> pd.DataFrame:
+    #     if len(self.metrics) == 0:
+    #         return pd.DataFrame()
+    #     else:
+    #         return pd.DataFrame(data=self.metrics)
```
geoalgo, line 88: Good catch, it does not seem used. We can remove the lines altogether though (preferable than adding dead comments https://gist.github.com/wojteklu/73c6914cc446146b8b533c0988cf8d29).

syne_tune/try_import.py
```diff
@@ -38,6 +38,12 @@ def try_import_raytune_message() -> str:
         tag="raytune"
     )
 
+def try_import_backend_message(backend_type: str) -> str:
+    return (
+        f"{backend_type} is not imported"
+        + " since dependencies are missing. You can install them with\n"
+        + "   pip install 'syne-tune[extra]'"
+    )
```
geoalgo, line 46: Is there a reason for not prefering the same convention as in methods above?
```suggestion
    return _try_import_message(
        f"Backend {backend} not imported",
        tag="extra",
    )
```

.github/workflows/check-standalone-bayesian-optimization.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
+        python -m pip install -r requirements-gpsearchers.txt --upgrade pip
+        python -m pip install typing_extensions
```
geoalgo, line 23: Is this one needed? (It should be included in 3.8 AFAIK)

examples/launch_standalone_bayesian_optimization.py
```diff
@@ -0,0 +1,68 @@
...
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+import logging
+import sys, os
+
+sys.path.append(os.path.join(os.path.dirname(__file__), '..'))
```
geoalgo, line 16: Can you try to remove this one? It seems that the error you got was an unused import, but this can be solved by using `#noqa` flag, for instance:

```python
# from syne_tune.optimizer.baselines import PASHA, BORE  # noqa: F401
```

geoalgo (commented):
LGTM overall, just have some small comments.

syne_tune/backend/__init__.py
```diff
@@ -13,6 +13,24 @@
 
 __all__ = ["LocalBackend", "SageMakerBackend", "PythonBackend"]
 
-from syne_tune.backend.local_backend import LocalBackend
-from syne_tune.backend.python_backend.python_backend import PythonBackend
-from syne_tune.backend.sagemaker_backend.sagemaker_backend import SageMakerBackend
+import logging
```
mseeger, line 16: David is the expert here. But I believe the dependencies for the backends are in the core.

Of course, we could imagine moving in particular the sagemaker dependencies out of the core.

syne_tune/try_import.py
jjaeyeon: The import for backend would not necessarily require specific tag, so in purpose of providing a better context, I've created separate function. Do you have any other suggestion on this part?

.github/workflows/check-standalone-bayesian-optimization.yml
jjaeyeon: As far as I've tested in my local repo, there was an error resulted from missing package of typing_extensions

examples/launch_standalone_bayesian_optimization.py
jjaeyeon: We need a manual configuring of the file path, since we are not running setup.py, it cannot identify syne_tune in the directory. The error message I've got when the line is omitted : https://github.com/jjaeyeon/syne-tune/runs/6939346421?check_suite_focus=true

.github/workflows/check-standalone-bayesian-optimization.yml
geoalgo: Oh I see, it is because we have a typing_extensions import left-over (should be removed and then we can ditch this dependency). Lets keep it that way for this PR, I will fix it in another one.

syne_tune/try_import.py
geoalgo: No I see now why you had to create it, lets keep it this way.

examples/launch_standalone_bayesian_optimization.py
jjaeyeon: Update: I've remove the line and add PYTHONPATH in the workflow

.github/workflows/check-standalone-bayesian-optimization.yml
```diff
@@ -0,0 +1,31 @@
...
+    - name: Check that stand-alone bayesian-optimation example runs
+      run: |
+        PYTHONPATH='./' python examples/launch_standalone_bayesian_optimization.py
```
geoalgo, line 26: Awesome, this is much cleaner, small nit the following is more standard AFAIK:
```suggestion
        PYTHONPATH=. python examples/launch_standalone_bayesian_optimization.py
```

geoalgo (approved):
LGTM, just left a small suggestion, thanks for the effort for simplifying.

geoalgo (dismissed):
Nice, thanks!

syne_tune/backend/__init__.py
mseeger: I have to say I still do not understand this. I recently ran random search HPO with the local backend, having installed just the core dependencies (which include SageMaker), and this works fine. So I don't really understand under which circumstances the backend dependencies are not present. Certainly, 'extra' is not needed.

mseeger (approved):
Maybe you could leave a comment in the new launcher script, on what is the purpose?

.github/workflows/check-standalone-bayesian-optimization.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
+        python -m pip install -r requirements-gpsearchers.txt --upgrade pip
```
mseeger, line 22: OK, I understand now, you are not installing requirements.txt.

Funny this works, the requirements do not even include numpy.

mseeger (approved):

syne_tune/backend/__init__.py
```diff
@@ -11,8 +11,26 @@
...
+
+from syne_tune.try_import import try_import_backend_message
+
+try:
+    from syne_tune.backend.local_backend import LocalBackend
+except ImportError:
+    __all__.remove("LocalBackend")
```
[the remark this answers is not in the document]
jjaeyeon: It was supposed to be appending, I've messed up with the commit :'( Will update this

.github/workflows/check-standalone-bayesian-optimization.yml
jjaeyeon: AFAIK scipy depends on numpy so numpy would be installed together with that requirements

jjaeyeon:
Hi @mseeger and @geoalgo, I've just updated a minor fix for importing. Thanks again for reviewing!

mseeger (approved):

PR #280, opened 2022-06-24
add tensorboard visualization

aaronkl:
*Description of changes:* Adds a simple tuner callback to log results such that they can be visualizes with Tensorboard.


syne_tune/tuner_callback.py
```diff
@@ -145,3 +146,90 @@ def on_tuning_end(self):
...
+class TensorboardCallback(TunerCallback):
+    def __init__(
+            self,
+
+            metric_to_optimize: 'str',
+            mode: 'str',
+            output_metrics: List[str] = [],
+    ):
+        """
+        Simple callback that logs different statistics such that we can visualize them with Tensorboard.
```
geoalgo, line 160:
```suggestion
        Simple callback that logs different statistics such that we can visualize with Tensorboard.
```

syne_tune/tuner_callback.py
```diff
@@ -145,3 +146,90 @@ def on_tuning_end(self):
         # store the results in case some results were not committed yet (since they are saved every
         # `results_update_interval` seconds)
         self.store_results()
+
+
+class TensorboardCallback(TunerCallback):
+    def __init__(
+            self,
+
```
geoalgo, line 154: Unwanted space.

syne_tune/tuner_callback.py
```diff
@@ -145,3 +146,90 @@ def on_tuning_end(self):
...
+            metric_to_optimize: 'str',
+            mode: 'str',
+            output_metrics: List[str] = [],
```
geoalgo, line 157: Should be `Optional[List[str]] = None` to avoid side-effects. 

Also, I think the default behavior should rather be that if this is None then all the report metrics are tracked (the user could still decide to track none by passing `[]`).

Also, passing the mode is a bit annoying especially if its mandatory. Maybe a good first version would be to log all metrics without cummin/cummax for now? (it is also closer to the typical usage of tensor-board).

syne_tune/tuner.py
```diff
@@ -403,6 +403,14 @@ def save(self, folder: Optional[str] = None):
...
         with open(tuner_serialized_path, "wb") as f:
             logger.debug(f"saving tuner in {tuner_serialized_path}")
+
+            # We need to delete the TensorboardCallback before we serialize the Tuner, to avoid runtime errors because
+            # of the MultiProcessing Queues of TensorboardX
+            for callback in self.callbacks:
+                from syne_tune.tuner_callback import TensorboardCallback
+                if isinstance(callback, TensorboardCallback):
+                    self.callbacks.remove(callback)
+
```
geoalgo, line 413: Can we remove this part? We should really avoid to have this depend on special use-cases (as much as we can of-course). 

In particular here, you can avoid the serialization issue by just having `__getstate__` and `__setstate__` in your Tensorboard callback that dont save stuff you dont want that are unneeded and not serializable.

examples/launch_tensorboard_example.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
geoalgo, line 1: Can you put a high-level comment on this example that explains what is happening? (e.g. that tensorboard is used to log metrics and that explains how to visualize the metrics)

Also it would be nice to have it link in the FAQ (something like How can I visualize the progress of my tuning experiment with Tensor board?)

geoalgo:
Cool stuff, thanks for adding this!

syne_tune/tuner_callback.py
```diff
@@ -145,3 +146,90 @@ def on_tuning_end(self):
...
+        self.trial_ids.add(trial.trial_id)
+        self.writer.add_scalar('num_trial', len(self.trial_ids),
+                               self._iter, display_name='total number of trials')
+
+        self._iter += 1
+
+    def on_tuning_start(self, tuner):
+
+        output_path = os.path.join(tuner.tuner_path, 'tensorboard_output')
+        from tensorboardX import SummaryWriter
```
mseeger, line 228: Could you do try except ImportError with a message hinting to install tensorboard?

syne_tune/tuner_callback.py
```diff
@@ -145,3 +146,90 @@ def on_tuning_end(self):
...
+
+        for key, value in self.curr_best_config.items():
+            self.writer.add_scalar(f'optimal_{key}', value, self._iter)
+
+        for metric in self.output_metrics:
+            self.writer.add_scalar(metric, result[metric], self._iter)
+
+        self.writer.add_scalar('runtime', result[ST_TUNER_TIME], self._iter)
+
+        self.trial_ids.add(trial.trial_id)
```
mseeger, line 219: Can you also output trial_id? Otherwise, how can the trial be identified? I'd not count on trial_id being included in result.

syne_tune/tuner_callback.py
```diff
@@ -145,3 +146,90 @@ def on_tuning_end(self):
...
+        :param mode: Defines whether we minimize ('min') or maximize ('max'). Should be the same what we
+        pass to scheduler / searcher.
+        :param output_metrics: All other metrics that we want to log. Make sure that these metrics
+         are reported correctly in the train function to Syne Tune via the report(...) function.
+        """
+        self.results = []
+
+        self.output_metrics = output_metrics
+        self.mode = mode
+        self.metric = metric_to_optimize
```
mseeger, line 173: A bit confusing. This is not the metric reported, but its cumulative optimum. If the user passes the same name as the main metric, there'll be a clash

syne_tune/tuner_callback.py
```diff
@@ -145,3 +146,90 @@ def on_tuning_end(self):
...
+        :param metric_to_optimize: The metric that we aim to optimize. We report the cumulative optimum over time. This
+        should be the same metric that we pass to the scheduler / searcher.
+        :param mode: Defines whether we minimize ('min') or maximize ('max'). Should be the same what we
+        pass to scheduler / searcher.
+        :param output_metrics: All other metrics that we want to log. Make sure that these metrics
+         are reported correctly in the train function to Syne Tune via the report(...) function.
+        """
+        self.results = []
+
+        self.output_metrics = output_metrics
```
mseeger, line 171: This needs to include resource_attr for multi-fidelity, maybe add some comment?

mseeger: One could also do some "magic" in on_tuning_start, where you have the tuner object, to make sure the relevant metrics are output for sure at least for the main schedulers we care about.

syne_tune/tuner_callback.py
```diff
@@ -145,3 +146,90 @@ def on_tuning_end(self):
...
+    def on_trial_result(self, trial: Trial, status: str, result: Dict, decision: str):
+
+        self._set_time_fields(result)
+
+        new_result = result[self.metric]
+
+        if self.mode == 'max':
+            new_result *= -1
+
+        if self.curr_best_value is None or self.curr_best_value > new_result:
```
mseeger, line 200: Instead of all this multiplying with -1, why not just do this?

metric_sign = -1 if self.mode == 'max' else 1

if ... or metric_sign * (self.curr_best_value - new_result) > 0

syne_tune/tuner_callback.py
geoalgo: I guess the current behavior is to show the performance across all trials as a first step as its easier in TB.

aaronkl: I am not entirely following. The idea here is just to show how many trials have been evaluated at any point in time. Note we also report the best values for each hyperparameter over time

syne_tune/tuner_callback.py
aaronkl: ah that's a good point

syne_tune/tuner.py
aaronkl: ah perfect, that's exactly what I was looking for :-)

syne_tune/tuner_callback.py
mseeger: Fine for now. I don't know TB. But I suppose if we want to visualize things like what David did in the gif, so the trajectory for every trial in a different color (say), then we need to know the trial_id for each record. But yes, that can come later I suppose.

It would, however, be very cool to show this off a bit in the docs, and to go beyond just the "current best" plot

syne_tune/tuner_callback.py
aaronkl: Good point. I changed this such that by default all metric are reported, but users could define which metric to ignore. I still think it would be quite handy to plot the cumulative optimum and the best hyperparameter found so far. I suggest to add an optional parameter, that, if set, will report the cumulative optimum to tensoboard.

syne_tune/tuner_callback.py
```diff
@@ -149,37 +152,38 @@ def on_tuning_end(self):
...
-        self.output_metrics = output_metrics
-        self.mode = mode
-        self.metric = metric_to_optimize
+        self.ignore_metrics = ignore_metrics
 
         self.curr_best_value = None
         self.curr_best_config = None
 
         self._start_time_stamp = None
         self.writer = None
```
geoalgo, line 179: Can we use `_` for fields consistently in this class? Right now some fields are with `_` and others aren't whereas it seems that all fields would be private.

syne_tune/tuner_callback.py
```diff
@@ -192,44 +196,55 @@ def on_trial_result(self, trial: Trial, status: str, result: Dict, decision: str
...
-            if self.mode == 'max':
-                opt *= -1
-            self.writer.add_scalar(self.metric, opt, self._iter)
+            for key, value in self.curr_best_config.items():
+                self.writer.add_scalar(f'optimal_{key}', value, self._iter)
 
-        for key, value in self.curr_best_config.items():
-            self.writer.add_scalar(f'optimal_{key}', value, self._iter)
+        for metric in result:
+            if self.ignore_metrics is not None and metric not in self.ignore_metrics:
```
geoalgo, line 217: Nit: It might be cleaner to set `ignore_metrics` to `{}` when nothing is passed and then directly do:
```suggestion
            if metric not in self.ignore_metrics:
```

syne_tune/tuner_callback.py
```diff
@@ -192,44 +196,55 @@ def on_trial_result(self, trial: Trial, status: str, result: Dict, decision: str
...
+                self.writer.add_scalar(metric, result[metric], self._iter)
 
-        for metric in self.output_metrics:
-            self.writer.add_scalar(metric, result[metric], self._iter)
+        for key, value in trial.config.items():
+            self.writer.add_scalar(key, value, self._iter)
```
geoalgo, line 221: For my understanding, the rational of dumping configs keys to TB is to let the user possibly look at their distribution over time?

syne_tune/tuner_callback.py
geoalgo: @mseeger we could support it in the future but TB is not great when you start having lots of metrics to choose from for display.

syne_tune/tuner_callback.py
aaronkl: yes exactly

syne_tune/tuner_callback.py
```diff
@@ -155,7 +155,7 @@ class TensorboardCallback(TunerCallback):
 
     def __init__(
             self,
-            ignore_metrics: Optional[List[str]] = None,
+            ignore_metrics: Optional[List[str]] = [],
```
geoalgo, line 158: Its important to avoid passing mutable arguments as defaults as it can have side-effects. Can you revert it to None?

```

    def __init__(
            self,
            ignore_metrics: Optional[List[str]] = None
    ):
      self.ignore_metrics = ignore_metrics if ignore_metrics else []
  ```

File: syne_tune/tuner_callback.py [hunk 15]
```diff
@@ -189,62 +189,95 @@ def _set_time_fields(self, result: Dict):
...
+            "iter": self.iter,
+            "mode": self.mode,
+            "target_metric": self.target_metric,
+            "trial_ids": self.trial_ids,
+            "metric_sign": self.metric_sign,
+            'output_path': self.output_path
+        }
+        return state
+
+    def __setstate__(self, state):
```
geoalgo, line 262: Cool, thanks for adding this.

syne_tune/tuner_callback.py
```diff
@@ -189,62 +189,95 @@ def _set_time_fields(self, result: Dict):
...
+            "ignore_metrics": self.ignore_metrics,
+            "curr_best_value": self.curr_best_value,
+            "curr_best_config": self.curr_best_config,
+            "start_time_stamp": self.start_time_stamp,
+            "iter": self.iter,
+            "mode": self.mode,
+            "target_metric": self.target_metric,
+            "trial_ids": self.trial_ids,
+            "metric_sign": self.metric_sign,
+            'output_path': self.output_path
```
geoalgo, line 258: It seems that the formatting is not working in your case.
Can you verify that you have black setup? Have you ran `./dev_setup.sh` to install black hook?

geoalgo (commented):
LGTM, it seems that you dont have black formatting activated, have you run `./dev_setup.sh`? I just have a small comment regarding not using mutable arguments for default parameters and then its good to go (looking forward to have it!).

geoalgo (approved):
Cool to have this, thanks!

Issue #265, opened 2022-06-15
Grid search in syne-tune


iaroslav-ai:
Hey folks,
would you be interested in grid search implemented in syne-tune? I had a few offline discussions with some of you already, and it seems that you are not against grid search added to syne-tune, but want to keep a record of that here. 

Additionally, would you have any pointers as to what would be the best way to add grid search to syne-tune?

mseeger:
Hi Iaro, as I mentioned, the easiest way would be to use FIFOScheduler with random search and send in the grid via points_to_evaluate.

aaronkl:
I agree that this would be the most straight-forward way to run grid search. However, it might be more accessible for non-expert syne tune users to have a simple baseline interface called GridSearch that wraps this functionality?

mseeger:
Agree with Aaron, but this could simply wrap what I was suggesting, so we don't need to write redundant code.

iaroslav-ai:
Thank you all for your inputs!!

iaroslav-ai:
I will keep this open for any communication on addition of grid search to syne-tune.

geoalgo:
> I agree that this would be the most straight-forward way to run grid search. However, it might be more accessible for non-expert syne tune users to have a simple baseline interface called GridSearch that wraps this functionality?

I agree with this point. The important part is probably to have a good way for the user to instantiate a grid-search. For instance, it could be something like:

```python
from syne_tune.optimizer.baselines import GridSearch
config_space = {
    "width": randint(0, 20),
    "height": randint(-100, 100),
}
scheduler = GridSearch(config_space, metric=metric, num_points=100)
```

We could also specify the list directly:

```python
from syne_tune.optimizer.baselines import GridSearch
scheduler = GridSearch(
  points_to_evaluate=[{"width": 0, "height": -100}, ...], 
  metric=metric, 
  num_points=100
)
```
however, this choice would make comparison with other schedulers harder (as they all take config_space as input) and also requires the user to split the grid manually. 

Given this, I think the best solution is to have the first option (config_space as input), we could also support an optional `points_to_evaluate` but having  both may be confusing.

mseeger:
We could be even more ambitious and include some support for Latin Hypercube and/or Sobel sequences or pseudo-random stuff. If anybody knows some good code we could wrap?

mseeger:
Actually, maybe it would be cleaner to implement this as a searcher.

geoalgo:
[...]

PR #284, opened 2022-06-28
Grid search

duck105:
*Issue #265 

*Description of changes: Adding GridSearcher and baseline for it (Unit test haven't included)

### DIFF
--- a/syne_tune/optimizer/baselines.py
+++ b/syne_tune/optimizer/baselines.py
@@ -32,6 +32,16 @@ class RandomSearch(FIFOScheduler):
         )
 
 
+class GridSearch(FIFOScheduler):
+    def __init__(self, config_space: Dict, metric: str, **kwargs):
+        super(GridSearch, self).__init__(
+            config_space=config_space,
+            metric=metric,
+            searcher="grid",
+            **kwargs,
+        )
+
+
 class BayesianOptimization(FIFOScheduler):
     def __init__(self, config_space: Dict, metric: str, **kwargs):
         super(BayesianOptimization, self).__init__(
@@ -318,6 +328,7 @@ class ASHACTS(HyperbandScheduler):
 # dictionary that allows to also list baselines who don't need a wrapper class such as PBT.
 baselines_dict = {
     "Random Search": RandomSearch,
+    "Grid Search": GridSearch,
     "Bayesian Optimization": BayesianOptimization,
     "ASHA": ASHA,
     "MOBSTER": MOBSTER,
--- a/syne_tune/optimizer/schedulers/searchers/searcher.py
+++ b/syne_tune/optimizer/schedulers/searchers/searcher.py
@@ -598,7 +598,7 @@ class GridSearcher(BaseSearcher):
         # Get hp values that are specified as Domain
         hp_keys = []
         hp_values = []
-        for key, hp_range in reversed(self.config_space.items()):
+        for key, hp_range in reversed(list(self.config_space.items())):
             if isinstance(hp_range, Domain):
                 hp_keys.append(key)
                 hp_values.append(hp_range.categories)
--- a/syne_tune/optimizer/schedulers/searchers/searcher.py
+++ b/syne_tune/optimizer/schedulers/searchers/searcher.py
@@ -12,6 +12,7 @@
 # permissions and limitations under the License.
 import logging
 import numpy as np
+from random import shuffle
 from typing import Dict, Optional, List, Tuple
 
 from syne_tune.config_space import Domain, is_log_space, Categorical
@@ -474,24 +475,55 @@ class RandomSearcher(SearcherWithRandomSeed):
 
 
 class GridSearcher(BaseSearcher):
+    """Searcher which generate all the configurations on search space grids
+    defined by config_space and suggest the next one to try.
+    It will first suggest those specified in points_to_evalutate,
+    then the rest of the configurations on the grids
+
+    Note: GridSearcher does not support `debug_log` anymore.
+    Instead, using `logging.debug` which is the python standard way of logging
+    debug information.
+
+    Parameters
+    ----------
+    config_space : Dict
+        The configuration space that defines search space grids. It contains
+        the full specification of the Hyperparameters, and the configurations
+        generated is the combinations of these Hyperparameters.
+    metric : str
+        Name of metric passed to update.
+    points_to_evaluate : List[Dict] or None
+        List of configurations to be evaluated initially (in that order).
+        Each config in the list can be partially specified,
+        or even be an empty dict.
+        Each specified hyperparameter must be within the config space. That is,
+        all the configurations in this list should be on the search space grids.
+        For each hyperparameter not specified, the default value is determined
+        using a midpoint heuristic.
+        If None (default), this is mapped to [dict()], a single default config
+        determined by the midpoint heuristic. If [] (empty list), no initial
+        configurations are specified.
+    shuffle_config : bool
+        If True (default), the order of configurations suggested after those
+        specified in points_to_evalutate is shuffled. Otherwised, the order
+        will follow the Cartesian product of the configurations.
+    """
+
     def __init__(self, config_space, metric, points_to_evaluate=None, **kwargs):
         super().__init__(config_space, metric, points_to_evaluate)
+        self._validate_config_space(config_space)
         self._hp_ranges = make_hyperparameter_ranges(config_space)
         self._excl_list = ExclusionList.empty_list(self._hp_ranges)
-        self._config_candidates = self._generate_config_candidates(config_space)
-        # Debug log printing (switched on by default)
-        debug_log = kwargs.get("debug_log", True)
-        if isinstance(debug_log, bool):
-            if debug_log:
-                self._debug_log = DebugLogPrinter()
-            else:
-                self._debug_log = None
-        else:
-            assert isinstance(debug_log, DebugLogPrinter)
-            self._debug_log = debug_log
+        shuffle_config = kwargs.get("shuffle_config", True)
+        if not isinstance(shuffle_config, bool):
+            shuffle_config = True
+        self._shuffle_config = shuffle_config
+        self._remain_candidates_on_grid = self._generate_all_candidates_on_grid(
+            config_space
+        )
 
     def get_config(self, **kwargs):
-        """Sample a new configuration at the grid of config_space
+        """Sample a new configuration from search space grid defined by config_space.
         This is done without replacement, so previously returned configs are
         not suggested again.
 
@@ -500,13 +532,10 @@ class GridSearcher(BaseSearcher):
         A new configuration that is valid, or None if no new config can be
         suggested.
         """
-        trial_id = kwargs.get("trial_id")
-        if self._debug_log is not None:
-            self._debug_log.start_get_config("grid", trial_id=trial_id)
         new_config = self._next_initial_config()
         if new_config is None:
             while True:
-                _config = self._next_config_candidate()
+                _config = self._next_candidate_on_grid()
                 if _config is None:
                     # If searcher returns no config at all, the
                     # search space is exhausted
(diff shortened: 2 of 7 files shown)


iaroslav-ai:
This does not contain support for the HPs other than Categorical yet - either me or Jack will have a follow up PR to support rest of hp types.

syne_tune/optimizer/baselines.py
```diff
@@ -27,6 +27,16 @@ def __init__(self, config_space: Dict, metric: str, **kwargs):
         )
 
 
+class GridSearch(FIFOScheduler):
+    def __init__(self, config_space: Dict, metric: str, **kwargs):
```
geoalgo, line 31: Can you add comment on what the method does? (they are things to say in the design, how configurations are ordered and in particular that only categorical is supported for now)

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -470,3 +472,113 @@ def clone_from_state(self, state: dict):
     @property
     def debug_log(self):
         return self._debug_log
+
+
+class GridSearcher(BaseSearcher):
+    def __init__(self, config_space, metric, points_to_evaluate=None, **kwargs):
+        super().__init__(config_space, metric, points_to_evaluate)
+        self._hp_ranges = make_hyperparameter_ranges(config_space)
```
geoalgo, line 480: If we only support categorical, we should assert it.

geoalgo (commented):
Can you also add tests? E.g. testing that your searcher returns the right list of configurations given a predefined search-space for instance and that it can be updated with results without errors.

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -470,3 +472,113 @@ def clone_from_state(self, state: dict):
...
+
+    def _generate_config_candidates(self, config_space: Dict) -> List[Dict]:
+        # Validates and get all hp values
+        hp_values = []
+        for hp_range in config_space.values():
+            assert isinstance(hp_range, Categorical)
+            hp_values.append(hp_range.categories)
+
+        # Generate all combinations of hp values
+        hp_values_combinations = list(product(*hp_values))
```
geoalgo, line 547: Can we offer the option to shuffle this list? We will typically not want to rely on the arbitrary order of parameters encoding and we will want to benchmark grid-search with multiple seeds.

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -470,3 +472,113 @@ def clone_from_state(self, state: dict):
...
     def debug_log(self):
         return self._debug_log
+
+
+class GridSearcher(BaseSearcher):
+    def __init__(self, config_space, metric, points_to_evaluate=None, **kwargs):
+        super().__init__(config_space, metric, points_to_evaluate)
+        self._hp_ranges = make_hyperparameter_ranges(config_space)
+        self._excl_list = ExclusionList.empty_list(self._hp_ranges)
+        self._config_candidates = self._generate_config_candidates(config_space)
```
mseeger, line 482: This could potentially become pretty big. It is probably OK, but a nicer solution would create the grid entries on demand.

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -470,3 +472,113 @@ def clone_from_state(self, state: dict):
...
+    def _next_config_candidate(self) -> Optional[Dict]:
+        if self._config_candidates:
+            return self._config_candidates.pop(0)
+        else:
+            return None  # No more grids
+
+    def get_state(self) -> dict:
+        state = dict(
+            super().get_state(),
+            config_candidates=self._config_candidates,
```
mseeger, line 565: Would be nice not to have to store this in the state. Can you not recreate it, say by just storing its size?

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -470,3 +472,113 @@ def clone_from_state(self, state: dict):
...
+            excl_list=self._excl_list,
```
mseeger, line 566: You are right to store excl_list in the state here, but in RandomSearcher we don't do this. It is probably OK not to store it, as it can be big.

mseeger (commented):
Please also link this into tst/test_schedulers.py, to make sure it works in principle.

I'd also request you to add some unit tests.

syne_tune/optimizer/schedulers/searchers/searcher.py
iaroslav-ai: Nit: `_config_candidates` is a bit generic and does not explain what those are. Per our discussion - could be `_grid_candidates` or `_all_candidates_on_grid` or something along the lines of this.

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -470,3 +472,113 @@ def clone_from_state(self, state: dict):
...
+        -------
+        A new configuration that is valid, or None if no new config can be
+        suggested.
+        """
+        trial_id = kwargs.get("trial_id")
+        if self._debug_log is not None:
+            self._debug_log.start_get_config("grid", trial_id=trial_id)
+        new_config = self._next_initial_config()
+        if new_config is None:
+            while True:
```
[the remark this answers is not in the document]
iaroslav-ai: Jack can say more - but for our purposes it makes it easier to restore the state of searcher with exclusion list; Also a bit easier to deal with initial configs. Let us know if you feel strongly about this.

syne_tune/optimizer/schedulers/searchers/searcher.py
duck105: Have added a `shuffle_config` option, default is True.

syne_tune/optimizer/schedulers/searchers/searcher.py
duck105: Rename the function as `_generate_all_candidates_on_grid()`, and class member to `_remain_candidates_on_grid`.
Since the scheduler will pop items from the list whenever `get_config()` is called, I name it begin with `_remain`

syne_tune/optimizer/schedulers/searchers/searcher.py
duck105: Right now I just stored the excl_list, not candidates anymore

syne_tune/optimizer/schedulers/searchers/searcher.py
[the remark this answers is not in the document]
duck105: Also, storing the `remain_candidates_on_grid` makes it easier to support the shuffle option.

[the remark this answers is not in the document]
geoalgo: I think it will be very inefficient for large cardinalities and that we should therefore avoid it. You can also store the index in the list of possibilities in addition to the list of excluded candidates and you wont have to do this reject sampling (nor store the list of candidates since it can be regenerated when you re-instantiate your state for instance).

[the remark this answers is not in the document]
iaroslav-ai: I see that `ExclusionList` which is used here is actually using `set` underneath; So adding / removing stuff from there should be ~constant. Looking at the code it appears that as long as GridSearcher is kept in memory getting new candidate should also be ~constant. What would be exact scenario for inefficiency here? Is it reconstruction of searcher from state? Unless we expect users to evaluate millions of candidates or more, overhead for that is unlikely to be noticeable, and is one time for the rest of GridSearcher being in memory.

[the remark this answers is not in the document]
geoalgo: The problem with the current approach is that the code will become very inefficient for large number of candidates both in term of time and space (and you can easily get many candidates since the number of possibilities grows exponentially with the number of hyperparameters). The current implementation stores all sampled candidates and performs reject sampling, so its runtime is O(n) for candidate generation and also O(n) for space whereas it could store only an index of current position in O(1) space and compute next configuration in O(1) time.

duck105:
Right now the order of configs to be enumerate is as follow:
1. Default (shuffle_config = True), the order is random
2. if shuffle_config = False, the hyperparameter specified first in the config space will be explore first.
For example: 
```
config_space = {'hp1': [1,2,3], 'hp2': [4,5], 'hp3': [6,7]}
order: (1, 4, 6), (2, 4, 6), (3, 4, 6), (1, 5, 6), (2, 5, 6), (3, 5, 6), (1, 4, 7), (2, 4, 7), (3, 4, 7), (1, 5, 7), (2, 5, 7), (3, 5, 7)
```

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -519,66 +548,90 @@ def get_config(self, **kwargs):
         if new_config is not None:
             # Add config to exclusion list so that it won't be suggested again
             self._excl_list.add(new_config)
-            if self._debug_log is not None:
-                self._debug_log.set_final_config(new_config)
-                # All get_config debug log info is only written here
-                self._debug_log.write_block()
+
+            # Write debug log for the the config
+            trial_id = kwargs.get("trial_id")
```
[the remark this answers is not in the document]
duck105: To be consistent with other searcher's get_config(), I didn't make it an argument. I use a different format of log if the trial_id is not specified to prevent error.

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -470,3 +473,165 @@ def clone_from_state(self, state: dict):
...
+            config = self.get_config(**kwargs)
+            if config is not None:
+                configs.append(config)
+        return configs
+
+    def _validate_config_space(self, config_space: Dict):
+        # GridSearcher now only support Categorical parameters
+        for hp_range in config_space.values():
+            if isinstance(hp_range, Domain):
+                assert isinstance(hp_range, Categorical)
```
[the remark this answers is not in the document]
duck105: I add a note in the comment of GridSearcher class.

syne_tune/optimizer/schedulers/searchers/searcher.py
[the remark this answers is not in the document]
duck105: I've removed the usage of ExclusionList. Instead, using `_remain_candidates_on_grid` (will be an uniq list) to generate next config, store/restore the searcher

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -470,3 +473,174 @@ def clone_from_state(self, state: dict):
...
+            config_space=self.config_space,
+            metric=self._metric,
+            shuffle_config=self._shuffle_config,
+        )
+        new_searcher._restore_from_state(state)
+        return new_searcher
+
+    def _restore_from_state(self, state: dict):
+        super()._restore_from_state(state)
+        self._remain_candidates_on_grid = state["remain_candidates"].copy()
```
iaroslav-ai, line 642: Minor - consider renaming to `remaining_candidates`.

iaroslav-ai:
Left one minor comment, but overall LGTM. We would need for someone from core maintainers to review.

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -470,3 +473,174 @@ def clone_from_state(self, state: dict):
...
+        all_candidates_on_grid = []
+        for values in hp_values_combinations:
+            all_candidates_on_grid.append(dict(zip(hp_keys, values)))
+        return all_candidates_on_grid
+
+    def _next_candidate_on_grid(self) -> Optional[Dict]:
+        if self._remaining_candidates:
+            return self._remaining_candidates.pop(0)
+        else:
+            # No more candidate
```
aaronkl, line 621: candidate -> candidates

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -470,3 +473,174 @@ def clone_from_state(self, state: dict):
...
+            excl_list.add(candidate)
+        remaining_candidates = []
+        for candidate in self._generate_all_candidates_on_grid():
+            if not excl_list.contains(candidate):
+                remaining_candidates.append(candidate)
+                excl_list.add(candidate)
+        return remaining_candidates
+
+    def _generate_all_candidates_on_grid(self) -> List[Dict]:
+        # Get hp values that is specified as Domain
```
aaronkl, line 600: is -> are

Issue #275, opened 2022-06-23
Custom results directory


martinferianc:
Dear creators, thank you again for your great work and perhaps sorry for being annoying with my suggestions/questions. Is it possible to change the home directory of different runs such that it is not `~/syne-tune` but a custom path? Thanks!

martinferianc:
Sure that could work, ideally if it could be set as an attribute in the tuner that would be the best I guess. Thank you!

geoalgo:
Ok thanks, we will work on this. The reason why I would prefer an environment variable is that there are many places where we access local files (launching tuning, getting experiments, downloading tabular datasets, ...) and it would probably be hard set them properly all at the same time.

geoalgo:
Thanks for the idea! We added support for this, see for instance the description in FAQ:
https://github.com/awslabs/syne-tune/blob/main/docs/faq.md#-how-can-i-change-the-default-output-folder-where-tuning-results-are-stored

ChongqiQu:
> Hi David,
> 
> ```
> export SYNETUNE_PATH="/home/martin/my-syne-tune-path/"
> python tuner.py
> ```
> Is this usage no longer valid?

geoalgo:
Hi @ChongqiQu , apologies for the delay I was not available this week, I will take a look next week.

geoalgo:
Hi @ChongqiQu , 

Using custom directory is still supported and is described in our FAQ

https://syne-tune.readthedocs.io/en/latest/faq.html#how-can-i-change-the-default-output-folder-where-tuning-results-are-stored
For instance, the following should work:
```bash
export SYNETUNE_FOLDER="~/new-syne-tune-folder"
python examples/launch_height_baselines.py
```

(the commands you are mentioning are just an illustration I gave to Martin)

ChongqiQu:
Thank you for the clarification! I appreciate you pointing me to the FAQ.

I've retested it and it does work using these two usages.There may be some problems with my previous tests.

### Another implementation method of using custom directory

`tuner_path` can be set in `tuner.py` through the following code:
```
self.tuner_path = Path(experiment_path(tuner_name=self.name, local_path="~/new-syne-tune-folder"))
```
Here we can set `local_path` to get custom directory. But  There is no initialization attribute of `Class Tuner` . I hope `local_path` can be used as an initialization parameter that needs to be passed in 'Class Tuner'. So that we can set it arbitrarily in the tuner.

Just like this:
```
class Tuner:
    def __init__(
[...]

PR #285, opened 2022-06-28
Add option to override syne-tune path with environment variable

geoalgo:
*Issue #, if available:* https://github.com/awslabs/syne-tune/issues/275

*Description of changes:* 


syne_tune/constants.py
```diff
@@ -15,7 +15,8 @@
 benchmarks.
 """
 
-SYNE_TUNE_FOLDER = "syne-tune"
+SYNE_TUNE_ENV_FOLDER = "SYNETUNE_FOLDER"  # environment variable that allows to overides default library folder
```
aaronkl, line 18: This might be a bit confusing. Suggestion SYNE_TUNE_FOLDER = "SYNE_TUNE_FOLDER_DEFAULT", "SYNE_TUNE_FOLDER_DEFAULT" = "syne-tune"

aaronkl (dismissed):
LGTM

syne_tune/constants.py
geoalgo: Thanks, I used your suggestion for the first one. For the second one, think it makes sense to keep `_ENV` to indicate that this is an environment variable.

geoalgo:
Hi @aaronkl can you approve again? (I addressed your proposed naming change)

aaronkl (approved):

PR #291, opened 2022-07-08
Control Decision for Running Trials When Stopping Criterion is Met

wistuba:
*Description of changes:* An argument to the Tuner was added which allows to control how to deal with running trials when the stopping criterion is met.

### DIFF
--- a/syne_tune/tuner.py
+++ b/syne_tune/tuner.py
@@ -53,7 +53,7 @@ class Tuner:
         max_failures: int = 1,
         tuner_name: Optional[str] = None,
         asynchronous_scheduling: bool = True,
-        stop_immediately_when_stopping_criterion_met: bool = True,
+        wait_trial_completion_when_stopping: bool = True,
         callbacks: Optional[List[TunerCallback]] = None,
         metadata: Optional[Dict] = None,
         suffix_tuner_name: bool = True,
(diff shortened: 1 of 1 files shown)


syne_tune/tuner.py
```diff
@@ -76,6 +77,8 @@ def __init__(
         :param asynchronous_scheduling: whether to use asynchronous scheduling when scheduling new trials. If `True`,
         trials are scheduled as soon as a worker is available, if `False`, the tuner waits that all trials are finished
          before scheduling a new batch.
+        :param stop_immediately_when_stopping_criterion_met: how to deal with running trials when stopping criterion is
```
geoalgo, line 80: The name is not ideal since we do not stop immediately when stopping condition is met (we do some clean-ups such as stopping running trials to avoid wasting ressource, save last result etc). Perhaps, `wait_trial_completion` or `wait_trial_completion_when_stopping`? (or do you have another one?)

geoalgo (commented):
LGTM, we should just rename the option given that it is a bit ambiguous.

syne_tune/tuner.py
wistuba: done

geoalgo (approved):
Could you also add a test for this mode? (otherwise it can easily break, it can be done in another PR)
One way to run tuning with one worker for a very small stop-condition and make sure that all the results of the trials are emitted.

PR #298, opened 2022-08-01
remove duplicate function for blackbox generation, add message printi…

geoalgo:
…ng reference to cite when generating them for the first time, log blackbox metadata, use cumulative time consistently in blackboxes

*Description of changes:* This is a clean-up PR on blackboxes that removes some duplication for user when writing recipes to generate them from scratch with a parent class.

In addition, this PR completely removes the "time_ressource_attr" and just use cumulative time everywhere as it is just simpler from a user perspective. 

Because we log the metadata, we would also be able to load the blackboxes without needing to specify what are columns for error/runtime (would be done in a separate PR to avoid making too many changes at once).

Finally load is replace by load_blackbox as load is used in lot of frameworks and is probably too general.


syne_tune/blackbox_repository/conversion_scripts/BlackboxRecipe.py
```diff
@@ -0,0 +1,46 @@
+import logging
```
wistuba, line 1: change file name to blackbox_recipe.py

syne_tune/blackbox_repository/conversion_scripts/scripts/icml2020_import.py
```diff
@@ -113,16 +112,30 @@ def serialize_xgboost():
...
+class XGBoostRecipe(BlackboxRecipe):
+    def __init__(self):
+        super(XGBoostRecipe, self).__init__(
+            name="icml-xgboost",
+            cite_reference="A quantile-based approach for hyperparameter transfer learning."
+            "Salinas, D., Shen, H., and Perrone, V. 2021.",
+        )
 
+    def _generate_on_disk(self):
+        serialize_xgboost()
```
wistuba, line 124: does it make sense to move these functions into the recipe?

wistuba (approved):

wistuba (approved):

PR #308, opened 2022-08-08
Fix for wait_trial_completion

wistuba:
There was a bug in the logic. Furthermore, disabled it by default.

### DIFF
--- a/benchmarking/__init__.py
+++ b/benchmarking/__init__.py
@@ -1,0 +1,12 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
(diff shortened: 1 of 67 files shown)


syne_tune/tuner.py
```diff
@@ -168,7 +168,11 @@ def run(self):
             config_space_exhausted = False
             stop_condition_reached = self._stop_condition()
 
-            while not stop_condition_reached:
+            while (
```
geoalgo, line 171: Could you name the boolean variable and add a comment explaining the logic? For instance:
```suggestion
            # we stop when either the stop condition is reached or when all trials are done if the wait_trial_completion is activated
            should_stop = stop_condition_reached or self.wait_trial_completion_when_stopping and len(running_trials_ids) > 0
            while should_stop:
```

geoalgo (commented):
LGTM, only have a a small comment on the readability.

geoalgo:
FYI, something seems off with your latest commit as the diff is empty.

geoalgo:
I see the other PR now, you can ignore my last comment.

PR #314, opened 2022-08-15
Differential Evolution Hyperband (DEHB)

mseeger:
*Issue #, if available:*

*Description of changes:*


syne_tune/optimizer/schedulers/synchronous/dehb.py
```diff
@@ -0,0 +1,736 @@
...
+        self.mode = kwargs["mode"]
+        self.max_resource_attr = kwargs.get("max_resource_attr")
+        self._resource_attr = kwargs["resource_attr"]
+        self.mutation_factor = kwargs["mutation_factor"]
+        self.crossover_probability = kwargs["crossover_probability"]
+        self._support_pause_resume = kwargs["support_pause_resume"]
+        # Generator for random seeds
+        random_seed = kwargs.get("random_seed")
+        if random_seed is None:
+            random_seed = np.random.randint(0, 2**32)
```
wistuba, line 244: `get` allows you to provide a default value as the second param: `get("random_seed", np.random.randint(0, 2**32))`.

syne_tune/optimizer/schedulers/synchronous/hyperband_bracket.py
```diff
@@ -91,16 +77,17 @@ def assert_check_rungs(rungs: List[Tuple[int, int]]):
             [-x for x in sizes]
         ), f"Rung sizes {sizes} are not decreasing"
 
-    def is_bracket_complete(self) -> bool:
-        return self.current_rung >= self.num_rungs
-
     @property
     def num_rungs(self) -> int:
-        return len(self._rungs)
+        raise NotImplementedError
```
wistuba, line 82: not sure I understand. is this an abstract class or why don't we implement this function anymore? If yes, I'd make this class abstract as well.

syne_tune/optimizer/schedulers/synchronous/hyperband_impl.py
```diff
@@ -61,9 +63,9 @@ class SynchronousGeometricHyperbandScheduler(SynchronousHyperbandScheduler):
         :class:`FIFOScheduler`. Must be positive int larger than
         `grace_period`. If this is not given, it is inferred like in
         :class:`FIFOScheduler`.
-    brackets : int
-        Number of brackets to be used. The value 1 corresponds to successive
-        halving. Is capped to the largest number of supported brackets.
+    brackets : int (optional)
+        Number of brackets to be used. The default is to use the maximum
```
wistuba, line 67: what is "maximum number of brackets"? this seems to be a breaking API change. Is there a way how we communicate this?

syne_tune/optimizer/schedulers/synchronous/hyperband_rung_system.py
```diff
@@ -98,10 +105,10 @@ def geometric(
...
             rungs.append((int(np.ceil(pre_fact)), max_resource))
             rung_systems.append(rungs)
-        parts = [
-            f"Bracket {i}: rungs = {rungs}" for i, rungs in enumerate(rung_systems)
-        ]
-        logger.info("\n".join(parts))
+        # parts = [
+        #     f"Bracket {i}: rungs = {rungs}" for i, rungs in enumerate(rung_systems)
+        # ]
+        # logger.info("\n".join(parts))
```
wistuba, line 111: delete?

tst/test_schedulers.py
```diff
@@ -134,6 +143,14 @@ def test_sync_scheduler(searcher, mode):
...
         random_seed=random_seed,
     )
+    if scheduler == "hb":
+        myscheduler = SynchronousGeometricHyperbandScheduler(
+            config_space, **scheduler_kwargs
+        )
+    else:
+        myscheduler = GeometricDifferentialEvolutionHyperbandScheduler(
+            config_space, **scheduler_kwargs
+        )
```
wistuba, line 153: nit: you could add the class names directly into parametrize and then you would not need to have this distinction here

wistuba (commented):
General comment, not limited to this PR: maybe we can have a general benchmarking script. It does not seem scalable if many new optimizers come with their own benchmarking script. At least we should make sure not to copy the same code over and over again. Some code reuse would be great.

mseeger:
I am all for a general benchmarking script, but it may be difficult, because it would come with many CL args.

syne_tune/optimizer/schedulers/synchronous/dehb.py
mseeger: This is a tricky one. If kwargs contains 'random_seed', but the value is None, get returns None. It only returns the default if the key is not there.

syne_tune/optimizer/schedulers/synchronous/hyperband_bracket.py
mseeger: Yes, SynchronousBracket is abstract. I am avoiding using abc, abstractmethod, etc, because this has downsides (I had serialization errors), and I find abc to be overkill, it is enough to just raise NotImplementedError.

syne_tune/optimizer/schedulers/synchronous/hyperband_impl.py
mseeger: I think this just changed the docstring to make it exact. Also in the paper, if you do not pass brackets, it will be set to the max value possible, which is the total number of rungs.

The first bracket always contains all rungs, and subsequent ones always have one rung less (they drop the front ones).

This is the default for *synchronous* HB. For async HB, the default is to have 1 bracket. This is because empirically, async works quite well with 1 bracket (even though maybe Hyper-Tune may do better), but sync HB works better with >1 bracket.

syne_tune/optimizer/schedulers/synchronous/hyperband_rung_system.py
mseeger: ok

tst/test_schedulers.py
mseeger: ok

syne_tune/optimizer/schedulers/synchronous/hyperband_impl.py
wistuba: In L35 you removed the default bracket value of 1, now changing it to a different value. I'm all open for that but from my own experience it is a bit annoying if defaults keep changing.

wistuba (approved):
I'm not super familiar with the Hyperband code but from what I understand, it looks ok.

syne_tune/optimizer/schedulers/synchronous/hyperband_impl.py
mseeger: You are right, I was missing that.

geoalgo (commented):
There are a few files/classes that seems duplicated in this PR but could be imported instead, for instance `MethodArguments`, all methods in `methods` dict but DEHB, benchmark definitions, `benchmark_main.py`, `launch_remote.py`. 

Would it be possible to import the definitions/classes when possible to avoid duplicating them?

mseeger:
OK. The definitions are somewhat different. What I can do is to extend MethodArguments in benchmark_automl, and then use the extended one in benchmark_dehb.

mseeger:
But I will not import Methods from benchmark_automl, because it is too different. For example, BOHB means async BOHB in automl, but sync BOHB in dehb.

mseeger:
OK: #321

Issue #304, opened 2022-08-05
Promotion Logic Bug


wistuba:
There seems to be a problem with the Hyperband promotion logic.

How to reproduce:
Add `type="promotion"` to https://github.com/awslabs/syne-tune/blob/main/benchmarking/nursery/benchmark_automl/baselines.py#L69

Run `python benchmarking/nursery/benchmark_automl/benchmark_main.py --num_seeds 1 --method ASHA --benchmark lcbench-airlines`

```Traceback (most recent call last):
  File "/syne-tune/benchmarking/nursery/benchmark_automl/benchmark_main.py", line 209, in <module>
    tuner.run()
  File "/syne-tune/syne_tune/tuner.py", line 240, in run
    raise e
  File "/syne-tune/syne_tune/tuner.py", line 175, in run
    new_done_trial_statuses, new_results = self._process_new_results(
  File "/syne-tune/syne_tune/tuner.py", line 345, in _process_new_results
    done_trials_statuses = self._update_running_trials(
  File "/syne-tune/syne_tune/tuner.py", line 465, in _update_running_trials
    decision = self.scheduler.on_trial_result(trial=trial, result=result)
  File "/syne-tune/syne_tune/optimizer/schedulers/hyperband.py", line 779, in on_trial_result
    task_info = self.terminator.on_task_report(trial_id, result)
  File "/syne-tune/syne_tune/optimizer/schedulers/hyperband.py", line 1124, in on_task_report
    rung_sys.on_task_report(trial_id, result, skip_rungs=skip_rungs)
  File "/syne-tune/syne_tune/optimizer/schedulers/hyperband_promotion.py", line 221, in on_task_report
    assert resource == milestone, (
AssertionError: trial_id 1: resource = 4 > 3 milestone. Make sure to report time attributes covering all milestones```

mseeger:
Hi Martin,

when testing DEHB, I also ran into an issue like this, for FCNet. It does not happen when you use max_resource_attr and pass it to the simulator backend as well, but this is of course not a proper fix.

Can we sync, and you tell me what you observed so far?

mseeger:
I can give this a try

mseeger:
```INFO:syne_tune.optimizer.schedulers.searchers.bayesopt.utils.debug_log:[0: random]
num_layers: 3
max_units: 181
batch_size: 91
learning_rate: 0.003162277660168382
weight_decay: 0.050005
momentum: 0.545
max_dropout: 0.5

[...]

PR #317, opened 2022-08-18
Fix bug in _BlackboxSimulatorBackend for pause/resume scheduling (iss…

mseeger:
…ue #304)

*Issue #, if available: #304

*Description of changes:*
Fixes bug which affects pause-and-resume scheduling (ASHA promotion, sync HB) with blackbox backend. The fix requires a small extension of the TrialScheduler API: pause_trial needs to receive the result which led the scheduler to make the pause decision. I left this optional, if the result is not passed, the simulated backend will behave differently as intended (namely, resumed trials will be started from scratch). But the Tuner can easily pass this information. To  me, this API extension makes sense: give the backend information about why a trial has been paused.

For API consistency, I did the same extension for stop_trial, even though this is not needed right now. Note that these changes are internal API between backend and Tuner, the user does not notice this at all.


syne_tune/backend/local_backend.py
```diff
@@ -229,7 +229,7 @@ def _all_trial_results(self, trial_ids: List[int]) -> List[TrialResult]:
             res.append(trial_results)
         return res
 
-    def _pause_trial(self, trial_id: int):
+    def _pause_trial(self, trial_id: int, result: Optional[dict]):
```
wistuba, line 232: there are multiple cases where the typing suggests optional but `None` was not provided as default value.

wistuba (approved):
Only one minor comment

syne_tune/backend/local_backend.py
mseeger: These are internal methods, you'd not call them directly.

PR #318, opened 2022-08-20
LCBench Fix

wistuba:
Drop first and last epoch from the current LCBench version. Add further metrics.

### DIFF
--- a/syne_tune/blackbox_repository/conversion_scripts/scripts/lcbench/lcbench.py
+++ b/syne_tune/blackbox_repository/conversion_scripts/scripts/lcbench/lcbench.py
@@ -41,7 +41,7 @@ METRIC_ELAPSED_TIME = "time"
 
 RESOURCE_ATTR = "epoch"
 
-MAX_RESOURCE_LEVEL = 52
+MAX_RESOURCE_LEVEL = 50
 
 CONFIGURATION_SPACE = {
     "num_layers": randint(1, 5),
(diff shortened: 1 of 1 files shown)


syne_tune/blackbox_repository/conversion_scripts/scripts/lcbench/lcbench.py
```diff
@@ -41,7 +41,7 @@
 
 RESOURCE_ATTR = "epoch"
 
-MAX_RESOURCE_LEVEL = 52
+MAX_RESOURCE_LEVEL = 50
```
mseeger, line 44: This naming is wrong. The max_resource_level is 51, right? It was wrong before as well, I suppose.

mseeger (approved):
Just one little thing to fix.

syne_tune/blackbox_repository/conversion_scripts/scripts/lcbench/lcbench.py
wistuba: The data has 52 epochs. Based on the notebook, the first epoch is before training started. Furthermore, they seem to discard the last epoch as well. The paper and the [repo](https://github.com/automl/LCBench) clearly state that they have 50 epochs. Therefore, I've dropped first and last data point which means we have 50 epochs in total. Does that make sense to you?

mseeger: Ah I see, you are right.

mseeger (approved):

PR #320, opened 2022-08-26
Clean up imports. Now, examples/launch_height_baselines.py runs witho…

mseeger:
…ut any AWS dependencies (only requirements_noaws.txt)

*Issue #, if available:*

*Description of changes:*


mseeger:
One thing missing here: Installing syne_tune[no_aws] installs both core + no_aws, since core is always installed. It would be good to allow users to call setup.py with the no_aws dependencies only.

mseeger:
First (necessary) step for integration into AutoGluon

syne_tune/backend/sagemaker_backend/sagemaker_backend.py
```diff
@@ -326,3 +326,7 @@ def set_path(
     ):
         # we use the tuner-name to set the checkpoint directory
         self.tuner_name = tuner_name
+
+    def on_tuner_save(self):
+        # Re-initialize the session after `Tuner` is serialized
+        self.initialize_sagemaker_session()
```
geoalgo, line 332: Could you take this change out? It seems not relevant to the import, also reinitializing the session at all saving could have undesirable effects.

syne_tune/blackbox_repository/blackbox_surrogate.py
```diff
@@ -12,25 +12,19 @@
 # permissions and limitations under the License.
 from typing import Optional, Dict
 import pandas as pd
-from sklearn.neighbors import KNeighborsRegressor
-from sklearn.pipeline import Pipeline, FeatureUnion, make_pipeline
-from sklearn.preprocessing import OneHotEncoder, StandardScaler
-from sklearn.base import BaseEstimator, TransformerMixin
```
geoalgo, line 18: Moving all imports into functions should be avoided (see [PEP8](https://peps.python.org/pep-0008/)), could you revert the changes that move import into functions there and bellow?

tst/blackbox_repository/test_blackbox.py
```diff
@@ -138,19 +139,22 @@ def test_blackbox_seed():
 
 
 def test_blackbox_offline_serialization():
-    y = x1 * x2
-    data = np.stack([x1, x2, y]).T
-    df = pd.DataFrame(data=data, columns=["hp_x1", "hp_x2", "metric_rmse"])
+    try:
```
geoalgo, line 142: This would be undesirable as the test will silently fail if this dependency is not installed, could you revert it? (also holds for the unit-test bellow)

syne_tune/tuning_status.py
```diff
@@ -16,11 +16,7 @@
 from collections import defaultdict, OrderedDict
 from typing import List, Dict, Tuple
 
-import pandas as pd
-
 from syne_tune.backend.trial_status import Status, Trial
-import numpy as np
```
geoalgo, line 22: I am not sure why we would remove this since we assume we have numpy in all our dependencies?

requirements-noaws.txt
```diff
@@ -0,0 +1,7 @@
+numpy>=1.16.0
```
geoalgo, line 1: This duplicates `requirements.txt`, to avoid this it would be better to have instead a `requirement-aws.txt` that just contains the aws dependencies.

geoalgo (changes requested):
There are a few things that I think we should address before merging this PR (see comments bellow). The first one is the duplication of requirements, the second is the import that are put  inside the classes and the third is to not let test silently fail if imports are not found.

requirements-noaws.txt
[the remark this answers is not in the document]
mseeger: Hi David, that would be my favorite. But can we configure things so that the default (used by pip install) is to install with AWS? So that installation without AWS is a special case for developers?

mseeger:
OK, I will redo this one

PR #321, opened 2022-08-30
Unify some benchmarking code

mseeger:
*Issue #, if available:*

*Description of changes:*


benchmarking/nursery/benchmark_automl/baselines.py
```diff
@@ -32,14 +32,17 @@
 
 @dataclass
 class MethodArguments:
-    config_space: Dict
+    config_space: dict
```
geoalgo, line 35: This changes seems unintended since Dict is the type.

benchmarking/nursery/benchmark_automl/benchmark_definitions.py
```diff
@@ -37,18 +38,34 @@ def fcnet_benchmark(dataset_name):
...
+    "cifar100": 6 * 3600,
+    "ImageNet16-120": 8 * 3600,
+}
+
+
+NAS201_N_WORKERS = {
+    "cifar10": 4,
+    "cifar100": 4,
+    "ImageNet16-120": 8,
+}
```
geoalgo, line 56: This changes the original definition, could you let the initial values here and update only those in the dehb benchmark? e.g.:
```suggestion
NAS201_MAX_WALLCLOCK_TIME = {
    "cifar10": 6 * 3600,
    "cifar100": 6 * 3600,
    "ImageNet16-120": 6 * 3600,
}


NAS201_N_WORKERS = {
    "cifar10": 4,
    "cifar100": 4,
    "ImageNet16-120": 4,
}
```

benchmarking/nursery/benchmark_dehb/benchmark_main.py
```diff
@@ -10,144 +10,10 @@
...
-            tuner_name=experiment_tag,
-            metadata={
-                "seed": seed,
-                "algorithm": method,
-                "tag": experiment_tag,
-                "benchmark": benchmark_name,
-            },
-        )
-        tuner.run()
+    main(methods)
```
geoalgo, line 19: Thanks for removing those duplicates.

benchmarking/nursery/benchmark_automl/baselines.py
mseeger: I am general trying to change Dict to dict when nothing else is said, like Dict[str, int] or so. Same for List -> list, or Tuple -> tuple.

geoalgo: Makes sense.

benchmarking/nursery/benchmark_automl/benchmark_main.py
```diff
@@ -95,10 +91,12 @@ def subsample(
     return transfer_learning_evaluations
 
 
-if __name__ == "__main__":
+def parse_args(methods: dict, benchmark_definitions: dict):
     parser = ArgumentParser()
     parser.add_argument(
-        "--experiment_tag", type=str, required=False, default=generate_slug(2)
```
geoalgo, line 101: I missed this one in my first review, I will be keen to keep the same behavior (allowing to not pass an experiment name) as it is not preventing from passing an explicit name. 

I can revert this specific one in another PR.

geoalgo (approved):

PR #322, opened 2022-08-30
New surrogate model for MOBSTER: Independent GPs at each rung level

mseeger:
*Issue #, if available:*

*Description of changes:*
Includes major refactoring to simplify Gaussian process code. This is the first part of the Hyper-Tune PR, second part yet to come.


syne_tune/optimizer/schedulers/searchers/bore/multi_fidelity_bore.py
```diff
@@ -15,13 +15,16 @@
...
+    DefaultHyperbandBracketSamplingSearcher,
+)
 
 __all__ = ["MultiFidelityBore"]
 
 logger = logging.getLogger(__name__)
 
 
-class MultiFidelityBore(Bore):
+class MultiFidelityBore(Bore, DefaultHyperbandBracketSamplingSearcher):
```
wistuba, line 27: do we use multi-inheritance now? please be careful with this, calling super.init is not trivial based on my experience.

syne_tune/optimizer/schedulers/searchers/gp_fifo_searcher.py
```diff
@@ -629,6 +629,9 @@ def configure_scheduler(self, scheduler):
             scheduler, FIFOScheduler
         ), "This searcher requires FIFOScheduler scheduler"
         super().configure_scheduler(scheduler)
+        # Allow model factory to depend on `scheduler` as well
+        model_factory = self.state_transformer.model_factory
+        model_factory.configure_scheduler(scheduler)
```
wistuba, line 634: nit: same line

tst/schedulers/bayesopt/gpautograd/test_gp.py
```diff
@@ -55,11 +57,12 @@ def f(x):
...
     y_train_np_ndarray = anp.array(y_train)
-    x_train_np_ndarray = anp.array(x_train)
-    x_test_np_ndarray = anp.array(x_test)
+    x_train_np_ndarray = anp.array(x_train).reshape((-1, 1))
+    x_test_np_ndarray = anp.array(x_test).reshape((-1, 1))
 
     model = GaussianProcessRegression(kernel=Matern52(dimension=1))
-    model.fit(x_train_np_ndarray, y_train_np_ndarray)
+    data = {"features": x_train_np_ndarray, "targets": y_train_np_ndarray}
+    model.fit(data)
```
wistuba, line 65: what's the motivation for this API change?

wistuba (commented):
Main points I have:
- Introduction of multi-inheritance is something we discussed earlier and agreed on using as little as possible. Not sure why we want to use it here.
- The API change from fit(X, y) to fit({"X": X, "y": y}) seems strange to me.

syne_tune/optimizer/schedulers/searchers/bore/multi_fidelity_bore.py
mseeger: Do you know a safer alternative? I'd like special searchers to provide another method `distribution_over_brackets`, this is done in `SearcherWithDistributionOverBrackets`. Then, I provide some default implementation in `DefaultHyperbandBracketSamplingSearcher`, which ends being used in quite some searchers.

If this can be done better, please let me know.

But I've been quite careful, in that all classes involved along the `SearcherWithDistributionOverBrackets` always just have the same `__init__` signature as the base class, and just call the super class init. This seems to me to avoid pitfalls.

tst/schedulers/bayesopt/gpautograd/test_gp.py
mseeger: We have some models (in `gpautograd/learncurve`, where the data is not just represented by X and y, but by a dictionary. The whole interplay between likelihood and model (and optimizer) is based on this simplification.

I could in GaussianProcessRegression still also support the previous style, but this would be a bit confusing. All in all, this is an internal API.

What do you think?

mseeger: I could also call the new style `fit_from_dict`, and reserve `fit` for the old style?

wistuba: I don't know the code very well but is this extreme flexibility required there or do you always use the same args some of which are optional?

syne_tune/optimizer/schedulers/searchers/bore/multi_fidelity_bore.py
wistuba: based on my understanding of your code, you might be able to use a delegator pattern instead

tst/schedulers/bayesopt/gpautograd/test_gp.py
mseeger: Things are a bit messy for these models, because they support pre-computations of additional info depending on the data.

syne_tune/optimizer/schedulers/searchers/bore/multi_fidelity_bore.py
mseeger: OK I see, this would be another class just for the purpose of computing this distribution. I can do it that way, I think.

tst/schedulers/bayesopt/gpautograd/test_gp.py
wistuba: I see. If they always take e.g. X and y, I would suggest to use fit(X, y, metadata) or similar. Otherwise I'm also fine with the current solution. Problem with passing dict is that it is very difficult to understand the API.

syne_tune/optimizer/schedulers/searchers/bore/multi_fidelity_bore.py
wistuba: Not sure you need another class. You might just create an instance of the one you have right now and call its function instead.

wistuba (approved):

PR #329, opened 2022-08-31
Remove AWS from core dependencies. Make sure local searchers still work

mseeger:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/syne_tune/backend/sagemaker_backend/sagemaker_backend.py
+++ b/syne_tune/backend/sagemaker_backend/sagemaker_backend.py
@@ -326,3 +326,7 @@ class SageMakerBackend(TrialBackend):
     ):
         # we use the tuner-name to set the checkpoint directory
         self.tuner_name = tuner_name
+
+    def on_tuner_save(self):
+        # Re-initialize the session after `Tuner` is serialized
+        self.initialize_sagemaker_session()
--- a/syne_tune/blackbox_repository/blackbox_surrogate.py
+++ b/syne_tune/blackbox_repository/blackbox_surrogate.py
@@ -33,6 +33,13 @@ class Columns(BaseEstimator, TransformerMixin):
         return X[self.names]
 
 
+def _default_surrogate(surrogate):
+    if surrogate is not None:
+        return surrogate
+    else:
+        return KNeighborsRegressor(n_neighbors=1)
+
+
 class BlackboxSurrogate(Blackbox):
     def __init__(
         self,
@@ -71,9 +78,7 @@ class BlackboxSurrogate(Blackbox):
         )
         assert len(X) == len(y)
         # todo other types of assert with configuration_space, objective_names, ...
-        self.surrogate = (
-            surrogate if surrogate is not None else KNeighborsRegressor(n_neighbors=1)
-        )
+        self.surrogate = _default_surrogate(surrogate)
         self.max_fit_samples = max_fit_samples
         self.fit_surrogate(
             X=X, y=y, surrogate=surrogate, max_samples=self.max_fit_samples
@@ -139,9 +144,7 @@ class BlackboxSurrogate(Blackbox):
         blackbox configuration. Possible example: KNeighborsRegressor(n_neighbors=1), MLPRegressor() or any estimator
         obeying Scikit-learn API.
         """
-        self.surrogate = (
-            surrogate if surrogate is not None else KNeighborsRegressor(n_neighbors=1)
-        )
+        self.surrogate = _default_surrogate(surrogate)
 
         self.surrogate_pipeline = self.make_model_pipeline(
             configuration_space=self.configuration_space,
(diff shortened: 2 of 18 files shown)


syne_tune/backend/sagemaker_backend/sagemaker_backend.py
```diff
@@ -326,3 +326,7 @@ def set_path(
     ):
         # we use the tuner-name to set the checkpoint directory
         self.tuner_name = tuner_name
+
+    def on_tuner_save(self):
+        # Re-initialize the session after `Tuner` is serialized
+        self.initialize_sagemaker_session()
```
geoalgo, line 332: Could you take this change out? It is not relevant to the import, also reinitializing the session at all saving could have undesirable effects.

syne_tune/blackbox_repository/conversion_scripts/scripts/fcnet_import.py
```diff
@@ -23,7 +23,11 @@
 import pandas as pd
 import numpy as np
 import ast
-import h5py
+
+try:
+    import h5py
+except ImportError:
+    print("Cannot import h5py. Use 'pip install h5py'")
```
geoalgo, line 30: Rather than having custom messages, it would be better to just use print the stack trace.
```
import traceback
try:
  import h5py
except ImportError:
  traceback.print_exc()
```
This way we dont need to write custom messages for everything but most importantly, when error happens users see the stack trace and can see what triggered the issue (maybe it was a file that they did not want to import for instance).

syne_tune/blackbox_repository/blackbox_surrogate.py
```diff
@@ -33,6 +33,13 @@ def transform(self, X):
         return X[self.names]
```
geoalgo, line 33: Same comment here, can you revert those changes? They are unrelated to import.

syne_tune/blackbox_repository/conversion_scripts/scripts/fcnet_import.py
mseeger: The problem is that these messages are also printed when h5py is not used, which is totally fine. That is the price for having the import on top of the file.

syne_tune/blackbox_repository/blackbox_surrogate.py
mseeger: Hmm, they clean up something and fix a docstring ...

mseeger: You want to change the default 1-NN (which does not work, we know that), you need to change 3 places in the code.

syne_tune/blackbox_repository/conversion_scripts/scripts/fcnet_import.py
geoalgo: This recipe will be only imported if you want to get access to blackboxes so it may a problem then, that being said I am fine leaving a custom comment there if you prefer this option.

syne_tune/blackbox_repository/blackbox_surrogate.py
geoalgo: To facilitate PRs, it is important to focus on one part whenever the changes are significant (which is the case here). 
For instance, in addition to changing imports this PR also modifies how sagemaker session are initialized and how surrogate default are instantiated. This makes it longer to get merged as those are quite independent changes. 

I would prefer to have the discussion on how we could change the default for surrogates on another PR to merge this one faster.

mseeger: You refer to `on_tuner_save` in  `TrialBackend`?

This is just refactoring some code, marked as "hack", in `Tuner`. You yourself criticize my code for doing that, so you should not criticize me for fixing it on the fly. We should avoid this "if isinstance" stuff in general code, and it helped me avoiding a sagemaker depedence.

It takes more time to have these discussions that just to merge things in. Feel free to clean up "my" code as well, I have no problem with that.

Because if I send you a PR with just a clean-up, you are going to complain this does not map to an issue.

It is general good practice to clean up code on the fly, no matter who originally wrote it.

geoalgo: Yes, I am referring to this one. Now, I understood that this change is included because of the missing SM import but I did not guess earlier. I asked about this point to the previous PR yesterday and on this PR too 2 hour ago, it would have been helpful for me to know earlier that this change was done because of the SM import as it was not obvious to me.

What I am saying is that having PRs focused on one point makes them easier and faster to merge. This is a better experience for everyone. If there is a bug or a docstring that needs to be fixed it can certainly be included as long as the PR does not become too complex. But if they are points that are included that are not related and that can be argued, it slows the discussion and makes the review more complicated. For instance, the default_surrogate is not fixing anything critical nor a bug fix.

Regarding criticizing your code, this is not my intent. I try to give explanations and references for my comments (such as PEP to justify the practice of putting comments on top). If one comment is not justified or referenced enough, I will appreciate your feedback on it.

geoalgo (approved):

syne_tune/blackbox_repository/blackbox_surrogate.py
mseeger: OK, sorry as well. Next time, I point out the reason for changes.

mseeger: BTW: I also found the reason why we have the surrogate problems. Pretty subtle

Issue #319, opened 2022-08-25
BlackboxSurrogate returns inconsistent values for LCBench blackbox


mseeger:
This is a follow-up issue for #304.

mseeger:
Debug code here: https://github.com/awslabs/syne-tune/blob/debug_martin/syne_tune/blackbox_repository/simulated_tabular_backend.py#L156
And here: https://github.com/awslabs/syne-tune/blob/debug_martin/syne_tune/blackbox_repository/utils.py#L54

mseeger:
If I then run python benchmarking/nursery/benchmark_automl/benchmark_main.py --num_seeds 1 --method ASHA --benchmark lcbench-airlines

I get output like this:

[...]

PR #333, opened 2022-09-01
Test blackboxes and fix LCBench surrogate issue

mseeger:
*Issue #, if available:* 319

*Description of changes:*
This is fixing issue #319 . The solution is to use KNN with K=1, because that does not interpolate between different points in the table. Notably, all other options (XGBoost, RandomForest) fail with even worse errors (elapsed_time not increasing with resource).

### DIFF
--- a/benchmarking/nursery/benchmark_automl/benchmark_definitions.py
+++ b/benchmarking/nursery/benchmark_automl/benchmark_definitions.py
@@ -26,6 +26,7 @@ class BenchmarkDefinition:
     max_resource_attr: str
     max_num_evaluations: Optional[int] = None
     surrogate: Optional[str] = None
+    surrogate_kwargs: Optional[dict] = None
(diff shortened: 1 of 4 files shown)


wistuba:
I think the problem is that Syne Tune uses 1-NN by default but since we pass `KNeighborsRegressor` as a string we use the default of sklearn which is 5: https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsRegressor.html

geoalgo:
I see indeed, makes sense.

geoalgo (approved):

ltiao:
Hi! Not sure if I entirely got the takeaway from this thread and from #319 -- is it currently the case that using any surrogate other than `KNeighborsRegressor` with `{"n_neighbors": 1}` will cause things to break?

Issue #330, opened 2022-09-01
Add plateau stopper


aaronkl:
Add a stopping criterion that stops the HPO process if it hasn’t improved for N consecutive steps

aaronkl:
addressed in #334

PR #334, opened 2022-09-02
Plateau stopper

aaronkl:
#330 

*Description of changes:* Implements a stopping criterion that terminates the HPO process if the metric hasn't improved for N consecutive steps.

### DIFF
--- a/syne_tune/stopping_criterion.py
+++ b/syne_tune/stopping_criterion.py
@@ -124,17 +124,14 @@ class StoppingCriterion:
 
 class PlateauStopper(object):
     """
-    Early stop the experiment when a metric plateaued across trials.
+    Stops the experiment when a metric plateaued for N consecutive trials
+    for more than the given amount of iterations specified in the patience parameter.
     This code is mostly copied from RayTune.
-
-    Stops the entire experiment when the metric has plateaued
-    for more than the given amount of iterations specified in
-    the patience parameter.
 
     :param metric: The metric to be monitored.
     :param std: The minimal standard deviation after which
              the tuning process has to stop.
-    :param top: The number of best models to consider.
+    :param num_trials: The number of consecutive trials
     :param mode: The mode to select the top results.
              Can either be "min" or "max".
     :param patience: Number of iterations to wait for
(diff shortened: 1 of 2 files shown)


syne_tune/stopping_criterion.py
```diff
@@ -118,3 +120,89 @@ def __call__(self, status: TuningStatus) -> bool:
                     )
                     return True
         return False
+
+
+class PlateauStopper(object):
```
mseeger, line 125: It would be nice to be able to combine this with another stop criterion (stop once any of them return true).

Could be done in another PR.

mseeger (approved):

mseeger (dismissed):

tst/test_plateau_stopper.py
```diff
@@ -0,0 +1,67 @@
...
+# permissions and limitations under the License.
+from syne_tune.optimizer.baselines import RandomSearch
+from syne_tune import Tuner
+from syne_tune.stopping_criterion import PlateauStopper
+from syne_tune.config_space import randint
+from syne_tune.util import script_height_example_path
+from tst.util_test import temporary_local_backend
+
+
+def test_plateau_scheduler():
```
geoalgo, line 21: This is more of a integration/smoke test. It would be nice to have a unit-test in addition that tests more cases with artificial cases. This can be done by just showing artificial results (updating a tuning status) to the plateau and checking that the right behavior (stop or not stop) is obtained for patience/mode/top values.

It is very easy to have an error sneaking in otherwise (for instance with some combination of mode/patience).

tst/test_plateau_stopper.py
```diff
@@ -0,0 +1,127 @@
...
+        },
+        new_results=[
+            (2, {metric: 3.00000}),
+            (3, {metric: 3.000}),
+            (4, {metric: 3.0000}),
+        ],
+    )
+
+    stop_criterion(status)
+    assert stop_criterion(status)
```
geoalgo, line 114: There seems to be a left-over:
```suggestion
    assert stop_criterion(status)
```

geoalgo (approved):

geoalgo (approved):

PR #340, opened 2022-09-09
replace Pytorch MLP by sklearn in BORE

geoalgo:
*Description of changes:* Replace Bore MLP implementation by sklearn instead of pytorch. This removes the pytorch dependency and makes the AutoML benchmark ~12x faster at no cost of accuracy.


aaronkl (approved):
LGTM

PR #345, opened 2022-09-12
PD1 Benchmark

### LINKED ISSUE #344: HyperBO integration
Integrating the HyperBO benchmark introduced in [Pre-trained Gaussian processes for Bayesian optimization](https://arxiv.org/pdf/2109.08215.pdf).

wistuba:
Closes #344 

*Description of changes:* Adding PD1 benchmark.

Implementation details:
- PD1 is not consistently reporting all metrics for all tasks. We keep only the metrics reported across all tasks
- Learning curves may have very different length. We pad them with nans to same length per task.
- Some learning curves do not exist. We keep the hyperparameters and have all-nan objectives.
- BlackboxTabular.hyperparameter_objectives_values replaces nans with last nan value in the same objective. This assumes that the elapsed time is cumulative. This would need to be considered in all future BlackboxTabular with missing values. All-nan hyperparameters are dropped.
- PD1 has epochs/steps that are not steadily increasing. We replaced the original ones with np.arange(1, length+1).


mseeger:
Why not drop configs which do not have any observations?

mseeger:
I am not sure I like this "imputation into the future". We discussed this. Would it not be a better idea to simply stop reporting metrics for a trial? We could also define a specific key in the report dictionary which signals to the scheduler that the trial has stopped and will not publish any further metrics. This could be interpreted as a "failure", but without the Tuner stopping the whole experiment.

syne_tune/blackbox_repository/conversion_scripts/scripts/pd1_import.py
```diff
@@ -0,0 +1,315 @@
...
+        bb_dict = {}
+        for _, task in tasks.iterrows():
+            task_name = "{}_{}{}_batch_size_{}".format(
+                task["dataset"],
+                task["model"],
+                ""
+                if task["hps.activation_fn"] is None
+                else f"_{task['hps.activation_fn']}",
+                task["hps.batch_size"],
+            )
```
geoalgo, line 174: This one is hard to read, can compute it one line above?

```suggestion
            activation_name = "" if task["hps.activation_fn"] is None else f"_{task['hps.activation_fn']}"
            task_name = "{}_{}{}_batch_size_{}".format(
                task["dataset"],
                task["model"],
                activation_name,
                task["hps.batch_size"],
            )
```

syne_tune/blackbox_repository/blackbox_tabular.py
```diff
@@ -150,15 +150,44 @@ def hyperparameter_objectives_values(self):
         :return: X, y of shape (num_evals * num_seeds * num_fidelities, num_hps)
         and (num_evals * num_seeds * num_fidelities, num_objectives)
         """
+        objectives_evaluations = self.objectives_evaluations
+        hyperparameters = self.hyperparameters
+        if np.isnan(np.sum(objectives_evaluations)):
+            # Replace nan with previous value. Assumes that elapsed time is cumulative.
```
mseeger, line 156: OK, so the outcome of this would be report the same error value for all later resources with the same elapsed_time?
This may lead to issues with the scheduler.

syne_tune/blackbox_repository/conversion_scripts/scripts/pd1_import.py
```diff
@@ -0,0 +1,315 @@
...
+            fidelity_space=fidelity_space,
+            objectives_evaluations=objectives_evaluations,
+            fidelity_values=fidelity_values,
+            objectives_names=objectives_names,
+        )
+    return bb_dict
+
+
+if __name__ == "__main__":
+    PD1Recipe().generate()
```
geoalgo, line 315: What checks did you do? Did you try to plot the learning-curves and run a simulation?

syne_tune/blackbox_repository/blackbox_tabular.py
mseeger: I'd favor the following. We pass through nan values to the simulator backend, but make sure that there is only one (or no) switch from not-nan to nan, and then nan until the end.

The simulator backend would react to this by creating a stop event just after the last not-nan report. So this a stop decided by the training script, not the scheduler.

But this would, IMHO, not lead to issues with our schedulers, and it seems the cleanest solution to me.

syne_tune/blackbox_repository/blackbox_tabular.py
```diff
@@ -150,15 +150,44 @@ def hyperparameter_objectives_values(self):
...
+                            ):
+                                objectives_evaluations[config_idx][seed_idx][
+                                    fidelity_idx
+                                ][objective_idx] = objectives_evaluations[config_idx][
+                                    seed_idx
+                                ][
+                                    fidelity_idx - 1
+                                ][
+                                    objective_idx
+                                ]
```
geoalgo, line 176: Rather than calling shape 4 times, this would be more readable I think:

```suggestion
            num_configs, num_seeds, num_fidelities, num_objectives = objective.evaluations.shape
```

syne_tune/blackbox_repository/blackbox_tabular.py
```diff
@@ -150,15 +150,44 @@ def hyperparameter_objectives_values(self):
...
+            # Drop all hyperparameters with all nan objectives.
+            nan_mask = np.isnan(objectives_evaluations).any((1, 2, 3))
+            hyperparameters = hyperparameters[~nan_mask]
+            objectives_evaluations = objectives_evaluations[~nan_mask]
```
geoalgo, line 180: This function used to be ~20 lines now it has ~30 lines more because of a particular case. Could you call a subroutine to avoid this?

```python
    def hyperparameter_objectives_values(self):
        """
        :return: X, y of shape (num_evals * num_seeds * num_fidelities, num_hps)
        and (num_evals * num_seeds * num_fidelities, num_objectives)
        """
        objectives_evaluations = self.objectives_evaluations
        hyperparameters = self.hyperparameters
        if np.isnan(...):
           objectives_evaluations, hyperparameters = self.impute_objectives()
        else:
          Xs = []
          ys = []
          ... 
```

geoalgo (commented):
LGTM, did you test it by running a simulation and plotting the learning curves?

One thing I did not get is your comment: `BlackboxTabular.hyperparameter_objectives_values replaces nans with last nan value in the same objective. This assumes that the elapsed time is cumulative. This would need to be considered in all future BlackboxTabular with missing values. All-nan hyperparameters are dropped.`

In what sense does it assume that the elapsed time is cumulative? Since you are copying the last value over, it seems that the assumption is rather that the value reported is the time per step?

syne_tune/blackbox_repository/blackbox_tabular.py
geoalgo: I think the solution you are suggesting makes sense (passing a sequence of consecutive nans and interpreting as a stop in the backend). 

However, this would require significant changes in other parts which may need some discussion. If the code Martin have is functional, I would be keen to ship it and iterate on this, it would be easy to change the behavior once we are sure our backend/schedulers can properly reacts to this.

mseeger: Makes sense

wistuba:
> Why not drop configs which do not have any observations?

I wanted to keep the original benchmark. For some that might be useful information.

syne_tune/blackbox_repository/conversion_scripts/scripts/pd1_import.py
wistuba: I've plotted some learning curves, manually spot-checked raw and transformed data and added the benchmark to the AutoML snippet and ran ASHA on it for one task only.

geoalgo (approved):

PR #362, opened 2022-09-27
Make saving Tuner object at end of tuning optional (default is to save)

mseeger:
*Issue #, if available:*

*Description of changes:*
Can save substantial amount of disk space for large experiments with many repetitions and cases


geoalgo (commented):
The tuner is also saved regularly (the callback is at l150). Can you also apply the filter there?

```python

            self.tuner_saver = RegularCallback(
                # only saves the tuner when save_tuner option is active
                callback=lambda tuner: tuner.save() if self.save_tuner else None,
                call_seconds_frequency=self.results_update_interval,
            )
```

mseeger:
Great catch!

geoalgo (approved):

Issue #253, opened 2022-05-26
[Feature Request] Parallel Categories Plot


austinmw:
(Apologies for creating multiple recent GitHub issues, this is the last one, I promise!)

I took the DataFrame from my experiment results and used Plotly's `plotly.express.parallel_categories` plot to visualize hyperparameter interactions, dropping any features that only have one unique value. This is [an interactive plot](https://plotly.com/python/parallel-categories-diagram/), and you can wrap it in a function that refreshes periodically when new data is available:

![image](https://user-images.githubusercontent.com/12224358/170576482-13525aba-b32e-4507-9f32-e10be1fabfe0.png)

This has been super useful for myself, so I thought that it may be useful to others as well if it were added as a plotting capability to the library? Although I'd understand if it's not desirable to add another dependency. Just thought I'd share!

geoalgo:
> (Apologies for creating multiple recent GitHub issues, this is the last one, I promise!)

Not at all! Your feedback and suggestions are highly appreciated :-)

It would certainly be helpful to add more visualization in general (for now we just have the best plot over time that is built-in). 

As you mention, adding too many dependencies is something we generally want to be watchful about. That being said, adding extra-dependencies is fine, as long as they are optional. 

We have this [folder](https://github.com/awslabs/syne-tune/tree/main/benchmarking) to store benchmarking functionalities, perhaps this tool could be added here?

austinmw:
Thanks, that would be great! Here's the quick and dirty implementation I used:

```python
import os
import pandas as pd
from pathlib import Path
import shutil
import json
import time
import boto3
import pickle
import logging
from typing import Optional
import matplotlib.pyplot as plt
import plotly.express as px
from IPython.display import display, clear_output
from botocore.exceptions import ClientError
import sagemaker
from sagemaker.debugger import TensorBoardOutputConfig, CollectionConfig, DebuggerHookConfig, ProfilerConfig
from syne_tune.backend.sagemaker_backend.sagemaker_utils import get_execution_role
from syne_tune.backend import LocalBackend, SageMakerBackend
from syne_tune.search_space import loguniform, uniform, finrange, randint, choice
from syne_tune.optimizer.schedulers.hyperband import HyperbandScheduler
from syne_tune.stopping_criterion import StoppingCriterion
from syne_tune.tuner import Tuner
from syne_tune.remote.remote_launcher import RemoteLauncher
from syne_tune.experiments import experiment_path, download_single_experiment, ExperimentResult
[...]

PR #367, opened 2022-09-29
Add IPython util function to display results as parallel categories plot

austinmw:
*Issue #, if available:*
https://github.com/awslabs/syne-tune/issues/253

*Description of changes:*
Added `plot_tuner_results` function with multiple options, but primarily displays the following interactive plot:

newplot

### DIFF
--- a/benchmarking/nursery/visualize_plotly/interactive_plots.py
+++ b/benchmarking/nursery/visualize_plotly/interactive_plots.py
@@ -1,0 +1,160 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+import json
+import logging
+import os
+import shutil
+import time
+from pathlib import Path
+from typing import Optional
+
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import plotly.express as px
+from IPython import get_ipython
+from IPython.display import clear_output, display
+from syne_tune.constants import SYNE_TUNE_DEFAULT_FOLDER
+from syne_tune.experiments import load_experiment
+
+
+def plot_tuner_results(tuner_name: str,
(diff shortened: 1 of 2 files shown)


benchmarking/nursery/visualize_plotly/interactive_plots.py
```diff
@@ -0,0 +1,160 @@
...
+
+    while True:
+        ready_to_display = False
+
+        # Clear cached data
+        clear_output(wait=True)
+        syne_local_path = os.path.join(str(Path.home()), SYNE_TUNE_DEFAULT_FOLDER)
+        if os.path.exists(syne_local_path):
+            logging.info('clearing local syne tune cache...')
+            shutil.rmtree(syne_local_path)
```
geoalgo, line 80: This will wipe out all experiments stored locally by default, could you remove it or use another temporary directory?

geoalgo (commented):
Thanks, I have one comment regarding the deletion of local files, otherwise it looks good to me.

Another thing, could you run `./dev-setup.sh` locally before updating the PR? (it will format and add license, see here for more information https://github.com/awslabs/syne-tune/blob/main/CONTRIBUTING.md#contributing-via-pull-requests)

benchmarking/nursery/visualize_plotly/interactive_plots.py
austinmw: How's this update? I changed it to just clear the single job so that it can continue to overwrite and retrieve live updates

austinmw: Ran `dev-setup.sh` committing

benchmarking/nursery/visualize_plotly/interactive_plots.py
```diff
@@ -28,27 +27,30 @@
...
-                      max_cols: int = 100,
-                      display_dataframe: bool = True,
-                      display_metadata: bool = False,
-                      dimensions_max_cardinality: int = 50,
-                      save_figure: bool = False,
-    ) -> Optional[pd.DataFrame]:
+def plot_tuner_results(
+    tuner_name: str,
+    experiment_name: str,
+    overwrite=True,
```
geoalgo, line 33: Could you add the type and also makes this one switched off by default?
```suggestion
    overwrite: bool = False,
```
(to make it less likely that a user runs plot_tuner_results("./") which would wipe out all experiments)

geoalgo (commented):
LGTM, thanks for adding the overwrite option, LGTM.
I just have a small remark left.

benchmarking/nursery/visualize_plotly/interactive_plots.py
austinmw: Oh yep sure, done!

geoalgo (approved):

PR #373, opened 2022-10-04
Making YAHPO benchmarks available in benchmarking/common/benchmark_de…

mseeger:
…finitions/. Refactoring to accommodate more benchmarks. Also some bug fixes, and a comparison test (test still WIP)

*Issue #, if available:* 372 (does not fix it so far)

*Description of changes:*
This is the first part of making all YAHPO benchmarks available and running experiments on them. It includes:
- Refactoring of benchmarking/commons, especially benchmarking/commons/benchmark_definitions, in order to accommodate the YAHPO benchmarks
- benchmarking/nursery/benchmark_dehb is also running YAHPO lcbench
- Cleaned up requirements.txt for the setups in nursery
- Replace Dict -> dict in blackbox_repository code
- syne_tune/blackbox_repository/conversion_scripts/scripts/yahpo.md provides overview of YAHPO scenarios
- syne_tune/blackbox_repository/conversion_scripts/scripts/yahpo_import.py:
   Some important changes
- tst/blackbox_repository/test_comparison.py: Comparison test covering fcnet and lcbench. This is still WIP, in the sense
   that differences are too big (esp for fcnet). In any case, this test will be skipped in the CI, because it needs the blackbox files 


syne_tune/blackbox_repository/conversion_scripts/scripts/yahpo_import.py
```diff
@@ -126,7 +166,7 @@ def set_instance(self, instance):
             instance_names = self.benchmark.config.instance_names
         else:
             instance_names = "instance-names"
-        self.configuration_space[instance_names] = cs.choice([instance])
+        self.configuration_space[instance_names] = instance
```
mseeger, line 169: We want constant values instead of choice of size 1. Also for other such HPs.

mseeger:
One comment about fcnet. We can of course take "yahpo-fcnet" out, because we are quite happy with our non-surrogate benchmark.
The reason I looked at it here, because I wanted to do this comparison.

benchmarking/commons/benchmark_definitions/yahpo.py
```diff
@@ -0,0 +1,477 @@
...
+
+
+yahpo_fcnet_benchmark_definitions = {
+    "yahpo-fcnet-" + name: yahpo_fcnet_benchmark(name) for name in yahpo_fcnet_instances
+}
+
+
+def yahpo_iaml_rpart_benchmark(dataset_name, metric, mode="max"):
+    return BenchmarkDefinition(
+        max_wallclock_time=1800,  # TODO
```
geoalgo, line 138: Optional comment for this PR: I would recommend to set a bound on `max_num_evaluations`
https://github.com/awslabs/syne-tune/blob/main/syne_tune/stopping_criterion.py#L35

For instance, one can set `max_num_evaluations` to `100 x max_t` which ensures that 100 full configurations are seen for fifo schedulers and a similar number of evaluation is seen for hyperband schedulers.

benchmarking/commons/benchmark_main.py
```diff
@@ -36,6 +37,17 @@
 from syne_tune.tuner import Tuner
 
 
+BenchmarkDefinitions = Union[
```
geoalgo, line 40: Could you add a comment on what the string are?

Also having a union can make things more complicated (in particular if it depends on several cases). Is there a particular reason why we would need it? For instance, is it because you want to be able to run all "yahpo-rbv2" benchmarks?

syne_tune/blackbox_repository/conversion_scripts/scripts/yahpo_import.py
```diff
@@ -64,6 +64,24 @@ def download(target_path: Path, version: str):
...
+def _check_whether_fcnet(benchmark: BenchmarkSet) -> bool:
+    return benchmark.config.config_id == "fcnet"
+
+
+def _check_whether_iaml(benchmark: BenchmarkSet) -> bool:
+    return benchmark.config.config_id.startswith("iaml_")
+
+
+def _check_whether_rbv2(benchmark: BenchmarkSet) -> bool:
+    return benchmark.config.config_id.startswith("rbv2_")
```
geoalgo, line 79: Those seem unused, if so could you remove it?

syne_tune/blackbox_repository/conversion_scripts/scripts/yahpo_import.py
```diff
@@ -64,6 +64,24 @@ def download(target_path: Path, version: str):
...
+
+
+# TODO: For fcnet, we also need to determine the "elapsed_time" metric from
+# "runtime", the latter is the time for all 100 epochs. Here, we assume that
+# "elapsed_time" for r epochs is `r / 100` times "runtime".
```
geoalgo, line 84: Regarding FCNET, we could also not expose it given that YAHPO fits a surrogate on a grid of evaluations which is unlikely to provide good results (and was not mentioned in the paper). 
Given this, I would have a preference to leave it out as it is one less thing to maintain but I leave the decision up to you.

syne_tune/blackbox_repository/conversion_scripts/scripts/yahpo_import.py
```diff
@@ -136,7 +176,7 @@ def instances(self) -> np.array:
     @property
     def fidelity_values(self) -> np.array:
         fids = next(iter(self.fidelity_space.values()))
-        return np.arange(fids.lower, fids.upper)
+        return np.arange(fids.lower, fids.upper + 1)
```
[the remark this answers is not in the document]
geoalgo: Good catch, thanks.

tst/blackbox_repository/test_comparison.py
```diff
@@ -0,0 +1,191 @@
...
+        benchmark,
+    )
+    for name, benchmark in fcnet_benchmark_definitions.items()
+]
+
+
+benchmark_pairs = benchmark_pairs_fcnet
+
+
+# This test is still WIP
```
geoalgo, line 53: Can we take this one out given that it is WIP and it would take a long time to run in the pipeline? (We would still have access to it with git)

benchmarking/commons/benchmark_definitions/yahpo.py
mseeger: In the next PR, I am providing better numbers for these defaults. But I'll also add max_num_evaluations

mseeger: Note: yahpo.py will change quite a bit in the next PR for iaml and rbv2, because the way it is done here is cumbersome to work with.

benchmarking/commons/benchmark_main.py
mseeger: Yes, this allows you to parallelize over benchmarks. The outer dict layer is dealt with in remote launcher, the inner dict layer in the sequential loop over benchmarks in benchmark_main.

This is applied in the next PR. YAHPO rbv2 has to many instances that there is no other way.

syne_tune/blackbox_repository/conversion_scripts/scripts/yahpo_import.py
mseeger: Yes, they are only needed in the next PR, so I'll remove them here.

syne_tune/blackbox_repository/conversion_scripts/scripts/yahpo_import.py
mseeger: This will simplify the code

tst/blackbox_repository/test_comparison.py
mseeger: Yes

mseeger: Actually, the test is not run in the pipeline. I'd like to keep it in for lcbench, because once they refit their surrogate, I'd like to do the comparison again.
Eventually, we should run such tests somehwere.

geoalgo: If the test is not run in the pipeline and need further work then I would be keen on not having it. We do not loose the code by including it in this PR but we have to maintain some code that is in WIP state if we include it.

mseeger: OK

geoalgo (approved):

PR #376, opened 2022-10-12
Support for launching experiments with real benchmarks (both local an…

mseeger:
…d SM back-end)

*Issue #, if available:*

*Description of changes:*
Templates for launcher scripts for local and SageMaker back-end, shares usability with those for simulator back-end

- Move lstm_wikitext2 from nursery
- Wrappers for real benchmarks in benchmarking/commons/benchmark_definitions
- Separate hpo_main and launch_remote code for simulated, local, sagemaker
- Management of requirements.txt files


benchmarking/commons/hpo_main.py
```diff
@@ -0,0 +1,95 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
aaronkl, line 1: There seems to be a lot of redundancy between the different hpo_*.py files. Would it make sense to merge them into a single file?

benchmarking/cli/benchmark_factory.py
```diff
@@ -35,7 +35,7 @@
     resnet_cifar10_benchmark,
     resnet_cifar10_default_params,
 )
-from benchmarking.nursery.lstm_wikitext2.definition_lstm_wikitext2 import (
+from benchmarking.definitions.definition_lstm_wikitext2 import (
```
aaronkl, line 38: Do we really want this in the benchmarks? Are we planning to use this in the future, or rather switch to a transformer based benchmark?

benchmarking/commons/benchmark_definitions/common.py
```diff
@@ -11,11 +11,12 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 from dataclasses import dataclass
+from pathlib import Path
 from typing import Optional, List
 
 
 @dataclass
-class BenchmarkDefinition:
+class SurrogateBenchmarkDefinition:
```
aaronkl, line 19: Maybe we can add some documentation here, explaining the difference between real and surrogate benchmark?

benchmarking/commons/hpo_main.py
mseeger: I already have common functions in hpo_main.py. For the remainder, so hpo_main_*.py, they are really all a bit different in important ways. Any suggestion what could still be unified?

But they should rather be templates for people who want to modify them. You should select one of them, based on the back-end you are using. While we always say that this need 1 line of change only, this is not quite the full story, as you see from the differences in these 3 files.

benchmarking/cli/benchmark_factory.py
mseeger: Do you think this benchmark is not good?

I'd like to have it, at least until there is something else more or less equivalent in terms of running time and difficulty.

benchmarking/commons/benchmark_definitions/common.py
mseeger: OK

aaronkl (approved):

PR #394, opened 2022-10-20
Make BORE evaluate pre-selected configurations first 

aaronkl:
*Description of changes:*

Small fix such that BORE actually evaluate all configs in `points_to_evaluate` before starting the search.

### DIFF
--- a/syne_tune/optimizer/baselines.py
+++ b/syne_tune/optimizer/baselines.py
@@ -290,7 +290,7 @@ class BORE(FIFOScheduler):
         super(BORE, self).__init__(
             config_space=config_space,
             metric=metric,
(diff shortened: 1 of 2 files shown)


mseeger (commented):
It would be great to have a test, something like create scheduler with points_to_evaluate  being something, ask for that many configs, and check they are the same?

There is tst/test_pointstoevaluate.py, but it is doing something else.

Could also put that in our backlog, but I think we should have it.

jgolebiowski (approved):
LGTM!

Agreed with Matthias that a test would be a great addition. Depending how fast we want to move with this change itmight need to be added to this PR

aaronkl:
I agree. But this is independent of BORE so I would do this in another PR

mseeger (approved):

PR #396, opened 2022-10-20
Further refactorings of benchmarking/commons to accomodate YAHPO scen…

mseeger:
…arios with many instances

*Issue #, if available:*

*Description of changes:*

- Organizes all remaining YAHPO benchmarks, by filtering out trivial ones and marking interesting ones (selected)
- Small refactoring of command line arguments in hpo_main_*.py
- Example in benchmarking/nursery/benchmark_yahpo to run RS, BO, ASHA on all yahpo_iaml, yahpo_rbv2
- Changes to yahpo_import.py to deal with yahpo_iaml, yahpo_rbv2 (e.g., transform of fidelities)


benchmarking/commons/benchmark_definitions/common.py
```diff
@@ -46,6 +46,7 @@ class SurrogateBenchmarkDefinition:
     surrogate_kwargs: Optional[dict] = None
     max_resource_attr: Optional[str] = None
     datasets: Optional[List[str]] = None
+    fidelities: Optional[List[int]] = None
```
jgolebiowski, line 49: NIT: Description of this parameter is missing from the docstring.

benchmarking/commons/benchmark_definitions/yahpo.py
```diff
@@ -199,21 +226,133 @@ def yahpo_iaml_xgboost_benchmark(dataset_name, metric, mode="max"):
...
+
+def yahpo_iaml_xgboost_benchmark_definitions(restrict_fidelities=False) -> dict:
+    return yahpo_iaml_benchmark_definitions("xgboost", restrict_fidelities)
+
+
+def yahpo_iaml_xgboost_selected_benchmark_definitions(
+    restrict_fidelities=False,
+) -> dict:
+    return yahpo_iaml_selected_benchmark_definitions("xgboost", restrict_fidelities)
+
```
jgolebiowski, line 324: Is it necessary to spell out all of those combinations? I have not seen them being used anywhere. Maybe it would more consise to `modify yahpo_iaml_benchmark_definitions` to the form of
```
def yahpo_iaml_benchmark_definitions(method, restrict_fidelities=False) -> dict:
    assert method in yahpo_iaml_benchmark, f"method must be one of {yahpo_iaml_benchmark.keys()}"
```

and do the same with `yahpo_iaml_selected_benchmark_definitions`. This way the definitions will be easier to extend.

benchmarking/commons/benchmark_definitions/yahpo.py
```diff
@@ -20,15 +20,40 @@
...
-# - Check that elapsed_time_attr is always cumulative
-# - Check that metric is validation, what we want
-# - Run 2 baseline methods with all these benchmarks, in order to figure out
-#   the most interesting ones
+# -----
+# nb301
+# -----
+
+
+def yahpo_nb301_benchmark(dataset_name):
```
jgolebiowski, line 33: NIT (naming): This goes to all the benchmark definition functions in the file. I would suggest renaming them as `yahpo_nb301_benchmark` -> `get_yahpo_nb301_benchmark` to better represent the function of each method.

benchmarking/commons/benchmark_definitions/yahpo.py
```diff
@@ -435,32 +473,1524 @@ def yahpo_rbv2_svm_benchmark(dataset_name, metric, mode="max"):
...
+# - auc:
+#   After sampling 1000 configs at random, best auc < 0.999
+#   Statistics of instances filtered out:
+#   rpart:   -
+#   glmnet:  min = 0.9990544319152832, avg = 0.999367892742157
+#   ranger:  min = 0.9990753531455994, avg = 0.9997223019599915
+#   xgboost: min = 0.9995461702346802, avg = 0.999969482421875
+#   svm:     min = 0.999025821685791, avg = 0.9996441006660461
+#   aknn:    min = 0.9991077780723572, avg = 0.9997339844703674
+yahpo_rbv2_instances = {
```
jgolebiowski, line 537: The data stored from here to L1912 should probably be stored in resource files (csv, json , etc.) rather than in function defintions.

benchmarking/commons/benchmark_definitions/yahpo.py
```diff
@@ -435,32 +473,1524 @@ def yahpo_rbv2_svm_benchmark(dataset_name, metric, mode="max"):
...
+def yahpo_rbv2_svm_selected_benchmark_definitions(restrict_fidelities=False) -> dict:
+    return yahpo_rbv2_selected_benchmark_definitions("svm", restrict_fidelities)
+
+
+def yahpo_rbv2_aknn_benchmark_definitions(restrict_fidelities=False) -> dict:
+    return yahpo_rbv2_benchmark_definitions("aknn", restrict_fidelities)
+
+
+def yahpo_rbv2_aknn_selected_benchmark_definitions(restrict_fidelities=False) -> dict:
+    return yahpo_rbv2_selected_benchmark_definitions("aknn", restrict_fidelities)
```
jgolebiowski, line 1996: This is similar to https://github.com/awslabs/syne-tune/pull/396/files#r1003251413

benchmarking/commons/benchmark_definitions/yahpo.py
mseeger: Makes sense

benchmarking/commons/benchmark_definitions/yahpo.py
mseeger: We typically avoid these no-info terms like "get". In the end, this is just meta-data for a benchmark.

benchmarking/commons/benchmark_definitions/yahpo.py
```diff
@@ -199,21 +226,133 @@ def yahpo_iaml_xgboost_benchmark(dataset_name, metric, mode="max"):
...
+    },
+    "ranger": {
+        "f1": ["41146", "1489", "1067"],
+        "auc": ["41146", "1489"],
+    },
+    "xgboost": {
+        "f1": ["40981", "1489"],
+        "auc": ["40981", "1489"],
+    },
+}
```
jgolebiowski, line 261: NIT: I think it might be easier to store in a table (in resources) with 
```
benchmark_ID, method, metric, selected
41146       , rpart , f1    , True
41146       , rpart , auc   , True
(...)

```

benchmarking/commons/benchmark_definitions/yahpo.py
mseeger: I don't think this added complexity buys anything here. These variables are not automatically generated.

benchmarking/commons/benchmark_definitions/yahpo.py
mseeger: Again, this is not data automatically generated, but selected more or less by hand.

benchmarking/commons/benchmark_definitions/yahpo.py
jgolebiowski: Makes sense, if thats the general strategy then LGTM.

jgolebiowski (commented):
Overall great PR! We should somehow resolve the issue of large data objects in code but all comments beyond that are minor.

Two general comments:
NIT: This might not be the right scope but the PR is for refactoring so it might be. It seems the hpo_main.py file does not include any entrypoint so the name is a little misleading. Maybe change to hpo_main_utilities, hpo_control_utilities or something alike.

NIT: I have seen that the functions in benchmarks lack type hints. This might be the norm in the benchmarking sub-repository but they are present in the rest of syne-tune code and I would argure they are useul.

benchmarking/nursery/benchmark_yahpo/benchmark_main.py
```diff
@@ -0,0 +1,46 @@
...
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+from benchmarking.commons.benchmark_main import main
```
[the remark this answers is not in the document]
mseeger: Good catch

benchmarking/nursery/benchmark_yahpo/requirements.txt
```diff
@@ -0,0 +1,2 @@
+syne-tune[gpsearchers,blackbox-repository,yahpo]
```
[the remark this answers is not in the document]
mseeger: Good idea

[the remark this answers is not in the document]
mseeger: I don't think this really works (see our trouble of remote launching of SM backend), but I agree that aws is needed.

syne_tune/blackbox_repository/conversion_scripts/scripts/yahpo_import.py
```diff
@@ -79,15 +112,74 @@ def __init__(self, benchmark):
...
 
     def _objective_function(
         self,
-        configuration: Dict,
-        fidelity: Optional[Dict] = None,
+        configuration: dict,
+        fidelity: Optional[dict] = None,
         seed: Optional[int] = None,
-    ) -> Dict:
+    ) -> dict:
```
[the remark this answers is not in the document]
mseeger: You'd have to explain. `dict` is a built-in type, no?

mseeger:
About the data in code: Happy if somebody does this afterwards. My only constraint is not to use pandas, because that is already an extra dependence. `json` is fine.

syne_tune/blackbox_repository/conversion_scripts/scripts/yahpo_import.py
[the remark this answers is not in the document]
jgolebiowski: AFAIK python version before 3.9 do not support parametrized generics (e.g. `dict[str, str]`) meaning a typing module needs to be used (e.g. `typing.Dict[str, str]`) for type hints with parametrized data types. In this case we are not using parameters (`configuration: dict` rather than `configuration: dict[str, str]`) so it makes no difference. However, I think it would be best practice to either use the future annotations import (`from __future__ import annotations`) or revert to using typing so that modifying those annotations with data dypes wouldnt break python 3.7+3.8.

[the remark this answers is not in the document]
mseeger: Sure, but we are using `dict` and not `dict[str, ...]` here.

I am just against writing `Dict` on its own, just like you'd not write `List` or `Tuple` on its own.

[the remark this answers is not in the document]
jgolebiowski: I understand that the convention here is to use `typing` for parametrized generics and build-in types for non-parametrized generics. I havent seen that before but its fine by me.

jgolebiowski:
I think we should address the data-stored-in-code issue in this PR, since it relates to the code introduced in this change. I believe that pushing PRs with unresolved issues and having somebody address them in a followup PR is not great practice. We risk issues never being addressed and generate unnecessary churn.

mseeger:
Haha, you are as tough as David ...

jgolebiowski:
I am glad there is a precedent already :) Someone needs to play the watchdog role unfortunately

jgolebiowski (approved):
Looks great, approved!

mseeger:
Keep giving good feedback, this is how I learned already quite a lot at Amazon (by getting feedback).

Issue #395, opened 2022-10-20
ASHA Promotion Fails


wistuba:
If you run the following example, this error is raise: `AssertionError: Internal error (trial_id 2): on_trial_result called with resource = 1, but largest_update_resource = 3`
`type=stopping` works.

```
from pathlib import Path

from syne_tune.backend import LocalBackend
from syne_tune.optimizer.baselines import ASHA
from syne_tune import Tuner, StoppingCriterion
from syne_tune.config_space import randint

config_space = {
    "steps": 100,
    "width": randint(0, 20),
    "height": randint(-100, 100),
}
entry_point = (
    Path("examples")
    / "training_scripts"
    / "height_example"
    / "train_height.py"
)

scheduler = ASHA(
    config_space,
    metric="mean_loss",
    resource_attr="epoch",
    max_resource_attr="steps",
    type="promotion",
    mode="min",
)

tuner = Tuner(
    trial_backend=LocalBackend(entry_point=str(entry_point)),
    scheduler=scheduler,
    stop_criterion=StoppingCriterion(max_wallclock_time=20),
    n_workers=4,
)

tuner.run()
```

mseeger:
This should become a unit test once this is fixed, preferably with smaller max_wallclock_time.

We already have tst/test_schedulers.py, but apparently the time there is too short.

mseeger:
Fixed by #399

PR #399, opened 2022-10-24
Fix for issue #395

mseeger:
*Issue #, if available:* 395

*Description of changes:*
The problem was that if a training function does not implement checkpointing, it is started from scratch, and its reports before the resource it was resumed, need to be ignored. This was not done properly.


mseeger:
I will add unit tests to run schedulers for a bit longer on the train_height example. This will happen in a subsequent PR, because it is a bit tricky. I need to use the simulator back-end, and service train_height as a blackbox, since otherwise the unit tests would run too long.

wistuba (approved):

Issue #378, opened 2022-10-13
Implement grid search for all datatypes, not just choice


mseeger:
Our `GridSearcher` currently only supports `choice` hyperparameters:

https://github.com/awslabs/syne-tune/blob/main/syne_tune/optimizer/schedulers/searchers/searcher.py#L477

The goal of this issue is to make it work with all types in `config_space.py`.

To this end, I recommend to use
https://github.com/awslabs/syne-tune/blob/main/syne_tune/optimizer/schedulers/searchers/utils/hp_ranges.py#L36

which provides encodings of each datatype to [0, 1]. A grid of size N is then obtained by mapping
[ 0.5/N, 1.5/N, 2.5/N, ... ] through the reverse encoding, which is given by this method:
https://github.com/awslabs/syne-tune/blob/main/syne_tune/optimizer/schedulers/searchers/utils/hp_ranges.py#L156

An object of this class can be created from a config_space dict by:
https://github.com/awslabs/syne-tune/blob/main/syne_tune/optimizer/schedulers/searchers/utils/hp_ranges_factory.py#L26

mseeger:
`HyperparameterRanges` is a very useful helper class to wrap around a config_space dict. It does all the encoding to [0, 1]. Apart from that, `config_to_match_string` maps a config to a string, which can then be used to test for equality. This is not trivial, because we have float values. We need to test for equality in order to avoid sampling at the same config 2x, for example.

Another thing this class is doing is to provide `active_config_space`, which is something needed for transfer tuning.

mseeger:
As discussed, I'd suggest the following for the different types (subclasses of `Domain`):
- reverse encoding for `Float`
- use all values for the finite types (except `Integer`)
- reverse encoding for `Integer`, and check whether the grid values are different

Maybe the user can decide how many anchor points to have for each of the HPs where reverse encoding is used.

PR #406, opened 2022-10-27
Gridsearcher issue 3

mina-ghashami:
*Issue #, if available:* #378

*Description of changes:* 
Added support for continuous hyperparameters to Gridsearch, added three unit tests for it as well


syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -488,6 +501,11 @@ class GridSearcher(BaseSearcher):
         the full specification of the Hyperparameters, and the configurations
         generated is the combinations of these Hyperparameters.
         The specified hyperparameter must be Categorical for now.
+    num_samples: dict
```
aaronkl, line 504: Can you add a comment, what will be used if you leave it to the default, i.e num_samples=None

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -516,13 +534,16 @@ def __init__(
...
         config_space,
         metric,
+        num_samples=None,
         points_to_evaluate=None,
         shuffle_config=True,
         **kwargs,
     ):
         super().__init__(config_space, metric, points_to_evaluate)
-        self._validate_config_space(config_space)
+        GridSearcher.DEFAULT_NSAMPLE = 5
```
aaronkl, line 543: I think it would be cleaner to set `self.num_samples` to some default value if `num_samples` is None instead of introducing another constant

tst/schedulers/test_grid_searcher.py
```diff
@@ -106,3 +113,132 @@ def test_store_and_restore_state_with_initial_config():
...
+        scheduler=myscheduler,
+        sleep_time=0.1,
+        n_workers=num_workers,
+        stop_criterion=stop_criterion,
+    )
+
+    tuner.run()
+
+
+def test_grid_scheduler_categorical():
```
aaronkl, line 170: What's the difference of this unit test to the previous one `test_grid_scheduler()`

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -577,11 +598,53 @@ def get_batch_configs(self, batch_size: int, **kwargs):
...
+                    logger.warning(
+                        "number of samples is required for {}. By default, {} is set as number of samples".format(
+                            hp, GridSearcher.DEFAULT_NSAMPLE
+                        )
+                    )
+            # n_sample for integer hp must be capped at length of the range
+            if isinstance(hp_range, Integer):
+                if hp in self.num_samples:
+                    if self.num_samples[hp] > len(hp_range):
+                        self.num_samples[hp] = GridSearcher.DEFAULT_NSAMPLE
```
mseeger, line 628: min(len(hp_range), GridSearcher.DEFAULT_NSAMPLE)

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -577,11 +598,53 @@ def get_batch_configs(self, batch_size: int, **kwargs):
...
+                        logger.info(
+                            'number of samples for "{}" is larger than its range. By default, {} is set as number of samples'.format(
+                                hp, GridSearcher.DEFAULT_NSAMPLE
+                            )
+                        )
+                else:
+                    self.num_samples[hp] = GridSearcher.DEFAULT_NSAMPLE
```
mseeger, line 635: min

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -577,11 +598,53 @@ def get_batch_configs(self, batch_size: int, **kwargs):
...
+                    logger.info(
```
mseeger, line 636: I'd remove this warning. If the user does not say anything, they don;'t want to see warnings.

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -595,15 +658,43 @@ def _generate_remaining_candidates(self) -> List[dict]:
...
         hp_values = []
-        for key, hp_range in reversed(list(self.config_space.items())):
-            if isinstance(hp_range, Domain):
-                hp_keys.append(key)
+        ## adding categorical, finiteRange, scalar parameters
+        for hp, hp_range in reversed(list(self.config_space.items())):
+            if isinstance(hp_range, Float) or isinstance(hp_range, Integer):
+                continue
+            if isinstance(hp_range, Categorical):
+                hp_keys.append(hp)
```
mseeger, line 673: Move this 2 lines up, removes duplicatons

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -595,15 +658,43 @@ def _generate_remaining_candidates(self) -> List[dict]:
...
+                    hpr.from_ndarray(np.array([point])) for point in _normalized_points
+                ]
+                _hpr_points = list(set(_hpr_points))
+                hp_keys.append(hpr.name)
+                hp_values.append(_hpr_points)
 
         hp_values_combinations = list(product(*hp_values))
+
         if self._shuffle_config:
             shuffle(hp_values_combinations)
```
mseeger, line 700: Sorry, Mina. While this is no your code, I'd love this to get fixed (don't know how this got through our reviews).

Two things:
- Please store hp_values_combinations (list of tuples) and hp_keys as members of the class, and not all_candidates_on_grid (list of dict)
- Create the dict on demand in _next_candidate_on_grid
- Do not use pop(0) in _next_candidate_on_grid. Leave the list of all tuples constant, and use a position index

Using pop is potentially really expensive on a large list. And it is totally unnecessary here. My favourite would be not to create this list of tuples at all, but let us keep this.

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -595,15 +658,43 @@ def _generate_remaining_candidates(self) -> List[dict]:
...
```
mseeger, line 699: Sorry, I noted another miss in the current code.
This shuffle here uses a random generator. We'd like to control these. Could you please inherit from SearcherWithRandomSeed and make sure that this shuffle here uses the random_state? Check RandomSearcher for how this works.

mseeger:
BTW: You can also remove the `get_batch_configs` method, as nobody needs this

mseeger:
Wow, this code is really poor! In `get_state`, it stores `self._remaining_candidates` as part of the state. Instead of just an index into this potentially huge list.

syne_tune/optimizer/schedulers/searchers/searcher.py
mina-ghashami: self.num_samples is a dictionary, setting it to a default value requires going through hps in config_space, checking their type and if they are float or int, add self.num_samples[hp] = 5. I think it is cleaner to have a default constant independent of hps used in config_space. wdyt?

tst/schedulers/test_grid_searcher.py
mina-ghashami: test_grid_scheduler_categorical() tests grid searcher only for categorical variables, and in the absence of num_samples argument. test_grid_scheduler() tests for continuous variables. 
You are right there is alot of overlap, i will keep one of them only.

syne_tune/optimizer/schedulers/searchers/searcher.py
mina-ghashami: good point

syne_tune/optimizer/schedulers/searchers/searcher.py
aaronkl: I see, but in this case I would define the constant outside of the class

tst/schedulers/test_grid_searcher.py
aaronkl: Makes sense. Could you rename test_grid_scheduler() to make this clear? Or merge them together?

syne_tune/optimizer/schedulers/searchers/searcher.py
mina-ghashami: I addressed first and third of bullet points above. But about the second bullet: We need to create the dict once in _generate_all_candidates_on_grid() so that we remove duplicates (if points_to_evaluate is not None, duplicates will exist).

mseeger: That is not so important. Strictly speaking, a scheduler may return a config more than once, it is just we try to avoid that. In this case, it makes things complicated.

syne_tune/optimizer/schedulers/searchers/searcher.py
mina-ghashami: Matthias, Im not super clear on this one. I inherited from SearcherWithRandomSeed and used random_state in the shuffle as following:

import random
random.seed(self.random_state)
shuffle(self.hp_values_combinations)

is this what you meant?

syne_tune/optimizer/schedulers/searchers/searcher.py
mina-ghashami: I changed it as this:
            if isinstance(hp_range, Categorical):
                values = hp_range.categories
            elif isinstance(hp_range, FiniteRange):
                values = hp_range.values
            elif not isinstance(hp_range, Domain):
                values = [hp_range]
            hp_keys.append(hp)
            hp_values.append(values)

but i wonder could we have a case that we need to catch in an else statement?

mseeger: Ah I see, then maybe your old code is OK

syne_tune/optimizer/schedulers/searchers/searcher.py
mseeger: You can just do `self.random_state.shuffle(...)`

syne_tune/optimizer/schedulers/searchers/searcher.py
mseeger: You can just do `self.random_state.shuffle(...)`

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -551,17 +568,139 @@ def __init__(
...
+        for hp, hp_range in reversed(list(self.config_space.items())):
+            if isinstance(hp_range, Float) or isinstance(hp_range, Integer):
+                continue
+            if isinstance(hp_range, Categorical):
+                values = hp_range.categories
+            elif isinstance(hp_range, FiniteRange):
+                values = hp_range.values
+            elif not isinstance(hp_range, Domain):
+                values = [hp_range]
+            hp_keys.append(hp)
```
mseeger, line 675: You were right rejecting my ask, please revert to old code. Some weirdo may call this with a non-supported Domain subtype.

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -551,17 +568,139 @@ def __init__(
...
+                ]
+                _hpr_points = list(set(_hpr_points))
+                hp_keys.append(hpr.name)
+                hp_values.append(_hpr_points)
+
+        self.hp_keys = hp_keys
+        self.hp_values_combinations = list(product(*hp_values))
+
+        if self._shuffle_config:
+            random.seed(self.random_state)
```
mseeger, line 696: No, just do `self.random_state.shuffle(self.hp_values_combinations)`

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -551,17 +568,139 @@ def __init__(
...
+                ]
+                _hpr_points = [
+                    hpr.from_ndarray(np.array([point])) for point in _normalized_points
+                ]
+                _hpr_points = list(set(_hpr_points))
+                hp_keys.append(hpr.name)
+                hp_values.append(_hpr_points)
+
+        self.hp_keys = hp_keys
+        self.hp_values_combinations = list(product(*hp_values))
```
mseeger, line 693: Why keep this in `self.hp_values_combinations` if you do not use it afterwards?

I'd be most happy with this, to be honest:
- Keep (and use) `self.hp_keys` and `self.hp_values_combinations` here
- Forget about `all_candidates_on_grid` and also about `_remove_duplicate_candidates`. As I said, you can
   assume that `self.hp_values_combinations` does not contain duplicates, and returning possibly a config or so 2x is not a problem
- Convert the tuple to dict only in `get_config`

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -551,17 +568,139 @@ def __init__(
...
+            # num_samples is ignored for categorical hps.
+            if isinstance(hp_range, Categorical) or isinstance(hp_range, FiniteRange):
+                if hp in self.num_samples:
+                    logger.info(
+                        'number of samples for categorical variable "{}" is ignored.'.format(
+                            hp
+                        )
+                    )
+
+    def _remove_duplicate_candidates(self, all_candidates) -> List[dict]:
```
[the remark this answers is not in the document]
mseeger: But I'd rather just skip the duplicate checking, see below.

syne_tune/optimizer/schedulers/searchers/searcher.py
mina-ghashami: I think you mean to generate it in `_next_candidate_on_grid()`

mseeger: Assuming that the unrolled list of all grid configs is very large, I'd recommend not storing dict, but tuple, and wrap tuple into grid only in `get_config`. I also recommend not doing the check for duplicates. Will all make the code simopler as well.

mina-ghashami: According to our conversation, we decided to remove de-duplication logic from gridsearch (due to it being expensive) and allow gridSearcher to return duplicate configuration (this happens when points_to_evaluate is not None) As the result one of previous unit tests i.e.`test_store_and_restore_state_with_initial_config()` test in tst/schedulers/test_grid_searcher.py fails.

mina-ghashami: basically, in the current implementation, if init_config (points_to_evaluate) is given, gridSearcher still produces those points and it leads to duplicate configurations.

mseeger: If this test is only testing this feature (no duplicates if points_to_evaluate is given), I'd remove it. If it tests something else, I'd just choose points_to_evaluate to not contain grid points.

mina-ghashami: to double check: we expect initial_points to be in the config_space, correct? As an example, the two following is NOT a proper setup: 
config_space = {
        "char_attr": choice(["a", "b"]),
        "int_attr": choice([1, 2]),
    }
inital_config = [
        {"char_attr": "c", "int_attr": 1},
        {"char_attr": "c", "int_attr": 2},
    ]

mina-ghashami: `test_store_and_restore_state_with_initial_config()` also tests to proper restoring of gridsearcher's state when initial_points is given. So I modified the test to remove the test for de-duplication, and would like to keep it.

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -523,6 +532,13 @@ class GridSearcher(BaseSearcher):
         the full specification of the Hyperparameters, and the configurations
         generated is the combinations of these Hyperparameters.
         The specified hyperparameter must be Categorical for now.
+    num_samples: dict
+        Number of samples per hyperparameter. This is required for hyperparameters
+        of type float, optional for integer hyperparameters, and will be ignored
+        for other types (categorical, scalar). If left unspecified, a default value
+        of GridSearcher.DEFAULT_NSAMPLE will be used for float parameters, and
```
aaronkl, line 539: should be only DEFAULT_NSAMPLE instead of GridSearcher.DEFAULT_NSAMPLE

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -551,17 +567,134 @@ def __init__(
...
+            # num_samples is ignored for categorical hps.
+            if isinstance(hp_range, Categorical) or isinstance(hp_range, FiniteRange):
+                if hp in self.num_samples:
+                    logger.info(
+                        'number of samples for categorical variable "{}" is ignored.'.format(
+                            hp
+                        )
+                    )
+
+    # def _remove_duplicate_candidates(self, all_candidates) -> List[dict]:
```
aaronkl, line 632: Could we remove it completely if we don't want to keep it?

aaronkl (dismissed):
LGTM, just two minor comments

aaronkl (approved):

PR #407, opened 2022-10-27
Fix for imputation of points_to_evaluate for ordinal types

mseeger:
*Issue #, if available:*

*Description of changes:*
Fixed the mid-point rule for the new ordinal types

### DIFF
--- a/syne_tune/optimizer/schedulers/searchers/searcher.py
+++ b/syne_tune/optimizer/schedulers/searchers/searcher.py
@@ -23,6 +23,10 @@ from syne_tune.config_space import (
     Ordinal,
     OrdinalNearestNeighbor,
 )
+from syne_tune.optimizer.schedulers.searchers.utils.common import (
+    Hyperparameter,
+    Configuration,
+)
 from syne_tune.optimizer.schedulers.searchers.bayesopt.utils.debug_log import (
     DebugLogPrinter,
 )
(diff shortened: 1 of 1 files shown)


jgolebiowski:
I think the function has a lof of nested ifs (in a loop no less) that are difficult to parse and work with. I believe we can refactor this function to make it much cleaner while maintaining the same logic. The general rule for the refactor is (1) extract the nested ifs from inside the loop and (2) invert ifs and use early return to reduce nesting. I have taken a stab at how the routine can look like when this is applied and discussed changes below. 


Changes
1. Changed the if for isinstance(hp_range, Domain) to inverse if with continue to reduce nesting
2. Extracted the logic for treatment of default and non-default values to reduce nesting
3. The is-numerical flag was guiding the flow through the function.
   I have opted for early return from inside the new subfunctions to achieve the same purpose.
   I think this is more readable
4. Inversed the ifs for isinstance(hp_range, Ordinal) and isinstance(hp_range, OrdinalNearestNeighbor) with early retun
   to avoid nesting
5. Removed the is_numerical flag since it is no longer needed

```
def _impute_default_config(default_config, config_space):
    new_config = dict()
    for name, hp_range in config_space.items():
        if not isinstance(hp_range, Domain):
            continue
        if name in default_config:
            new_config[name] = _default_config_value(default_config, hp_range, name)
        else:
            new_config[name] = _non_default_config(hp_range, name, new_config)
    return new_config


def _default_config_value(default_config, hp_range, name):
    # Check validity
    # Note: For `FiniteRange`, the value is mapped to
    # the closest one in the range
    val = hp_range.cast(default_config[name])
    if isinstance(hp_range, Categorical):
        assert val in hp_range.categories, (
                f"default_config[{name}] = {val} is not in "
                + f"categories = {hp_range.categories}"
        )
    else:
        assert hp_range.lower <= val <= hp_range.upper, (
                f"default_config[{name}] = {val} is not in "
                + f"[{hp_range.lower}, {hp_range.upper}]"
        )
    return val


def _non_default_config(hp_range, name, new_config):
    if isinstance(hp_range, Categorical):
        if not isinstance(hp_range, Ordinal):
            # For categorical: Pick first entry
            return hp_range.categories[0]

        if not isinstance(hp_range, OrdinalNearestNeighbor):
            num_cats = len(hp_range)
            return hp_range.categories[num_cats // 2]

        lower = hp_range.categories[0]
        upper = hp_range.categories[-1]
    else:
        lower, upper = float(hp_range.lower), float(hp_range.upper)

    if not is_log_space(hp_range):
        midpoint = 0.5 * (upper + lower)
    else:
        midpoint = np.exp(0.5 * (np.log(upper) + np.log(lower)))

    # Casting may involve rounding to nearest value in
    # a finite range
    midpoint = hp_range.cast(midpoint)
    lower = hp_range.value_type(lower)
    upper = hp_range.value_type(upper)
    midpoint = np.clip(midpoint, lower, upper)
    return midpoint
```

jgolebiowski (approved):
LGTM! Feel free to assign a second reviewer if you think I am biased towards my own suggestion :)

mseeger:
Thanks, the code is better now. Feel free to push back in this way.

PR #408, opened 2022-10-31
Avoid Matplotlib Warning

### LINKED ISSUE #251: ExperimentResult plot warning
Hi, with recent versions of matplotlib, `ExperimentResult.plot()` gives me the warning:

> WARNING:matplotlib.legend:No artists with labels found to put in legend.  Note that artists whose label start with an underscore are ignored when legend() is called with no argument.

Valavanca:
*Closes #251*

*Description of changes:*
The warning has been corrected because the metric label is explicitly indicated; Display all metrics in an experiment relative to time.

### DIFF
--- a/syne_tune/experiments.py
+++ b/syne_tune/experiments.py
@@ -49,20 +49,20 @@ class ExperimentResult:
     def plot(self, **plt_kwargs):
         import matplotlib.pyplot as plt
 
-        metric = self.metric_names()[0]
         df = self.results
         if df is not None and len(df) > 0:
             df = df.sort_values(ST_TUNER_TIME)
             x = df.loc[:, ST_TUNER_TIME]
-            y = (
-                df.loc[:, metric].cummax()
-                if self.metric_mode() == "max"
-                else df.loc[:, metric].cummin()
-            )
-            plt.plot(x, y, **plt_kwargs)
-            plt.xlabel("wallclock time")
-            plt.ylabel(metric)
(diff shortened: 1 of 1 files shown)


syne_tune/experiments.py
```diff
@@ -49,19 +49,19 @@ def creation_date(self):
...
-                if self.metric_mode() == "max"
-                else df.loc[:, metric].cummin()
-            )
-            plt.plot(x, y, **plt_kwargs)
-            plt.xlabel("wallclock time")
-            plt.ylabel(metric)
+            for metric in self.metric_names():
+                y_metric = (
+                    df.loc[:, metric].cummax()
+                    if self.metric_mode() == "max"
```
wistuba, line 59: As far as I understand the code, `self.metric_mode()` returns the mode for the first metric. Plots for all metrics with a different mode than the first one, will be plotted incorrectly. I think the user is able to provide as many metrics as she wants but does not necessarily need to provide the corresponding modes. Therefore, I'm not sure we can actually plot all metrics.

wistuba (changes requested):

syne_tune/experiments.py
Valavanca: Сlear. I also found several similar cases in the source code when only a first metric and a singular mode are used. 
For the sake of simplification and consistency, I simply removed the legend call.

wistuba (approved):
Thanks a lot for your contribution.

PR #414, opened 2022-11-03
Isinstance issue

mina-ghashami:
*Issue #, if available:*

*Description of changes:*
This is to check if isinstance will fail the CI test on python3.7 . Im sending this PR so that all CI tests are triggered to run.


geoalgo:
Thanks for adding a minimal code. I see what you mean now more clearly now, doing isinstance with a generic type is indeed not possible in python 3.7.

That being said, are there any example of using `isinstance("x", Hyperparameter)` in the code? I cannot find any, so it is not clear to me if that would be an issue.

mina-ghashami:
> 

When working on gridsearcher, there was a need to use it. See this function: https://github.com/awslabs/syne-tune/blob/main/syne_tune/optimizer/schedulers/searchers/searcher.py#L632
I circumvented the issue by using `elif not isinstance(hp_range, Domain)` in line L651. Ideally i wanted to check if `isinstance(hp_range, Hyperparameter)`.

geoalgo:
Ok thanks I see. The solution you currently have would work although it may be better to explicitly check the types that are allowed (e.g. to assert that you have the str or numeric when entering the branch).

Regarding being able to test `isinstance(x, Hyperparameter)`, I think it is ok not to support it as it is for now a bit of a niche case which we can bypass.

mina-ghashami:
> Ok thanks I see. The solution you currently have would work although it may be better to explicitly check the types that are allowed (e.g. to assert that you have the str or numeric when entering the branch).
> 
> Regarding being able to test `isinstance(x, Hyperparameter)`, I think it is ok not to support it as it is for now a bit of a niche case which we can bypass.
fair enough. I will close the PR then.

PR #418, opened 2022-11-09
Helper Function for Config Space Conversion

wistuba:
Simple helper function to convert a config space to dict and back.

### DIFF
--- a/syne_tune/config_space.py
+++ b/syne_tune/config_space.py
@@ -19,7 +19,7 @@ from copy import copy
 from inspect import signature
 from math import isclose
 import sys
-from typing import Any, Callable, List, Optional, Sequence, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Union
 import argparse
 
 import numpy as np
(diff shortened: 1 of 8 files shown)


syne_tune/config_space.py
```diff
@@ -1281,6 +1281,25 @@ def from_dict(d: dict) -> Domain:
     return domain
 
 
+def config_space_to_dict(
```
geoalgo, line 1284: Technically, we are converting to a dict so the name may a bit ambiguous (I took me some time to realize what this function was doing.

 Do you think `config_space_to_json_dict` (or `to_json_dict`) would be a better name? (and also config_space_from_dict -> `config_space_from_json_dict` or `from_json_dict`)

tst/test_config_space.py
```diff
@@ -132,6 +134,19 @@ def test_serialization():
...
 
+def test_config_space_from_and_to_dict():
+    config_space = {
+        "int": randint(1, 2),
+        "logint": lograndint(3, 4),
+        "float": uniform(5.5, 6.5),
+        "logfloat": loguniform(7.5, 8.5),
+        "categorical": choice(["a", "b", "c"]),
+        "normal": randn(2.0, 1.0),
+        "const_str": "constant",
```
geoalgo, line 145: We would probably want to add an integer and a float to cover all the union types.

geoalgo (commented):
Very nice pull-request, I have just two minor comments.

tst/test_config_space.py
wistuba: added both cases

syne_tune/config_space.py
geoalgo: Makes sense, I agree with that `to_json_dict` would be a better name.

geoalgo (approved):

Issue #404, opened 2022-10-27
YAHPO download and upload issues


wistuba:
When loading the YAHPO benchmark for the first time, it is not downloading it automatically and therefore fails.

Manually calling `generate` will overcome this problem, however, the function does not upload the files to s3 (there is an issue with the folder structure, `yahpo-xgboost` vs `yahpo/xgboost`).

mseeger:
This is odd, why do things work for me? I noted it does not seem to write stuff to S3, but also I am not sure it has to. David is the person to comment on this.

wistuba:
Does it still work after `rm -rf ~/.blackbox-repository/yahpo*`?
The reason to upload it to S3 is to not download it from the public repos all the time. Besides that, for most other benchmarks the preprocessing takes quite a bit of time which we can avoid by saving the preprocessed version.

PR #419, opened 2022-11-10
Fix issue with yahpo uploading/syncing with S3

geoalgo:
*Issue #404 

*Description of changes:*


wistuba (changes requested):
It's not fully addressing the issue. Generation of the benchmark is still not happening.

geoalgo:
> It's not fully addressing the issue. Generation of the benchmark is still not happening.

Could you be more specific? 

It is generating the benchmark in my case and is currently fixing an issue (local files were incorrectly checked for yahpo).

wistuba:
When creating the issue, `generate()` was not triggered. I don't see how these changes will account for that. If you delete files from S3 and locally, does it recreate it correctly?

wistuba:
For `load_blackbox("yahpo-rbv2_xgboost")` I get following exception:

```FileNotFoundError: [Errno 2] No such file or directory: '~/.blackbox-repository/yahpo/rbv2_xgboost/encoding.json'```

**Comment by geoalgo** [comment]
@wistuba it should work now, the logic with paths had to be updated to support sub directories on s3 (which is needed for yahpo).

File: syne_tune/blackbox_repository/repository.py [hunk 1]
```diff
@@ -98,20 +98,12 @@ def load_blackbox(
...
     """
-    if name.startswith("yahpo-"):
-        if yahpo_kwargs is None:
-            yahpo_kwargs = dict()
-        return instantiate_yahpo(name, **yahpo_kwargs)
-
-    tgt_folder = Path(repository_path) / name
+    tgt_folder = blackbox_local_path(name)
 
+    print(f"tgt folder {tgt_folder}")
```
wistuba, line 103: remove

syne_tune/blackbox_repository/repository.py
```diff
@@ -130,17 +122,19 @@ def load_blackbox(
             f"Skipping download of {name} as {tgt_folder} already exists, change skip_if_present to redownload"
         )
     else:
+        logging.info("Local files not found, attempting to copy from S3.")
         tgt_folder.mkdir(exist_ok=True, parents=True)
         try:
-            s3_folder = s3_blackbox_folder(s3_root)
+            s3_folder = blackbox_s3_path(name=name, s3_root=s3_root)
+            print(f"s3_folder {s3_folder}")
```
wistuba, line 129: remove

syne_tune/blackbox_repository/repository.py
geoalgo: Ok, done.

wistuba (approved):

PR #426, opened 2022-11-15
Make train_height example available as Blackbox, so unit tests can us…

mseeger:
…e simulator back-end and run for longer

*Issue #, if available:*

*Description of changes:*
This is wrapping the height example into a Blackbox, so we can run unit tests with the simulator back-end, probing schedulers for a longer time. This PR is reacting to a few bugs Martin found (with promotion based Hyperband), which
would have been caught by the experiment running for longer.

In  particular, `tst/test_schedulers.py` now also runs many of our schedulers for longer with the simulator.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -461,6 +461,8 @@ To visualize the progress of Syne Tune in
 `TensorboardCallback` to the `Tuner` object:
 
 ```
+from syne_tune.callbacks import TensorboardCallback
+
 tuner = Tuner(
     ...
     callbacks=[TensorboardCallback()],
--- a/examples/launch_height_baselines.py
+++ b/examples/launch_height_baselines.py
@@ -11,7 +11,6 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 import logging
-from pathlib import Path
 
 from syne_tune.backend import LocalBackend
 from syne_tune.optimizer.baselines import (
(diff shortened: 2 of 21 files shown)


docs/faq.md
```diff
@@ -461,6 +461,8 @@ To visualize the progress of Syne Tune in
 `TensorboardCallback` to the `Tuner` object:
 
 ```
+from syne_tune.callbacks import TensorboardCallback
```
mseeger, line 464: Fix from previous change

examples/launch_height_baselines.py
```diff
@@ -36,26 +41,17 @@
     max_steps = 100
     n_workers = 4
 
-    config_space = {
```
mseeger, line 39: This is just a simple clean-up, avoiding all this copy and paste, same for all other `launch_XYZ.py` changes

examples/training_scripts/height_example/train_height.py
```diff
@@ -15,20 +15,46 @@
...
+    }
+
+
 if __name__ == "__main__":
     root = logging.getLogger()
     root.setLevel(logging.INFO)
 
     parser = ArgumentParser()
-    parser.add_argument("--steps", type=int)
+    parser.add_argument("--" + MAX_RESOURCE_ATTR, type=int)
```
wistuba, line 57: you could use `f"--{MAX_RESOURCE_ATTR}"`. isn't type `str`?

mseeger: MAX_RESOURCE_ATTR is str, but it is just the name, its value is int.

examples/launch_height_baselines.py
geoalgo: Can we please keep the inline definitions on the example? 

I believe importing constants from another file make them quite less readable for someone checking the library (for instance you cant know that RESOURCE_ATTR is a string without loading another file).

mseeger: OK, reverting back our examples to bad practices. I suppose cleaning up the unit tests is OK?

mseeger: Shall I add a comment:
"If you like to know where the strings "mean_loss" and "epoch" are coming from, just open the file whose name is hard-coded below, search for "report", and you'll find the names hard-coded there?"

And then copy and paste this comment across all `launch_xyz.py` files?

geoalgo: I do not think this comment would help given that we already explain that report keys must be connected to the Tuner in the home page and I would expect this to be reasonably intuitive for users.

mseeger: OK, then let us keep it as it is.

wistuba (approved):

PR #428, opened 2022-11-15
BB hashing fix

aaronkl:
*Description of changes:*

Computes a logical hash for parquet files to avoid differences between machines


geoalgo:
Cool, thanks for the quick fix!

wistuba (changes requested):
doesn't work. I've tried `icml-xgboost` and it keeps regenerating. my hash: `25286c38b3e5a8286a1469613a875cce7d67e527ce48d85d83750ced78fcb0bf`

mseeger:
This is a really difficult task, because Aaron himself cannot test it. I suggest Aaron and Martin work closely together for a bit, until this works out for Martin, which is probably sufficient for the others as well.

aaronkl:
The problem seems to be different a precision of the pandas dataframes

wistuba (approved):
it works for me now.

PR #429, opened 2022-11-16
Update the number of random retires to a larger value

### LINKED ISSUE #430: Bug with Seeded Searchers
Opened on behalf of @timyber:

When we are setting a fixed seed, random sampling always have the same behaviors. and It would be running out of search space if we run a large number of training jobs.
This can be reproduced by testing large budget (e.g. max_training_jobs: 100, batch_size: 1) and setting the seed to a fixed value.

timyber:
*Description of changes:*
When we are setting a fixed seed, random sampling always have the same behaviors. and It would be running out of search space if we run a large number of training jobs. 
This can be reproduced by testing large budget (e.g. max_training_jobs: 100, batch_size: 1) and setting the seed to a fixed value.
So we are increasing the number of random retries (Exclusion List size) to 100k. which is above our config limit(10k).

Closes #430 

### DIFF
--- a/syne_tune/optimizer/schedulers/searchers/gp_fifo_searcher.py
+++ b/syne_tune/optimizer/schedulers/searchers/gp_fifo_searcher.py
@@ -90,7 +90,7 @@ from syne_tune.optimizer.schedulers.utils.simple_profiler import SimpleProfiler
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


mseeger (commented):
Sorry, we need a bit more motivation why this would be a good and required change.

mseeger:
Do not merge this. It is not OK to just make random changes to Syne Tune without talking to us

PR #432, opened 2022-11-17
Timeout

geoalgo:
*Description of changes:* 
* add configuration to timeout slow tests to avoid adding slow tests in the suites without realizing it
* add exception for currently slow tests
* remove grid search tuner tests as they do not test additional edge-cases compared to the other test but they are quite slow (30-60s)
* fasten the estimator test by computing the sagemaker session only once


tst/schedulers/test_grid_searcher.py
```diff
@@ -102,107 +94,6 @@ def test_store_and_restore_state_with_initial_config():
         assert new_config in all_candidates_on_grid
 
 
-def test_grid_scheduler_continuous():
```
mseeger, line 105: In the PR I just merged in, these two tests are run with simulator back-end, so they are quite fast then. We'd not have to remove them.

mseeger: Also, you'd need to check `tst/test_schedulers.py`, where we now also run things with simulator back-end, for a longer wallclock time. Some of these cases may be >1s. Marking them with 5s (for each case, I suppose?) should be fine.

geoalgo: Thanks for the pointer for the grid search, it is indeed very fast now. Regarding the schedulers test, I removed the BO tests for now as they take currently 3 min to run.

mseeger (approved):
OK. BTW, I just caught a nasty bug (during docstring refactoring, I deleted a code line by accident) with test_scheduler.py for hypertune. So I'll see whether we can have one variant of this back in, but shorter.

PR #438, opened 2022-11-25
Optimtable

mlblack:
*Issue #, if available:*

Did not create an issue.

*Description of changes:*

Update of the intro paragraph.
Added section and table of HPO methods.
Added section and table of MO methods.
(Did not add the check-pointing info for now.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,21 +119,24 @@ MOBSTER | Klein, et al. (2020) | model-based | yes | yes | no | dunno
 DEHB | Awad, et al. (2021) | random | no | yes | no | dunno
 HyperTune | Li, et al. (2022) | model-based | yes | yes | no | dunno
 PASHA | Bohdal, et al. (2022)| random | yes | yes | no | dunno
-REA | Real, et al. (2019) | ?? | yes | no | no | dunno
-PBT | Jaderberg, et al. (2017) | random | no | yes | no | dunno
+REA | Real, et al. (2019) | model-based | yes | no | no | dunno
+PBT | Jaderberg, et al. (2017) | model-based | no | yes | no | dunno
 ZeroShotTransfer | Wistuba, et al. (2015) | non-random | yes | no | yes | dunno
 ASHABB | Perrone, et al. (2019)| random | yes | yes | yes | dunno
 AHSACTS | Salinas, et al. (2021)| random | yes | yes | yes | dunno
 RUSH | Zappella, et al. (2021)| random | yes | yes | yes | dunno
 
-The searchers fall into three broad categories, **non-random**, **random** or **model-based**. The random searchers sample candidate hyperparameter configurations uniformly at random, while the model-based searchers sample them non-uniformly at random, according to a model (e.g., Gaussian process, density ration estimator, etc.).
+The searchers fall into three broad categories, **non-random**, **random**, and **model-based**. The random searchers sample candidate hyperparameter configurations uniformly at random, while the model-based searchers sample them non-uniformly at random, according to a model (e.g., Gaussian process, density ration estimator, etc.) or by applying an evolutionary algorithm. Note that Syne Tune also supports ["BoTorch"](https://github.com/pytorch/botorch) searchers.
 
 BLAH on checkpointing.
 
 ## Supported multi-objective optimization methods
 
 Method | Reference | Searcher | Asynchronous? | Multi-fidelity? | Transfer? | Checkpointing?
+:--- | :---: | :---: | :---: | :---: | :---: | :---:
 Constrained Bayesian Optimization | Gardner, et al. (2014) | ?? | yes | no | no | dunno
+
+Non-dominated sorting
 MOASHA | Schmucker, et al. (2021) | random | yes | yes | no | dunno
 
 
@@ -217,7 +220,7 @@ See [CONTRIBUTING](CONTRIBUTING.md#security-issue-notifications) for more inform
 
 If you use Syne Tune in a scientific publication, please cite the following paper:
 
-["Syne Tune: A Library for Large Scale Hyperparameter Tuning and Reproducible Research"](https://openreview.net/forum?id=BVeGJ-THIg9&referrer=%5BAuthor%20Console%5D(%2Fgroup%3Fid%3Dautoml.cc%2FAutoML%2F2022%2FTrack%2FMain%2FAuthors%23your-submissions)) First Conference on Automated Machine Learning 2022
+["Syne Tune: A Library for Large Scale Hyperparameter Tuning and Reproducible Research"](https://openreview.net/forum?id=BVeGJ-THIg9&referrer=%5BAuthor%20Console%5D(%2Fgroup%3Fid%3Dautoml.cc%2FAutoML%2F2022%2FTrack%2FMain%2FAuthors%23your-submissions)) First Conference on Automated Machine Learning, 2022.
 
 
 ```bibtex
--- a/README.md
+++ b/README.md
@@ -125,12 +125,17 @@ ZeroShotTransfer | Wistuba, et al. (2015) | non-random | yes | no | yes | dunno
 ASHABB | Perrone, et al. (2019)| random | yes | yes | yes | dunno
 AHSACTS | Salinas, et al. (2021)| random | yes | yes | yes | dunno
 RUSH | Zappella, et al. (2021)| random | yes | yes | yes | dunno
-# CBO | Gardner, et al. (2014) | ?? | yes | no | no | dunno
-# MOASHA | Schmucker, et al. (2021) | random | yes | yes | no | dunno
 
 The searchers fall into three broad categories, **non-random**, **random** or **model-based**. The random searchers sample candidate hyperparameter configurations uniformly at random, while the model-based searchers sample them non-uniformly at random, according to a model (e.g., Gaussian process, density ration estimator, etc.).
 
 BLAH on checkpointing.
+
+## Supported multi-objective optimization methods
+
+Method | Reference | Searcher | Asynchronous? | Multi-fidelity? | Transfer? | Checkpointing?
+Constrained Bayesian Optimization | Gardner, et al. (2014) | ?? | yes | no | no | dunno
+MOASHA | Schmucker, et al. (2021) | random | yes | yes | no | dunno
+
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,15 +7,14 @@
...
-  to get results in seconds while preserving the real dynamics of asynchronous or synchronous HPO with any
-  number of workers
+This package provides state-of-the-art algorithms for hyperparameter optimization (HPO) with the following key features:
+* Wide coverage (>20) of different HPO methods, including:
+  * Asynchronous versions to maximize utilization and distributed versions (i.e., with multiple workers);
+  * Multi-fidelity methods supporting model-based decisions (BOHB and MOBSTER);
+  * Hyperparameter transfer learning to speed up (repeated) tuning jobs;
+  * Multi-objective optimizers that can tune multiple objectives simultaneously (such as accuracy and latency).
+* HPO can be run in different environments (locally, AWS, simulation) by changing just one line of code.
+* Out-of-the-box tabulated benchmarks that allows you simulate results in seconds while preserving the real dynamics of asynchronous or synchronous HPO with any number of workers.
```
geoalgo, line 17: Thanks for simplifying this one.

README.md
```diff
@@ -99,6 +98,45 @@ tuner.run()
 
 The above example runs ASHA with 4 asynchronous workers on a local machine.
 
+## Supported HPO methods
+
+The following hyperparameter optimization (HPO) methods are available in Syne Tune:
+
+Method | Reference | Searcher | Asynchronous? | Multi-fidelity? | Transfer? 
+:--- | :---: | :---: | :---: | :---: | :---: 
+Grid Search | Someone (sometime) | deterministic | no | no | no
```
geoalgo, line 107: Could you update the reference "Someone (sometime)"? (we might put Bergstra as the method is described compared there as I am not sure what would be the original reference)

Also, I think we should mark this one as asynchronous given that there is no difficulty for the async case and it is supported in syne tune.

README.md
```diff
@@ -99,6 +98,45 @@ tuner.run()
...
+ASHA | Li, et al. (2019) | random | yes | yes | no 
+BOHB | Falkner, et al. (2018) | model-based | yes | yes | no 
+MOBSTER | Klein, et al. (2020) | model-based | yes | yes | no 
+DEHB | Awad, et al. (2021) | random | no | yes | no 
+HyperTune | Li, et al. (2022) | model-based | yes | yes | no 
+PASHA | Bohdal, et al. (2022)| random | yes | yes | no 
+REA | Real, et al. (2019) | model-based | yes | no | no 
+PBT | Jaderberg, et al. (2017) | model-based | no | yes | no 
+ZeroShotTransfer | Wistuba, et al. (2015) | deterministic | yes | no | yes 
+AHSACTS | Salinas, et al. (2021)| random | yes | yes | yes
```
geoalgo, line 124: Typo:
```suggestion
ASHA-CTS | Salinas, et al. (2021)| random | yes | yes | yes 
```

README.md
```diff
@@ -99,6 +98,45 @@ tuner.run()
...
+SyncMOBSTER | Klein, et al. (2020) | model-based | no | yes | no 
+ASHA | Li, et al. (2019) | random | yes | yes | no 
+BOHB | Falkner, et al. (2018) | model-based | yes | yes | no 
+MOBSTER | Klein, et al. (2020) | model-based | yes | yes | no 
+DEHB | Awad, et al. (2021) | random | no | yes | no 
+HyperTune | Li, et al. (2022) | model-based | yes | yes | no 
+PASHA | Bohdal, et al. (2022)| random | yes | yes | no 
+REA | Real, et al. (2019) | model-based | yes | no | no 
+PBT | Jaderberg, et al. (2017) | model-based | no | yes | no 
+ZeroShotTransfer | Wistuba, et al. (2015) | deterministic | yes | no | yes
```
geoalgo, line 123: Perhaps we could use just ZeroShotTransfer given that otherwise all transfer methods could be suffixed with "Transfer".

README.md
```diff
@@ -99,6 +98,45 @@ tuner.run()
...
+PASHA | Bohdal, et al. (2022)| random | yes | yes | no 
+REA | Real, et al. (2019) | model-based | yes | no | no 
+PBT | Jaderberg, et al. (2017) | model-based | no | yes | no 
+ZeroShotTransfer | Wistuba, et al. (2015) | deterministic | yes | no | yes 
+AHSACTS | Salinas, et al. (2021)| random | yes | yes | yes 
+RUSH | Zappella, et al. (2021)| random | yes | yes | yes 
+
+The searchers fall into three broad categories, **deterministic**, **random**, and **model-based**. The random searchers sample candidate hyperparameter configurations uniformly at random, while the model-based searchers sample them non-uniformly at random, according to a model (e.g., Gaussian process, density ration estimator, etc.) and an acquisition function, or by applying an evolutionary algorithm.
+
+Syne Tune also supports [BoTorch](https://github.com/pytorch/botorch) searchers.
```
geoalgo, line 129: Can we add this one to the table rather? (to avoid having methods either in the table or in the text)

We can put Botorch with https://proceedings.neurips.cc/paper/2020/file/f5b1b89d98b7286673128a5fb112cb9a-Paper.pdf as the reference.

README.md
```diff
@@ -99,6 +98,45 @@ tuner.run()
...
+
+## Supported multi-objective optimization methods
+
+Method | Reference | Searcher | Asynchronous? | Multi-fidelity? | Transfer?
+:--- | :---: | :---: | :---: | :---: | :---: 
+Constrained Bayesian Optimization | Gardner, et al. (2014) | model-based | yes | no | no
+MOASHA | Schmucker, et al. (2021) | random | yes | yes | no
```
geoalgo, line 136: Could we put those in the same table given that they are also optimizers/HPO methods? (with a column multi-objective as we do for transfer for instance)

geoalgo (commented):
Looks great thanks for adding this!

README.md
mseeger: Yeah, Bergstra and Bengio seem a default reference, even though they rather bash grid search

README.md
mseeger: Somebody needs to update this, because I am getting warnings now that we use deprecated API.

README.md
mseeger: We also have multi-fidelity BORE, please ask Aaron whether this is functional, but it is in the code.

mseeger: I'd say "GP" instead of "model-based", or "Gaussian process"

mseeger: We also have what I'd right now would call BO-KDE or so, namely this is like BOHB, but not multi-fidelity.
Or we can call it TPE-KDE.

mseeger: "model-based" is too broad, use "GP" and "KDE".

One nice idea would be to use the name you pass to searcher in FIFOScheduler, which would be "bayesopt" for GP, and "kde" for BOHB and TPE-KDE

mseeger: PBT is not model-based, it is random like RS. It just does clever scheduling.
We could call it "EA" (for evol algo), in which case DEHB would also be "EA"

README.md
mseeger: Hmm, maybe a separate table could be a good idea, or at least some kind of structure in the table?
Why not have horizontal lines, and separate like: fifo, multi-fidelity, transfer, multi-objective?

mseeger: And we also have cost-aware BO methods

README.md
mlblack: I deliberately left BoTorch out of the table because it is an ensemble of methods and not just one.

README.md
mlblack: > I'd say "GP" instead of "model-based", or "Gaussian process"

Some model-based methods are not GP-based. I first used the finer denomination, but then we end up with too many IMHO.

mlblack: > Perhaps we could use just ZeroShotTransfer given that otherwise all transfer methods could be suffixed with "Transfer".

I am not sure I follow. Are you suggesting to drop Transfer?

README.md
```diff
@@ -99,6 +98,45 @@ tuner.run()
...
+SyncBOHB | Falkner, et al. (2018) | model-based | no | yes | no 
+SyncMOBSTER | Klein, et al. (2020) | model-based | no | yes | no 
+ASHA | Li, et al. (2019) | random | yes | yes | no 
+BOHB | Falkner, et al. (2018) | model-based | yes | yes | no 
+MOBSTER | Klein, et al. (2020) | model-based | yes | yes | no 
+DEHB | Awad, et al. (2021) | random | no | yes | no 
+HyperTune | Li, et al. (2022) | model-based | yes | yes | no 
+PASHA | Bohdal, et al. (2022)| random | yes | yes | no 
+REA | Real, et al. (2019) | model-based | yes | no | no 
+PBT | Jaderberg, et al. (2017) | model-based | no | yes | no
```
mseeger, line 122: I'd still say that PBT is not model-based but random, in the same sense as DEHB is. Both propose all new configs at random, but then use evolutionary ideas to mutate them, which I'd count towards scheduling.

mseeger (commented):
Just one change about PBT not being model-based, otherwise all OK.

I'd still think we should split "model-based" up into more categories, but OK for now.

README.md
mlblack: > PBT is not model-based, it is random like RS. It just does clever scheduling. We could call it "EA" (for evol algo), in which case DEHB would also be "EA"

I defined model-based as sampling non-uniforrmly at random so in that sense PBT falls in that category I believe. Using EA would be more accurate of course. Happy to change DEHB, PBT and REA to EA.

README.md
mlblack: A separate table is deliberate as multi-objective diserves its own in my opinion.  A bunch of method are not included in the baselines. This would enable us to expand the list here too.

Re the formatting: I agree, but couldn't find a way to nicely format tables in markdown. Happy to take another look, but I suggest later, in order not to block the release. What do you think?

mlblack (commented):
Addressed comments and replied to individual comments.

README.md
mlblack: I changed this in the subsequent commit.

mseeger (approved):
I am OK once DEHB, PBT, REA are switched from "random" to "evolutionary" or "EA" (but need to explain somewhere what EA means).

PR #441, opened 2022-11-30
Preparations for docs generation by sphinx

mseeger:
*Issue #, if available:*

*Description of changes:*
This PR is in preparation of the large PR to enable docs generation by sphinx. It cointains all changes which are not just changing or adding docs. This will make the next PR free of any non-doc changes, and easier to review.

The main change here is to add (back) `__all__` to all `__init__.py` files which import things. This is needed for sphinx to understand these imports (cudos to Martin W for pointing that out!), while they do no harm (we do not use `*` imports in the project).


syne_tune/config_space.py
```diff
@@ -32,11 +32,11 @@ class Domain:
     """Base class to specify a type and valid range to sample parameters from.
 
     This base class is implemented by parameter spaces, like float ranges
-    (`Float`), integer ranges (`Integer`), or categorical variables
```
mseeger, line 35: Mainly doing docs refactoring here, but also moved "legacy" from Ray Tune to the end, so the generated docs are easier to read.

syne_tune/optimizer/baselines.py
```diff
@@ -307,6 +435,11 @@ def __init__(
         sample_size: int = 10,
         **kwargs,
     ):
+        search_options = kwargs.get("search_options")
```
mseeger, line 438: The only non-doc change here. This is a fix

benchmarking/definitions/definition_nasbench201.py
```diff
@@ -1,110 +0,0 @@
...
-
-
-# First is default value
-x_range = ["skip_connect", "none", "nor_conv_1x1", "nor_conv_3x3", "avg_pool_3x3"]
-
-
-_config_space = {k: choice(x_range) for k in CONFIG_KEYS}
-
-
-def nasbench201_default_params(params=None):
```
jgolebiowski, line 39: I think when we remove this some docs/examples might fail, same for `nasbench201_benchmark()`.

For example
`https://github.com/awslabs/syne-tune/blob/prepare_sphinx2/docs/tutorials/pasha/scripts/launch_pasha_nasbench201.py#L9`
`https://github.com/awslabs/syne-tune/blob/prepare_sphinx2/docs/tutorials/pasha/scripts/launch_pasha_nasbench201_study.py#L22`

I didnt find any more but maybe I was not good enough at looking

jgolebiowski (approved):
Those are great changes! I think this CR might break some examples that need to be adjusted before merging but it should be only a small tweak.

syne_tune/optimizer/baselines.py
```diff
@@ -38,6 +38,17 @@
 
 
 class RandomSearch(FIFOScheduler):
+    """Random search.
```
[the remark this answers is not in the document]
mseeger: References are sometimes provided in the underlying classes. But agree, they should be copied here as well.

benchmarking/definitions/definition_nasbench201.py
mseeger: Yes, you are right. The PASHA tutorial is defunct. But it has been defunct before already. Note that it does not feature in our hosted docs.

PR #442, opened 2022-12-01
Sphinx documentation

mseeger:
*Issue #, if available:*

*Description of changes:*
This is the overhaul and extension of our docs, so they generate nicely with sphinx and can be hosted.
All non-doc changes here are already in #441. Once that is merged, I can rebase this here, and then you'll only see doc changes.

To generate the docs:
```
cd docs
make html
```

Careful, this needs extra dependencies I added to `dev`, so make sure to install at least `dev`.

Then, look at `docs/build/html/index.html`.


mseeger:
https://quip-amazon.com/VI1mA2LEKsVA/Feedback-on-Sphinx-Docs-Update

This link is for general feedback, beyond this PR here.

geoalgo (approved):
LGTM, thanks for doing this. I think it will be best to merge soon as the merge can quickly becomes hard if we make doc change in the meantime.

One thing regarding `SKLEARN_ALLOW_DEPRECATED_SKLEARN_PACKAGE_INSTALL` is that we may not need it as we have one sklearn install that is left. It would be nice to remove it if it is not needed after (see my comment on the other PR).

mseeger:
Thanks, the remaining sklearn -> scikitl=-learn fixed it.

geoalgo (approved):

PR #443, opened 2022-12-01
Change deprecated sklearn -> scikit-learn

mseeger:
*Issue #, if available:*

*Description of changes:*
This is breaking our build since today.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -30,6 +30,7 @@ jobs:
       run: python -c "import sys; print(sys.version)"
     - name: Install dependencies
       run: |
+        SKLEARN_ALLOW_DEPRECATED_SKLEARN_PACKAGE_INSTALL=True \
         python -m pip install .[extra] --upgrade pip
     - name: Test with pytest
       run: |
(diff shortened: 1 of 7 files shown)


mseeger:
This is an intermediary fix. #444 will have to sort this out properly

.github/workflows/unit-tests.yml
```diff
@@ -30,6 +30,7 @@ jobs:
       run: python -c "import sys; print(sys.version)"
     - name: Install dependencies
       run: |
+        SKLEARN_ALLOW_DEPRECATED_SKLEARN_PACKAGE_INSTALL=True \
```
geoalgo, line 33: I think you are missing this one `syne_tune/blackbox_repository/requirements.txt`, could you try to see if this can be removed after removing this one?

mseeger: Thanks. I thought I grepped everything

mseeger: That solves it. I am removing the issue. Thanks

mseeger:
Also contained in #442 , so whatever is faster

mseeger:
Already contained in #442

PR #449, opened 2022-12-06
Refactor transformer_wikitext2 benchmark (e.g., checkpointing)

mseeger:
*Issue #, if available:*

*Description of changes:*
Cleans up benchmark code, using lstm_wikitext2 as blueprint. In particular, the checkpointing is simplified, and unused args are removed.


mseeger:
@aaronkl The default for `ffn_ratio` is 1. Is this correct? I thought it was 2 (at least in BERT).

aaronkl:
Defaults are from the original implementationl

aaronkl (approved):
LGTM. Is there a reason why you merged the transformer code all into one file?

mseeger:
I used to run into trouble when working with a venv where torch was not installed. Also, we do not use relative imports elsewhere. In the end, I just tried to make it look like lstm_wikitext2, because we know this works in all circumstances (SageMaker backend, checkpointing, ...).

Issue #422, opened 2022-11-14
Reaching the best trained model checkpoint


ElifCerenGok:
Hi, 
In Ray Tune I can get the checkpoint value of the "best" trial like this:
```
result = tune.run(
        tune.with_parameters(train),
        resources_per_trial={"cpu": 12, "gpu": gpus_per_trial},
        config=config,
        num_samples=num_samples,
        search_alg=hyperopt_search,
        scheduler=scheduler)
```
```
best_trial = result.get_best_trial("accuracy", "max", "last")
best_checkpoint_dir = best_trial.checkpoint.value
```
Then, I can load the best checkpoint to my model like this:
```
model_state, optimizer_state = torch.load(os.path.join(best_checkpoint_dir, "checkpoint"))
my_model = Lenet5()
my_model.load_state_dict(model_state)
```
I couldn't understand how to do the exact same thing with Syne Tune. (I can not reach the best trained model)
Could you please help with this ?
Thank you so much :)

mseeger:
Hello,
Syne Tune is agnostic to how checkpointing is implemented, in the sense that your training script has to implement it (there are helper functions for PyTorch models). For this reason, we also do not have a simple way of accessing individual checkpoints and load corresponding models.

If we standardized the way in which training scripts write/load checkpoints more (as apparently Ray Tune is doing), we could offer this functionality. For example, for a user with PyTorch models, if they use our provided functions to save/load checkpoints (and not some home-made code), we could do it.

However, some training scripts may not include checkpointing (our schedulers still work then), some may checkpoint not after every epoch, and previous checkpoints for a trial are overwritten by newer ones. I am therefore not sure it is great practice to assume that certain checkpoints written during HPO (for the purpose of pause and resume) are available later on.

Also, I'd recommend to retrain your best model on all data, including the validation data, in which case the HPO checkpoint is of no use.

mseeger:
Hello, we have this request on our back-log, to write some docs on how to do what you like to do.

However, please note you will have to implement checkpointing (save to / load from local files) on your own in your training script. If you use PyTorch, we provide some helper functions, please check the docs. If you use Hugging Face, we have an example that includes checkpointing.

ElifCerenGok:
Okay, I understand. Thank yo soo much for your reply :)

geoalgo:
Hi @ElifCerenGok, we added an example let us know if it solves your use-case! :-)

PR #451, opened 2022-12-09
Add example showing how to retrieve a checkpoint from an xgboost model.

geoalgo:
*Issue #, if available:* https://github.com/awslabs/syne-tune/issues/422

*Description of changes:* Provides an example to retrieve the best checkpoint of a model after tuning. Also show log of trials to user when no metrics are emitted.

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -208,18 +208,13 @@ More details are found in
 `this tutorial <tutorials/benchmarking/bm_sagemaker.rst#using-sageMaker-managed-warm-pools>`__.
 
 Retrieving the best checkpoint
-=================================
+==============================
 
-.. literalinclude:: ../../examples/launch_standalone_bayesian_optimization.py
-   :caption: examples/launch_standalone_bayesian_optimization.py
+.. literalinclude:: ../../examples/launch_checkpoint_example.py
+   :caption: examples/launch_checkpoint_example.py
    :lines: 13-
 
-Syne Tune combines a scheduler (HPO algorithm) with a backend to provide a
-complete HPO solution. If you already have a system in place for job
-scheduling and managing the state of the tuning problem, you may want to
-call the scheduler on its own. This example demonstrates how to do this
-for Gaussian process based Bayesian optimization.
-
+Example showing how to retrieve the best checkpoint obtained after tuning.
 
 Launch with SageMaker Backend and Custom Docker Image
 =====================================================
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -207,6 +207,19 @@ are strongly recommended for multi-fidelity HPO with the SageMaker backend.
 More details are found in
 `this tutorial <tutorials/benchmarking/bm_sagemaker.rst#using-sageMaker-managed-warm-pools>`__.
 
+Retrieving the best checkpoint
+=================================
+
+.. literalinclude:: ../../examples/launch_standalone_bayesian_optimization.py
+   :caption: examples/launch_standalone_bayesian_optimization.py
+   :lines: 13-
+
+Syne Tune combines a scheduler (HPO algorithm) with a backend to provide a
+complete HPO solution. If you already have a system in place for job
+scheduling and managing the state of the tuning problem, you may want to
+call the scheduler on its own. This example demonstrates how to do this
+for Gaussian process based Bayesian optimization.
+
 
 Launch with SageMaker Backend and Custom Docker Image
 =====================================================
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -269,6 +269,14 @@ Finally, the scheduler provides additional information about checkpointing in
 this: ``add_checkpointing_to_argparse(parser)`` adds corresponding arguments to
 the parser.
 
+
+How can I retrieve the best checkpoint obtained after tuning?
+===========================================
+
+You can take a look at this example
+`examples/launch_checkpoint_example.py <examples.html#retrieving-the-best-checkpoint>`__
+which shows how to retrieve the best checkpoint obtained after tuning an XGBoost model.
+
 Which schedulers make use of checkpointing?
 ===========================================
 
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -271,7 +271,7 @@ the parser.
 
 
 How can I retrieve the best checkpoint obtained after tuning?
-===========================================
+=============================================================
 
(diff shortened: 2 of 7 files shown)


docs/source/examples.rst
```diff
@@ -207,6 +207,19 @@ are strongly recommended for multi-fidelity HPO with the SageMaker backend.
...
 `this tutorial <tutorials/benchmarking/bm_sagemaker.rst#using-sageMaker-managed-warm-pools>`__.
 
+Retrieving the best checkpoint
+=================================
+
+.. literalinclude:: ../../examples/launch_standalone_bayesian_optimization.py
+   :caption: examples/launch_standalone_bayesian_optimization.py
+   :lines: 13-
+
+Syne Tune combines a scheduler (HPO algorithm) with a backend to provide a
```
mseeger, line 217: Copy and pasted from another paragraph? Also, does not include your example

docs/source/faq.rst
```diff
@@ -269,6 +269,14 @@ Finally, the scheduler provides additional information about checkpointing in
 this: ``add_checkpointing_to_argparse(parser)`` adds corresponding arguments to
 the parser.
 
+
+How can I retrieve the best checkpoint obtained after tuning?
+===========================================
```
mseeger, line 274: Doesn't this === need to be as long as the line above?

examples/launch_checkpoint_example.py
```diff
@@ -0,0 +1,66 @@
+"""
```
mseeger, line 1: Below the copyright comments?

examples/launch_checkpoint_example.py
```diff
@@ -0,0 +1,66 @@
...
+        scheduler=BayesianOptimization(config_space, metric="merror", mode="min"),
+        stop_criterion=StoppingCriterion(max_wallclock_time=10),
+        n_workers=n_workers,
+    )
+
+    tuner.run()
+
+    exp = load_experiment(tuner.name)
+    best_config = exp.best_config()
+    checkpoint = exp.path / str(best_config["trial_id"]) / "checkpoints"
```
mseeger, line 58: Are you sure the config always has this "trial_id" field?

mseeger: I also think it would be better to define some helper function, maybe `checkpoint_path(trial_id)`, which is then both called by `LocalBackend` and by this example here. This allows us to change things later on.

examples/training_scripts/xgboost/xgboost_checkpoint.py
```diff
@@ -0,0 +1,107 @@
...
+    loaded.load_model(path)
+    return loaded
+
+
+def evaluate_accuracy(checkpoint_dir):
+    X, y = load_digits(return_X_y=True)
+
+    clf = load_model(checkpoint_dir=checkpoint_dir)
+    y_pred = clf.predict(X)
+    return (np.equal(y, y_pred) * 1.0).mean()
```
mseeger, line 86: Why this * 1.0 ?

examples/training_scripts/xgboost/xgboost_checkpoint.py
```diff
@@ -0,0 +1,107 @@
...
+
+import numpy as np
+import xgboost
+from sklearn.datasets import load_digits
+
+from syne_tune import Reporter
+from syne_tune.constants import ST_CHECKPOINT_DIR
+
+
+class SyneTuneCallback(xgboost.callback.TrainingCallback):
```
mseeger, line 25: Cool to have an xgboost example

examples/training_scripts/xgboost/xgboost_checkpoint.py
```diff
@@ -0,0 +1,107 @@
...
+    parser = ArgumentParser()
+    parser.add_argument("--max_depth", type=int, required=False, default=1)
+    parser.add_argument("--gamma", type=float, required=False, default=2)
+    parser.add_argument("--reg_lambda", type=float, required=False, default=0.001)
+    parser.add_argument("--n_estimators", type=int, required=False, default=10)
+    parser.add_argument(f"--{ST_CHECKPOINT_DIR}", type=str, default="./")
+
+    args, _ = parser.parse_known_args()
+
+    checkpoint_dir = Path(vars(args)[ST_CHECKPOINT_DIR])
```
mseeger, line 99: getattr(args, ST_CHECKPOINT_DIR) ?

syne_tune/experiments.py
```diff
@@ -46,6 +46,7 @@ class ExperimentResult:
     results: pd.DataFrame
     metadata: dict
     tuner: Tuner
+    path: Path
```
mseeger, line 49: Please also extend the docstring.

BTW: Is this a fix, or something that is needed for this example?

docs/source/examples.rst
geoalgo: That's a type, it is fixed.

docs/source/faq.rst
geoalgo: Did not know, fixed it.

syne_tune/experiments.py
geoalgo: It is needed for this example to get the path of experiment.

examples/training_scripts/xgboost/xgboost_checkpoint.py
geoalgo: Is there an advantage of using getattr?

examples/training_scripts/xgboost/xgboost_checkpoint.py
geoalgo: It allows to compute the average with np.mean.

examples/launch_checkpoint_example.py
geoalgo: > Are you sure the config always has this "trial_id" field?

Yes, this field is always added.

geoalgo: > I also think it would be better to define some helper function, maybe checkpoint_path(trial_id), which is then both called by LocalBackend and by this example here. This allows us to change things later on.

Thanks, I updated with this suggestion.

examples/training_scripts/xgboost/xgboost_checkpoint.py
mseeger: Not so important, but vars(args) probably creates a full dict from, which you just use one entry.

mseeger (approved):

PR #453, opened 2022-12-10
Fix conditional dependencies for backends

mseeger:
*Issue #, if available:*

*Description of changes:*
Previous checks were too wide. `LocalBackend` and `PythonBackend` work with core dependencies only, while `SageMakerBackend` just needs `aws`. None of them need `extra`.


jgolebiowski (approved):
LGTM

PR #459, opened 2022-12-15
Example usage of the ask-tell interface in syne-tune

jgolebiowski:
*Description of changes:*
What: Setup an ask-tell interface like in scikit-optimize ([library](https://github.com/scikit-optimize/scikit-optimize), [issue that asks for it](https://github.com/scikit-optimize/scikit-optimize/issues/68) in scikit-opt) to enable users with very specific requirements to plug into the HPO engines implemented in syne-tune. This work would include providing an example that shows how to set up this mode of operation by calling a trial scheduler suggest and communicating results with on_trial_results.

Why: In order to use syne-tune at the moment, user needs to wrap their function evaluator into a script that can be executed by syne-tune. In some cases, experiments needed for function evaluations can be more complex and require extra orchestration (example vary from setting up jobs on non-aws clusters to runnig physical lab experiments)

### DIFF
--- a/examples/launch_ask_tell_scheduler.py
+++ b/examples/launch_ask_tell_scheduler.py
@@ -132,7 +132,9 @@ def plot_objective():
 
 def tune_with_random_search() -> TrialResult:
     metric, mode, config_space, max_iterations = get_objective()
-    scheduler = AskTellScheduler(base_scheduler=RandomSearch(config_space, metric=metric, mode=mode))
+    scheduler = AskTellScheduler(
+        base_scheduler=RandomSearch(config_space, metric=metric, mode=mode)
+    )
     for iter in range(max_iterations):
         trial_suggestion = scheduler.ask()
         test_result = target_function(**trial_suggestion.config)
@@ -142,7 +144,9 @@ def tune_with_random_search() -> TrialResult:
 
 def save_restart_with_gp() -> TrialResult:
     metric, mode, config_space, max_iterations = get_objective()
-    scheduler = AskTellScheduler(base_scheduler=BayesianOptimization(config_space, metric=metric, mode=mode))
+    scheduler = AskTellScheduler(
+        base_scheduler=BayesianOptimization(config_space, metric=metric, mode=mode)
+    )
     for iter in range(int(max_iterations / 2)):
         trial_suggestion = scheduler.ask()
         test_result = target_function(**trial_suggestion.config)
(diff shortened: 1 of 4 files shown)


examples/launch_ask_tell_scheduler.py
```diff
@@ -0,0 +1,185 @@
...
+
+def save_restart_with_gp() -> TrialResult:
+    metric, mode, config_space, max_iterations = get_objective()
+    scheduler = AskTellScheduler(base_scheduler=BayesianOptimization(config_space, metric=metric, mode=mode))
+    for iter in range(int(max_iterations / 2)):
+        trial_suggestion = scheduler.ask()
+        test_result = target_function(**trial_suggestion.config)
+        scheduler.tell(trial_suggestion, {metric: test_result})
+
+    scheduler.save("scheduler-checkpoint.dill")
```
geoalgo, line 159: If we just serialize with dill, couldnt we skip the save/load method and directly dill the scheduler to simplify the example?

examples/launch_ask_tell_scheduler.py
```diff
@@ -0,0 +1,185 @@
...
+    metric = "mean_loss"
+    mode = "min"
+    max_iterations = 100
+    config_space = {
+        "x": uniform(-1, 1),
+    }
+    return metric, mode, config_space, max_iterations
+
+
+def inspect_objective():
```
geoalgo, line 120: plot_objective might me a better name.

examples/launch_ask_tell_schedululer_hyperband.py
```diff
@@ -0,0 +1,84 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
geoalgo, line 1: There is a funny typo in the name of the file :-)

examples/launch_ask_tell_scheduler.py
```diff
@@ -0,0 +1,185 @@
...
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+import datetime
```
geoalgo, line 13: It could be helpful to add a comment at the top of the file to briefly explains what this example does.

geoalgo (commented):
LGTM, thanks for providing this the code is very clean I just have minor comments.

mseeger:
Please also update `examples.rst`

examples/launch_ask_tell_scheduler.py
jgolebiowski: added

examples/launch_ask_tell_scheduler.py
jgolebiowski: makes sense, chnaged

examples/launch_ask_tell_scheduler.py
jgolebiowski: Good point, I thought making it a function is easier but I buy the shorter=simpler argument so moved it to the function.

geoalgo (approved):

PR #468, opened 2022-12-19
Comment out all tests involving RayTuneScheduler

mseeger:
*Issue #, if available:*

*Description of changes:*
Commenting out tests which touch RayTuneScheduler (which currently is buggy). This is a problem in Ray Tune, not on our end.

### DIFF
--- a/.github/workflows/check-code-examples.yml
+++ b/.github/workflows/check-code-examples.yml
@@ -31,6 +31,6 @@ jobs:
         python examples/launch_height_baselines.py
         python examples/launch_height_python_backend.py
         python examples/launch_height_moasha.py
-        python examples/launch_height_ray.py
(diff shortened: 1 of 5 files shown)


mseeger:
We still need to comment out tests using RayTuneScheduler, because their code still uses np.int, or at least the version of their code we depend upon.

Ray Tune is in the business of wrapping any code they can get their hands on. It is a pretty nasty thing to depend upon.

jgolebiowski (approved):
Looks good, approved. Should we consider commenting out https://github.com/awslabs/syne-tune/blob/main/tst/test_config_space.py#L36 also to avoid any raytune dependencies in tests? If so, I will add it in a followup.

mseeger:
I left in RayTuneScheduler code which does not really call Ray Tune.

PR #469, opened 2022-12-19
Changing np.int into np.int_ to follow the new protocol

jgolebiowski:
Changing np.int into np.int_ to follow the new protocol - https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.int_

### DIFF
--- a/syne_tune/optimizer/schedulers/searchers/bore/bore.py
+++ b/syne_tune/optimizer/schedulers/searchers/bore/bore.py
@@ -258,9 +258,9 @@ class Bore(SearcherWithRandomSeed):
             self.model = CalibratedClassifierCV(
                 self.model, cv=2, method=self.calibration
             )
-            self.model.fit(X, np.array(z, dtype=np.int))
+            self.model.fit(X, np.array(z, dtype=np.int_))
         else:
-            self.model.fit(X, np.array(z, dtype=np.int))
+            self.model.fit(X, np.array(z, dtype=np.int_))
 
         z_hat = self.model.predict(X)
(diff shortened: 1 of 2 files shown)


mseeger:
This does not suffice, we can use #468 . Also need to remove Ray Tune touching tests

syne_tune/optimizer/schedulers/searchers/bore/bore.py
```diff
@@ -258,9 +258,9 @@ def _train_model(self, train_data, train_targets):
             self.model = CalibratedClassifierCV(
                 self.model, cv=2, method=self.calibration
             )
-            self.model.fit(X, np.array(z, dtype=np.int))
+            self.model.fit(X, np.array(z, dtype=np.int_))
```
aaronkl, line 261: is this the recommend way by Numpy now? It looks a bit odd with the underscore

aaronkl (approved):

syne_tune/optimizer/schedulers/searchers/bore/bore.py
mseeger: Please use the other PR I opened. We want to have np.int64 here.

jgolebiowski: yesh approved this. I think int_ is the updated version of np.int - I agree int64 is probably safer but didnt want to change the semantics of the code.

jgolebiowski:
Closing since the change is already in https://github.com/awslabs/syne-tune/pull/468

PR #474, opened 2022-12-21
Update README.md

610v4nn1:
Updating the table of the supported algorithms to make it more precise


README.md
```diff
@@ -125,7 +125,7 @@ BOHB | Falkner, et al. (2018) | model-based | yes | yes | no
 MOBSTER | Klein, et al. (2020) | model-based | yes | yes | no 
 DEHB | Awad, et al. (2021) | evolutionary | no | yes | no 
 HyperTune | Li, et al. (2022) | model-based | yes | yes | no 
-PASHA | Bohdal, et al. (2022)| random | yes | yes | no 
+PASHA | Bohdal, et al. (2022)| random or model-based | yes | yes | no
```
mseeger, line 128: "any" is better

610v4nn1: I reported only the searcher with which the algorithm was tested. While, in principle, it should work reliably also with a deterministic searcher, we never tested this combination.

mseeger (approved):

PR #480, opened 2022-12-22
pytest: Add pytest-xdist

ConnorBaker:
# feat/parallelize-tests

## Why

With `pytest-xdist` we can parallelize the test suite. This results in a quick win for a faster feedback loop.

### Numbers

Below are the results of the test suite on my machine (an i9-13900k):

| command | run # | real | user | sys |
| --- | --- | --- | --- | --- |
| pytest | 1 | 0m35.141s | 1m12.942s | 2m6.306s |
| pytest | 2 | 0m37.991s | 2m18.906s | 2m49.160s |
| pytest | 3 | 0m36.207s | 1m36.687s | 2m39.535s |
| pytest -n 1 --dist loadgroup | 1 | 0m35.896s | 1m25.656s | 2m17.317s |
| pytest -n 1 --dist loadgroup | 2 | 0m38.354s | 2m2.959s | 3m7.145s |
| pytest -n 1 --dist loadgroup | 3 | 0m38.792s | 2m14.633s | 3m8.681s |
| pytest -n 2 --dist loadgroup | 1 | 0m25.270s | 2m24.090s | 3m14.851s |
| pytest -n 2 --dist loadgroup | 2 | 0m28.600s | 3m51.649s | 4m6.193s |
| pytest -n 2 --dist loadgroup | 3 | 0m28.093s | 3m43.806s | 3m49.875s |
| pytest -n 3 --dist loadgroup | 1 | 0m22.370s | 2m59.735s | 3m32.893s |
| pytest -n 3 --dist loadgroup | 2 | 0m19.252s | 1m55.877s | 2m37.809s |
| pytest -n 3 --dist loadgroup | 3 | 0m22.168s | 2m56.645s | 3m25.640s |
| pytest -n 4 --dist loadgroup | 1 | 0m20.715s | 2m50.816s | 3m20.745s |
| pytest -n 4 --dist loadgroup | 2 | 0m20.518s | 2m48.112s | 3m36.456s |
| pytest -n 4 --dist loadgroup | 3 | 0m20.832s | 3m2.586s | 3m28.027s |

The average of each run's real times are:

| command | real |
| --- | --- |
| pytest | 0m36.446s |
| pytest -n 1 --dist loadgroup | 0m37.681s |
| pytest -n 2 --dist loadgroup | 0m27.321s |
| pytest -n 3 --dist loadgroup | 0m21.263s |
| pytest -n 4 --dist loadgroup | 0m20.688s |

Going beyond four processes doesn't seem to yield any further improvements (likely because some portions of the test suite involve parallelized operations).

## How

I added `pytest-xdist` to the `dev` extra requirements. I also updated the `pytest.ini` file with `addopts` to configure the test suite to run with four processes and run tests with the same group on the same worker (to avoid resource contention). The `test_cholesky_factorization` test was refactored into two smaller tests which both increase their timeout with respect to input size. Additionally, tests which involved parallelized operations were added to an `xdist_group` named `parallel` to ensure they are run on the same worker (to avoid resource contention).

While the standard GitHub runners only have two cores (the `unit-tests.yml` workflow has been edited accordingly), local development can benefit from this.

---

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,3 @@
 [pytest]
 timeout = 1
+addopts = -n 4 --dist loadgroup
--- a/tst/schedulers/bayesopt/gpautograd/test_cholesky_factorization.py
+++ b/tst/schedulers/bayesopt/gpautograd/test_cholesky_factorization.py
@@ -17,6 +17,7 @@ from autograd import grad
 
 # from autograd.test_util import check_grads
 import time
+import pytest
 
 from syne_tune.optimizer.schedulers.searchers.bayesopt.gpautograd.custom_op import (
     cholesky_factorization,
@@ -56,26 +57,18 @@ def _testfunc_mahal_from_xb(xb, use_my):
     return _testfunc_mahal(a, b, use_my)
 
 
-def test_cholesky_factorization():
-    # num_rep = 10
-    # min_n = 100
-    # max_n = 2500
-    # Not so useful for time comparison, but runs faster:
-    num_rep = 8
-    min_n = 10
-    max_n = 250
+@pytest.mark.xdist_group("parallel")
+@pytest.mark.parametrize(
+    "xmat_size",
+    [pytest.param(2**i, marks=pytest.mark.timeout(i - 3)) for i in range(4, 9)],
+)
+def test_cholesky_factorization_logdet(xmat_size: int) -> None:
+    num_rep: int = 8
     grad_logdet_my = grad(lambda x: _testfunc_logdet_from_x(x, use_my=True))
     grad_logdet_cmp = grad(lambda x: _testfunc_logdet_from_x(x, use_my=False))
-    grad_mahal_my = grad(lambda xb: _testfunc_mahal_from_xb(xb, use_my=True))
-    grad_mahal_cmp = grad(lambda xb: _testfunc_mahal_from_xb(xb, use_my=False))
     for rep in range(num_rep):
-        n = np.random.randint(min_n, max_n)
-        xmat = np.random.randn(n, n)
-        # check_grads(
-        #    lambda x: testfunc_logdet_from_x(x, use_my=True),
-        #    modes=['rev'], order=1)(xmat)
-        # logdet
-        print("\nn = {}\nlogdet:".format(n))
+        xmat = np.random.randn(xmat_size, xmat_size)
+        print("\nn = {}\nlogdet:".format(xmat_size))
         ts_start = time.time()
         gval_my = grad_logdet_my(xmat)
         time_my = time.time() - ts_start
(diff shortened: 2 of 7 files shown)


mseeger:
Hello Connor, thanks a lot for this contribution, it looks very useful.

I'd like to have @geoalgo to comment on this PR, as he is the expert on our CI setup. He is currently on EoY holiday, but will be back first week of January.

mseeger:
One question: Does this change also speed up the process on GitHub? Or are we already using all available resources there in a good way?

ConnorBaker:
Sadly, because the GitHub runners only have two threads, I don't think there will be much of a speed up (they are also not fast threads). More on their hardware here https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources and here https://learn.microsoft.com/en-us/azure/virtual-machines/dv2-dsv2-series#dsv2-series. (The Linux and Windows runners use Standard_DS2_v2 instances.)

pytest.ini
```diff
@@ -1,2 +1,3 @@
 [pytest]
 timeout = 1
+addopts = -n 4 --dist loadgroup
```
geoalgo, line 3: Is there a benefit from not using auto to avoid being slower in a machine with 2 cores and could underutilize a machine with many core? 
If we switch to auto, we could also probably remove the argument in `.github/workflows/unit-tests.yml`?

tst/schedulers/bayesopt/gpautograd/test_cholesky_factorization.py
```diff
@@ -89,8 +82,20 @@ def test_cholesky_factorization():
             )
         )
         assert max_diff_grad_logdet < 1e-12
-        # mahal
-        print("mahal:")
+
+
+@pytest.mark.xdist_group("parallel")
```
geoalgo, line 87: Is there a particular reason for grouping those tests in the same worker? (rather than just letting be run among any worker)

geoalgo (commented):
Thanks for proposing this, the change looks good to me. I have two small comments.

pytest.ini
ConnorBaker: In the case of your specific test suite, no I don't think we'd underutilize a machine with many cores. There are enough tests that require multiple cores that there wasn't really an observable speedup. See my notes under `Numbers` in the PR.

Additionally, my understanding is that, when given auto, `pytest-xdist` will create as many workers as you have logical cores. In the case that your test suite did parallelize super well, we'd run into an issue with the ordering of the test groups. Tests which use multiple cores would need to be run last to avoid resource starvation while every other core is executing a test. If we didn't handle that, some might fail because they would exceed their time limit.

tst/schedulers/bayesopt/gpautograd/test_cholesky_factorization.py
ConnorBaker: That was done mostly to avoid resource contention. A few of your tests automatically spawn multiple cores and when they ran in parallel, they almost always ran out of the time you had allotted to them. Running them sequentially ensures that the tests which use multiple cores aren't competing among themselves.

ConnorBaker:
Sorry for the delay in getting back to you @geoalgo! Let me know if you have any further questions or comments :)

geoalgo:
Thanks for answering my questions and for proposing this change. It makes lot of sense to have this parallelization.

geoalgo (approved):

mseeger:
Hello @ConnorBaker . Could you rebase and get your branch up-2-date, so that unit tests pass again? Then, we can merge this

mseeger:
Closing this for now. Please reopen if you'd still like to contribute this

Issue #465, opened 2022-12-16
Deprecated package "sklearn" is imported when doing remote launching


mseeger:
Branch: `tmlr_experiments`

Command:
`python benchmarking/nursery/tmlr_experiments/tmlr3/launch_remote.py --experiment_tag tmlr-3 --num_seeds 1 --ignore_blackbox_hash 1 --method RS --verbose 1`

mseeger:
If `basic_cpu_instance_sagemaker_estimator` in `syne_tune/benchmarking/commons/utils.py` returns a `SKLearn` estimator, the training job fails with a log like this:

```
2022-12-16 14:02:22,424 sagemaker-containers INFO     Imported framework sagemaker_sklearn_container.training
2022-12-16 14:02:22,427 sagemaker-training-toolkit INFO     No GPUs detected (normal if no gpus installed)
2022-12-16 14:02:22,434 sagemaker_sklearn_container.training INFO     Invoking user training script.
2022-12-16 14:02:22,875 sagemaker-training-toolkit INFO     Installing dependencies from requirements.txt:
/miniconda3/bin/python -m pip install -r requirements.txt
Collecting syne-tune[aws,blackbox-repository,gpsearchers,yahpo]
  Downloading syne_tune-0.3.2-py3-none-any.whl (537 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 537.7/537.7 kB 13.3 MB/s eta 0:00:00
Requirement already satisfied: tqdm in /miniconda3/lib/python3.8/site-packages (from -r requirements.txt (line 2)) (4.64.1)
Requirement already satisfied: numpy>=1.16.0 in /miniconda3/lib/python3.8/site-packages (from syne-tune[aws,blackbox-repository,gpsearchers,yahpo]->-r requirements.txt (line 1)) (1.19.2)
Collecting dill
  Downloading dill-0.3.6-py3-none-any.whl (110 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 110.5/110.5 kB 32.4 MB/s eta 0:00:00
Requirement already satisfied: pandas in /miniconda3/lib/python3.8/site-packages (from syne-tune[aws,blackbox-repository,gpsearchers,yahpo]->-r requirements.txt (line 1)) (1.1.3)
Collecting typing-extensions
  Downloading typing_extensions-4.4.0-py3-none-any.whl (26 kB)
[...]

Issue #388, opened 2022-10-18
Fix remote launching for SageMaker back-end experiments


mseeger:
This is about remote launching of experiments with SageMaker backend:
https://github.com/awslabs/syne-tune/blob/main/benchmarking/commons/launch_remote_sagemaker.py
https://github.com/awslabs/syne-tune/tree/main/benchmarking/nursery/launch_sagemaker

We currently run Syne Tune via a SM estimator, using the SKLearn framework (PyTorch also does not work). This kind of works, but there are at least two glitches:
- SageMaker warm pool feature does not work (seems like sagemaker PyPI version is too old, even if listed as explicit dependency)
- `sagemaker_search` in `sagemaker_utils.py` does not work properly (lists only subset of running jobs), which does not allow
    to manage busy workers properly

mseeger:
Need to ensure that Syne Tune is run remotely in the same was as on a laptop, i.e. the latest sagemaker PyPI is used, and `sagemaker_search` works correctly.

mseeger:
For now, I'll fix the problem with `sagemaker_search` by keeping trials marked as busy if their status is not returned. Since this error seems to concern trials which have just been started, this seems the right behaviour.

Still, we should fix this difference between Syne Tune running remotely and on the laptop.

mseeger:
Note: @geoalgo  used to build a Docker image of Syne Tune in order to do remote launching. Only later, he proposed to do remote launching with a SM framework.

We may have to go back to building an image. But if so, this needs to be a robust process, which for example also works when you are on a SM notebook instance.

mseeger:
Dealt with in #482

PR #482, opened 2022-12-22
Streamline how sagemaker backend uses pre-build containers

jgolebiowski:
Changes:
1. Moved all estimator framework definitions to a new file estimators.py so we can easily update framework versions to make sure the most up-to-date containers are used everywhere.
2. Updated the usage of estimators to rely on basic_cpu_instance/instance_sagemaker_estimator where possible to make sure we can update the container used for it when needed
3. Added constats for instance codes to make sure they can be easily updated across all code.
4. Enabled warm pooling for remote launching. The new PyTorch container ships with sagemaker 122 which supports warm pooling
5. Emabled passing logs client to the sagemaker_search function to be consistent with the sagemaker client. This change was needed for testing in issue 388
6. Removed unused parameters from height_with_cost_default_params (hard-coded instance/sagemaker estimator) Small changes:
Updated the jobname in hpo_main_common.py to make sure it aligs with sagemaker requirements (no '_' characters). Required to run tests for this change.

*Issue #, if available:*
https://github.com/awslabs/syne-tune/issues/465
https://github.com/awslabs/syne-tune/issues/388


benchmarking/commons/benchmark_definitions/lstm_wikitext2.py
```diff
@@ -18,16 +18,20 @@
     METRIC_NAME,
     RESOURCE_ATTR,
 )
+from syne_tune.backend.sagemaker_backend.estimators import (
+    DEFAULT_GPU_INSTANCE_SMALL,
```
mseeger, line 22: Better names would be DEFAULT_GPU_INSTANCE_1GPU and DEFAULT_GPU_INSTANCE_4GPU

benchmarking/commons/launch_remote_sagemaker.py
```diff
@@ -30,9 +31,11 @@
 from benchmarking.commons.hpo_main_sagemaker import parse_args
 from benchmarking.commons.utils import (
     message_sync_from_s3,
-    basic_cpu_instance_sagemaker_estimator,
     find_or_create_requirements_txt,
 )
+from syne_tune.backend.sagemaker_backend.estimators import (
```
mseeger, line 36: It is a bit funny to have this in `sagemaker_backend`, because it is used for remote launching.

benchmarking/nursery/benchmark_automl/launch_remote.py
```diff
@@ -14,12 +14,14 @@
 from pathlib import Path
```
mseeger, line 14: David asked me to leave this code unchanged (everything in `nursery/benchmark_automl`)

examples/launch_height_sagemaker.py
```diff
@@ -56,14 +56,10 @@
 
     trial_backend = SageMakerBackend(
         # we tune a PyTorch Framework from Sagemaker
-        sm_estimator=PyTorch(
+        sm_estimator=basic_cpu_instance_sagemaker_estimator(
```
mseeger, line 59: You may have to argue with David. He wanted all these examples to be "elementary".

examples/launch_height_sagemaker_checkpoints.py
```diff
@@ -13,17 +13,18 @@
 import logging
 from pathlib import Path
 
-from sagemaker.sklearn import SKLearn
-
+from syne_tune import Tuner, StoppingCriterion
 from syne_tune.backend import SageMakerBackend
+from syne_tune.backend.sagemaker_backend.estimators import (
+    sklearn_estimator,
```
mseeger, line 19: It would be better to use `basic_cpu_instance_sagemaker_estimator` here

syne_tune/backend/sagemaker_backend/estimators.py
```diff
@@ -0,0 +1,135 @@
...
+        framework_version="1.9.0",
+        py_version="py38",
+        **estimator_kwargs,
+    )
+
+
+sagemaker_estimator = {
+    "PyTorch": pytorch_estimator,
+    "HuggingFace": huggingface_estimator,
+    "BasicEstimator": instance_sagemaker_estimator,
```
mseeger, line 129: Drop "Estimator". Also, this is a bit vague. Maybe "BasicCPU" ?

mseeger:
Given my changes, I'd accept this, but please check with @geoalgo about whether the scripts in `examples/` should make use of your functions. He pushed back on me doing this before.

mseeger:
BTW: From these issues and PR, I sense that the underlying problem here is that since SageMaker framework containers are not always up-2-date, we run into issues that certain things are available as dependencies (including AWS features itself) to us locally, but not when we do remote launching.

In this sense, we should reopen the discussion about whether building our own image for remote launching is not the better option. I'd love somebody (maybe @geoalgo ?) to list pros and cons.

examples/launch_height_sagemaker.py
geoalgo, line 59: Regarding this one in all change bellow in examples, I believe the following would be better:

```python
from syne_tune.backend.sagemaker_backend.estimators import  pytorch_latest_framework, pytorch_latest_py_version
...
sm_estimator=PyTorch(
  entry_point=str(entry_point),
  instance_type="ml.m5.large",
  instance_count=1,
  framework_version=pytorch_latest_framework,
  py_version=pytorch_latest_py_version,
  ...
)
```

the reason is that otherwise, the examples hides a bunch of things to a user of sagemaker (he may have no idea that pytorch estimator is used or which instance type is used whereas it is quite critical to get this)

geoalgo: > BTW: From these issues and PR, I sense that the underlying problem here is that since SageMaker framework containers are not always up-2-date, we run into issues that certain things are available as dependencies (including AWS features itself) to us locally, but not when we do remote launching.

SageMaker frameworks are updated quite frequently, however one gets the version specified (the version of frameworks were for instance update to move away from python 3.6). As long as the examples runs, I don't think it is a problem.

Regarding building our containers, we can do it but it adds operational pain which is why we removed this as the default option.

jgolebiowski: OK makes sense, changed.

syne_tune/backend/sagemaker_backend/estimators.py
jgolebiowski: I wanted this to be a basic container we can use but maybe you are right that its too vague to be useful. Switched to 

```
    "BasicCPU": basic_cpu_instance_sagemaker_estimator,
```

examples/launch_height_sagemaker_checkpoints.py
jgolebiowski: Done

benchmarking/nursery/benchmark_automl/launch_remote.py
jgolebiowski: Ok I se the point of not updating the versions to make sure the results from the paper are reproducible. However, we are updaing the syne-tune version (since code is pulled from local source rather than fixed version) so I think we should update the container used for execution or otherwise it might not be able to install up-to-date dependencies.

benchmarking/commons/launch_remote_sagemaker.py
jgolebiowski: Ok makes sense, moved it to remote so we can disentangle the backend code and the plumbing code

benchmarking/commons/benchmark_definitions/lstm_wikitext2.py
jgolebiowski: Ok, makes sense. Changed

examples/launch_height_sagemaker.py
mseeger: Nit: Either capitalize these constants, or make them functions.

mseeger (approved):

examples/launch_height_sagemaker.py
jgolebiowski: I would vote for SM containers untill we run into a dependency issue we cannot fix by updating the container version (assuming 1-2 weeks leeway for SM containers to catch up). I would rather keep it simple untill we have no choice since adding any maintanance overhead slows down developent.

PR #484, opened 2022-12-29
Refactoring of model-based searchers and schedulers, includes fixes of HyperTune and SyncMOBSTER

mseeger:
*Issue #, if available:*

*Description of changes:*

Major refactoring/simplification of model-based searchers and schedulers, along with bug fixes for HyperTune and SyncMOBSTER. In detail:

- Allow to ignore BB hash checking as an option
- nursery/benchmark_yahpo: Also allow for MOBSTER
- Extend developer tutorial corresponding to refactoring
- Replacing logging by logger in some code
- Remove defaults in baselines.HyperTune
- Refactoring: Collect common code in TrialSchedulerWithSearcher, which extends ResourceLevelsScheduler, and makes sure self.searcher.configure_scheduler is called
- Refactoring: MultiFidelitySchedulerMixin as API for multi-fidelity schedulers, allows for configure_scheduler implementations of searchers
- Special case "hypertune" searcher in HyperbandScheduler: Number of brackets is needed for construction
- HyperbandScheduker: rung_levels does not include max_resource_level
- Simplify RungSystem subclasses: Always pass max_t, not contained in rung_levels
- Rewrite configure_scheduler methods to use TrialSchedulerWithSearcher, MultiFidelitySchedulerMixin instead of specific classes
- SynchronousHyperbandCommon: Common code for constructing SynchronousHyperbandScheduler, DifferentialEvolutionHyperbandScheduler

### DIFF
--- a/docs/source/tutorials/developer/new_searcher.rst
+++ b/docs/source/tutorials/developer/new_searcher.rst
@@ -74,6 +74,8 @@ some best practices for linking a new searcher into the factory:
   being specified by the user. You will always have the fields contributed by
   the generic schedulers, and for all others, your code should ideally come with
   sensible defaults.
+* Make sure to implement the ``configure_scheduler`` method of your new searcher,
+  restricting usage to supported scheduler types.
 
 Contribute your Extension
 -------------------------
--- a/syne_tune/optimizer/schedulers/searchers/bayesopt/models/gp_model.py
+++ b/syne_tune/optimizer/schedulers/searchers/bayesopt/models/gp_model.py
@@ -433,22 +433,25 @@ class GaussProcModelFactory(TransformerModelFactory):
         )
 
     def configure_scheduler(self, scheduler):
-        from syne_tune.optimizer.schedulers.hyperband import HyperbandScheduler
+        from syne_tune.optimizer.schedulers.multi_fidelity import (
+            MultiFidelitySchedulerMixin,
+        )
 
         if isinstance(
             self._gpmodel, (IndependentGPPerResourceModel, HyperTuneJointGPModel)
         ):
-            assert isinstance(scheduler, HyperbandScheduler), (
+            assert isinstance(scheduler, MultiFidelitySchedulerMixin), (
                 "gpmodel of type IndependentGPPerResourceModel requires "
-                + "HyperbandScheduler scheduler"
+                "MultiFidelitySchedulerMixin scheduler"
             )
             # Likelihood of internal model still has to be created (depends on
-            # rung levels of scheduler). Note that ``max_t`` must be included
-            max_t = scheduler.max_t
-            if scheduler.rung_levels[-1] == max_t:
+            # rung levels of scheduler). Note that ``max_resource_level`` must be
+            # included
+            max_resource_level = scheduler.max_resource_level
+            if scheduler.rung_levels[-1] == max_resource_level:
                 rung_levels = scheduler.rung_levels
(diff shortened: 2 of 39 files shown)


mseeger:
This PR is not just refactoring, but also fixes some bugs:
- Defaults of `baselines.HyperTune`
- `baselines.SyncMOBSTER` did not work

docs/source/tutorials/developer/new_searcher.rst
```diff
@@ -74,6 +74,8 @@ some best practices for linking a new searcher into the factory:
   being specified by the user. You will always have the fields contributed by
   the generic schedulers, and for all others, your code should ideally come with
   sensible defaults.
+* Make sure implement the ``configure_scheduler`` method of your new searcher,
```
wistuba, line 77: you/to implement

syne_tune/optimizer/schedulers/searchers/bayesopt/models/gp_model.py
```diff
@@ -433,22 +433,25 @@ def _draw_fantasy_values(self, state: TuningJobState) -> TuningJobState:
...
-            max_t = scheduler.max_t
-            if scheduler.rung_levels[-1] == max_t:
+            # rung levels of scheduler). Note that ``max_resource_level`` must be
+            # included
+            max_resource_level = scheduler.max_resource_level
+            if scheduler.rung_levels[-1] == max_resource_level:
                 rung_levels = scheduler.rung_levels
             else:
-                rung_levels = scheduler.rung_levels + [max_t]
+                rung_levels = scheduler.rung_levels + [max_resource_level]
```
wistuba, line 454: I see max_t and max_resource_level used. What's the difference and why do we need both?

wistuba (commented):
Do you think it's worth changing everything to allow ignoring the BB hash? We probably need to do this logic soon again.

docs/source/tutorials/developer/new_searcher.rst
mseeger: OK

mseeger:
Hi Martin, do you mean the option `ignore_blackbox_hash` I introduced in `benchmarking`? This is indeed something that can be removed once BB hashing is bomb-proof.

I think it would be better to switch hash checking off by default until this is properly fixed, but the PR on that one is still open. I need to be able to do my work.

syne_tune/optimizer/schedulers/searchers/bayesopt/models/gp_model.py
mseeger: This is a bit of a legacy issue. `max_t` comes from Ray Tune code which was the original source for `HyperbandScheduler`. Since resource is quite different from time, I renamed this in later MF schedulers, but did not dare to change it for `HyperbandScheduler`, as people got used to it. In fact, it is almost always better to pass `max_resource_attr` to your scheduler ...

mseeger: To answer your question: `max_t` is a synonym for `max_resource_level`, which is used in `HyperbandScheduler`

wistuba: Would be great if we could get rid of max_t eventually (maybe 0.4)

wistuba (dismissed):

wistuba (approved):

PR #485, opened 2023-01-02
Management of random seeds in benchmarking

mseeger:
*Issue #, if available:*

*Description of changes:*
Experiments which compare methods against each other, should share the same "master random seed", to which the repetition seed (0, 1, ...) is added. The current code just uses the repetition seed.

This is a problem of low entropy. Also, we want to pull together "real benchmark" data into new surrogate blackboxes, so we need to use different random seeds.

On the other hand, when comparing methods against each other, the same random seed (per repetition) should be used, which reduces variance between them (this is called "paired comparison"). So, the same master random seed should be used across an experiment, at least for the same benchmark.

### DIFF
--- a/benchmarking/commons/utils.py
+++ b/benchmarking/commons/utils.py
@@ -12,6 +12,7 @@
 # permissions and limitations under the License.
 from pathlib import Path
 from typing import Optional
+import numpy as np
 
 from syne_tune.util import s3_experiment_path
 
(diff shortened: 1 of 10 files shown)


benchmarking/commons/utils.py
```diff
@@ -142,3 +143,14 @@ def find_or_create_requirements_txt(
     else:
         fname = files[0]
     return fname
+
+
+def get_master_random_seed(random_seed: Optional[int]) -> int:
+    if random_seed is None:
+        random_seed = np.random.randint(0, 2**32)
```
jgolebiowski, line 150: NIT: Python runtimes have maxint of `2**63-1` ([long renamed as int](https://peps.python.org/pep-0237/)) so its fine but if we want to be compatible with standard 32bit int, I think we should use `[0-2**31) == [0-maxint + 1)`

jgolebiowski (approved):
LGTM, NIT comment can be ignored to move quickly.

benchmarking/commons/utils.py
mseeger: Hmm, but I think the random seed range of numpy.random.RandomState is only 32 bit.

BTW: RandomState has been replaced by something else recently in Python, but a first glance looked a bit complex to me, so I left it. Feel free to look into this, I am not bound to RandomState, just we should use the same thing all over the place.

jgolebiowski: Good point, I think random state is indeen unsigned int so 0-2**32 is great.

jgolebiowski (approved):

PR #487, opened 2023-01-04
Allow get_config to return the same config more than once

mseeger:
*Issue #, if available:* 415

*Description of changes:*
Introduces flag `allow_duplicates` to searchers, which allows to return the same config more than once. Also contains a new test on searchers, whether they properly implement `allow_duplicates=False` (the default).


syne_tune/optimizer/schedulers/searchers/bore/bore.py
```diff
@@ -122,6 +127,11 @@ def __init__(
 
         self._hp_ranges = make_hyperparameter_ranges(self.config_space)
         self._excl_list = ExclusionList.empty_list(self._hp_ranges)
+        self._allow_duplicates = allow_duplicates
+        # Maps ``trial_id`` to configuration. This is used to blacklist
```
mseeger, line 131: This is a technicality:
- If not `allow_duplicates`, then `_excl_list` collects all configs suggested, so none is suggested 2x
- If `allow_duplicates`, then `_excl_list` only collects configs for which the trial failed. In this case, we need to know the config for a trial, because `evaluation_failed` only gets the `trial_id`.


The same code is used in other places, same comment applies.

syne_tune/optimizer/schedulers/searchers/bore/bore.py
```diff
@@ -293,5 +304,24 @@ def _update(self, trial_id: str, config: dict, result: dict):
...
+                self._config_for_trial_id[trial_id] = config
+            else:
+                logger.warning(
+                    f"register_pending called for trial_id {trial_id} without passing config"
+                )
+
+    def evaluation_failed(self, trial_id: str):
+        if self._allow_duplicates and trial_id in self._config_for_trial_id:
+            # Blacklist this configuration
+            self._excl_list.add(self._config_for_trial_id[trial_id])
```
mseeger, line 324: Only get `trial_id` here, but need to know the config. This  is why `_config_for_trial` is needed, only if `allow_duplicates`.

syne_tune/optimizer/schedulers/searchers/random_grid_searcher.py
```diff
@@ -136,12 +147,45 @@ def _update(self, trial_id: str, config: dict, result: dict):
...
+                self._config_for_trial_id[trial_id] = config
+            else:
+                logger.warning(
+                    f"register_pending called for trial_id {trial_id} without passing config"
+                )
+
+    def evaluation_failed(self, trial_id: str):
+        if self._allow_duplicates and trial_id in self._config_for_trial_id:
+            # Blacklist this configuration
+            self._excl_list.add(self._config_for_trial_id[trial_id])
```
wistuba, line 167: I've seen these functions now a couple of times. Is it possible to move it up in the class hierarchy or use a utils function instead?

tst/schedulers/test_searchers.py
```diff
@@ -0,0 +1,135 @@
...
+        (HyperTune, True),
+        (SyncHyperband, True),
+        (DEHB, True),
+        (BOHB, True),
+        (SyncBOHB, True),
+        (BORE, False),
+        (KDE, False),
+    ],
+)
+def test_allow_duplicates_or_not(scheduler_cls, multifid):
```
wistuba, line 58: seems natural to move `allow_duplicates` here

syne_tune/optimizer/schedulers/searchers/random_grid_searcher.py
mseeger: Yes, doing that

tst/schedulers/test_searchers.py
mseeger: What do you mean?

wistuba: add another `parametrize` line with `allow_duplicates` before the function

mseeger:
Hi @wistuba , I am still working on this. I'd like to extend the unit test to the case where all trials fail, which should filter them out even if `allow_duplicates == True`. But this does not work right now.

mseeger:
OK, I am done with this now

syne_tune/optimizer/schedulers/synchronous/hyperband.py
```diff
@@ -290,6 +290,9 @@ def _suggest(self, trial_id: int) -> Optional[TrialSuggestion]:
             config = self.searcher.get_config(trial_id=str(trial_id))
             if config is not None:
                 config = cast_config_values(config, self.config_space)
+                self.searcher.register_pending(
```
wistuba, line 293: are these changes related to the feature?

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -450,3 +451,72 @@ def _restore_from_state(self, state: dict):
 
     def set_random_state(self, random_state: np.random.RandomState):
         self.random_state = random_state
+
+
+class FilterDuplicatesMixin:
```
wistuba, line 456: the mixin should not have functions overlapping with BaseSearcher

wistuba (changes requested):
The multi-inheritance is not a good choice here. What you could do instead is make it part of the base searcher. Searchers who don't support it can simply hard code the choice of allow_duplicates to False.

syne_tune/optimizer/schedulers/synchronous/hyperband.py
mseeger: Yes, this is something that was missing and came to light because of the new test

syne_tune/optimizer/schedulers/searchers/searcher.py
mseeger: Yes, so this would basically be multiple inheritance then.

The problem with moving this into BaseSearcher:
- BaseSearcher is really supposed to be just an API. David would rightly reject putting code in there
- Some of the searchers do this `allow_duplicates` in a different way

Is multiple inheritance really so evil?

mseeger: OK, I think what I can do, is I can integrate this code into `SearcherWithRandomSeed`. This way, the `BaseSearcher` remains clean.

mseeger: Or better: make it a subclass of that one, this is best

syne_tune/optimizer/schedulers/searchers/bore/bore.py
```diff
@@ -85,12 +82,16 @@ def __init__(
         random_prob: Optional[float] = None,
         init_random: Optional[int] = None,
         classifier_kwargs: Optional[dict] = None,
+        allow_duplicates: Optional[bool] = None,
         **kwargs,
     ):
+        if allow_duplicates is None:
+            allow_duplicates = False
```
wistuba, line 89: deal with this in parent class

syne_tune/optimizer/schedulers/searchers/bore/bore.py
```diff
@@ -246,7 +244,8 @@ def wrapper(x):
                 f"config={config}] "
                 f"optimization time : {opt_time}"
             )
-            self._excl_list.add(config)  # Should not be suggested again
+            if not self.allow_duplicates:
+                self._excl_list.add(config)  # Should not be suggested again
```
wistuba, line 248: can't we delegate to parent class as well?

syne_tune/optimizer/schedulers/searchers/kde/kde_searcher.py
```diff
@@ -96,12 +90,16 @@ def __init__(
         num_candidates: Optional[int] = None,
         bandwidth_factor: Optional[int] = None,
         random_fraction: Optional[float] = None,
+        allow_duplicates: Optional[bool] = None,
         **kwargs,
     ):
+        if allow_duplicates is None:
+            allow_duplicates = False
```
wistuba, line 97: same

wistuba (commented):
There are few lines of code that might be suited in the newly added class.

syne_tune/optimizer/schedulers/searchers/bore/bore.py
mseeger: OK

syne_tune/optimizer/schedulers/searchers/bore/bore.py
mseeger: I'll  try something

wistuba (dismissed):

wistuba (approved):

PR #488, opened 2023-01-05
Make that RandomSearcher and GPFIFOSearcher suggest the same initial …

mseeger:
…(random) configs

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/syne_tune/optimizer/schedulers/searchers/gp_fifo_searcher.py
+++ b/syne_tune/optimizer/schedulers/searchers/gp_fifo_searcher.py
@@ -127,7 +127,7 @@ def check_initial_candidates_scorer(initial_scoring: Optional[str]) -> str:
 class ModelBasedSearcher(SearcherWithRandomSeed):
     """Common code for surrogate model based searchers
 
-    If ``num_initial_candidates > 0``, initial configurations are drawn using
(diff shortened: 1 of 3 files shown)


mseeger:
Hi David, it is a good idea to try and make sure that *all* schedulers which do random draws, also return the same configs, at least initially.

Since the ones you mentioned also inherit from `SearcherWithRandomSeed`, this may already be the case, actually. It is just that they use their own code to do random draws, instead of using `HyperparameterRanges`. Maybe the easiest would be to refactor them, so they use the same underlying code.

mseeger:
I am a bit against fixing the initial configs at the benchmarking level, because this will remove variability from the benchmark that I think we want to have, otherwise the risk of overfitting on a benchmark becomes higher. What I mean is that when I use a benchmark in two different experiments, I'd like the initial (random) configs to be different. For the same reason, I'd be against fixing the random_seed at that level.

mseeger:
Actually, there is a nice way to do this even for external code, given that it allows for `points_to_evaluate`. We create a wrapper that samples the desired number of initial configs uo front, mixes it with `points_to_evaluate`, and then use the so-extended `points_to_evaluate`.

PR #489, opened 2023-01-08
Different searchers suggest same initial random configs. New methods …

mseeger:
…in baselines

*Issue #, if available:*

*Description of changes:*
Ensures that all searchers return the same random initial configs when started with the same seed.
Also:
- New classes in baselines.py
- Split searcher.py (got too large)
- Make sure that BOHB schedulers do not return duplicates
- Fixes computation of rung levels, which relied on floating point accuracy (resolved now)

### DIFF
--- a/syne_tune/optimizer/baselines.py
+++ b/syne_tune/optimizer/baselines.py
@@ -11,7 +11,6 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 from typing import Dict, Optional
-import numpy as np
 import logging
 import copy
 
@@ -31,10 +30,24 @@ from syne_tune.optimizer.schedulers.synchronous import (
 from syne_tune.optimizer.schedulers.transfer_learning import (
     TransferLearningTaskEvaluations,
 )
+from syne_tune.optimizer.schedulers.random_seeds import RandomSeedGenerator
 from syne_tune.try_import import (
     try_import_blackbox_repository_message,
     try_import_bore_message,
+    try_import_botorch_message,
 )
+
+
+def _random_seed_from_generator(random_seed: int) -> int:
+    """
+    This helper makes sure that a searcher within
+    :class:`~syne_tune.optimizer.schedulers.FIFOScheduler` is seeded in the same
+    way whether it is created by the searcher factory, or by hand.
+
+    :param random_seed: Random seed for scheduler
+    :return: Random seed to be used for searcher created by hand
+    """
+    return RandomSeedGenerator(random_seed)()
 
 
 class RandomSearch(FIFOScheduler):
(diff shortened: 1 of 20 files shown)


syne_tune/optimizer/baselines.py
```diff
@@ -31,12 +30,26 @@
...
 )
+from syne_tune.optimizer.schedulers.random_seeds import RandomSeedGenerator
 from syne_tune.try_import import (
     try_import_blackbox_repository_message,
     try_import_bore_message,
+    try_import_botorch_message,
 )
 
 
+def _random_seed_from_generator(random_seed: int) -> int:
```
mseeger, line 41: To explain: Inside FIFOScheduler, random_seed is used to create a seed generator, which is used to obtain seeds for different components. The first one is the random_state of the searcher.

This works only if the searcher is created inside FIFOScheduler construction (if the `searcher` argument is a string). So, the helper here is for cases where the searcher object is created up front.

An alternative would be to use the same seed for the searcher and all the FIFOScheduler components. But that results in bad behaviour, such as the random configs being sampled to optimize the acquisition function always include all the configs which were suggested already before.

syne_tune/optimizer/baselines.py
```diff
@@ -375,6 +420,24 @@ def __init__(
...
+    kwargs: dict,
+) -> dict:
+    searcher_kwargs = dict(
+        config_space=config_space,
+        metric=metric,
+        points_to_evaluate=kwargs.get("points_to_evaluate"),
+    )
+    search_options = kwargs.get("search_options", dict())
+    searcher_kwargs.update(search_options)
+    if random_seed is not None:
```
mseeger, line 436: This is how random_seed for searcher relates to overall random seed passed to scheduler, if the searcher is created up front

syne_tune/optimizer/schedulers/hyperband.py
```diff
@@ -974,7 +968,9 @@ def hyperband_rung_levels(
         ), f"max_t ({max_t}) must be greater than grace_period ({grace_period})"
         rf = reduction_factor
         min_t = grace_period
-        max_rungs = int(np.log(max_t / min_t) / np.log(rf) + 1)
+        max_rungs = 0
```
mseeger, line 971: This created a nasty bug, which works on Mac, but fails on CI system, I think due to round-off.

syne_tune/optimizer/schedulers/searchers/__init__.py
```diff
@@ -17,22 +17,24 @@
 from syne_tune.optimizer.schedulers.searchers.searcher import (  # noqa: F401
     BaseSearcher,
     SearcherWithRandomSeed,
-    RandomSearcher,
-    GridSearcher,
     impute_points_to_evaluate,
     extract_random_seed,
 )
+from syne_tune.optimizer.schedulers.searchers.random_grid_searcher import (  # noqa: F401
```
mseeger, line 23: I split searcher.py into two parts, it was getting too large.

syne_tune/optimizer/schedulers/searchers/bore/bore.py
```diff
@@ -20,10 +20,16 @@
...
 
-from syne_tune.optimizer.schedulers.searchers.searcher import SearcherWithRandomSeed
+from syne_tune.optimizer.schedulers.searchers.searcher import (
+    SearcherWithRandomSeed,
+    sample_random_configuration,
+)
 from syne_tune.optimizer.schedulers.searchers.utils.hp_ranges_factory import (
     make_hyperparameter_ranges,
 )
+from syne_tune.optimizer.schedulers.searchers.bayesopt.tuning_algorithms.common import (
```
mseeger, line 30: I also make sure that Bore does not suggest the same config 2x.

syne_tune/optimizer/schedulers/searchers/bore/bore.py
```diff
@@ -232,12 +233,14 @@ def wrapper(x):
                     ind = np.array(values).argmin()
                     config = X[ind]
 
-        opt_time = time.time() - start_time
-        logging.debug(
-            f"[Select new candidate: "
-            f"config={config}] "
-            f"optimization time : {opt_time}"
-        )
+        if config is not None:
```
mseeger, line 236: config is None only when the config space has been exhausted

syne_tune/optimizer/schedulers/searchers/kde/kde_searcher.py
```diff
@@ -16,11 +16,20 @@
...
+    sample_random_configuration,
+)
 import syne_tune.config_space as sp
 from syne_tune.optimizer.schedulers.searchers.bayesopt.utils.debug_log import (
     DebugLogPrinter,
 )
+from syne_tune.optimizer.schedulers.searchers.utils.hp_ranges_factory import (
+    make_hyperparameter_ranges,
+)
+from syne_tune.optimizer.schedulers.searchers.bayesopt.tuning_algorithms.common import (
```
mseeger, line 30: I make such these searchers do not suggest the same config 2x

syne_tune/optimizer/schedulers/searchers/random_grid_searcher.py
```diff
@@ -0,0 +1,358 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mseeger, line 1: Was previously in searcher.py, no changes

syne_tune/optimizer/schedulers/hyperband.py
wesk: Nice catch

wesk (approved):

PR #497, opened 2023-01-10
lock numpy version to <1.24.0

wesk:
*Issue #, if available:* N/A

*Description of changes:*

## Problem

Running the following command:
```
python examples/launch_height_ray.py
```

...results in the following error: `AttributeError: module 'numpy' has no attribute 'int'`

This error is because newer versions of numpy deprecate the `np.int` alias: https://numpy.org/doc/stable/release/1.24.0-notes.html and code in the ray side or the syne tune side has not been updated yet to use the new syntax.

## Solution

As an immediate mitigation, limit the version of numpy to less than `1.24.0` until the code can be updated.

### DIFF
--- a/benchmarking/nursery/benchmark_automl/requirements.txt
+++ b/benchmarking/nursery/benchmark_automl/requirements.txt
@@ -1,7 +1,7 @@
 syne-tune[extra]
 tqdm
 coolname
-numpy
+numpy>=1.16.0, <1.24.0
 pandas
 fastparquet
 s3fs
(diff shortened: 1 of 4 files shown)


mseeger:
Note that our code (Syne Tune) does not use np.int or np.float (anymore). It is just (apparently) Ray Tune

mseeger:
There is also `container/source/requirements.txt`.

I did something like `grep -r numpy | grep requirement`

mseeger (commented):
Missing `container/source/requirements.txt`

mseeger (approved):

PR #502, opened 2023-01-16
Update RemoteLauncher to use SM default containers as default.

jgolebiowski:
*Description of changes:*
1. Updated RemoteLauncher to use PyTorch container instead of custom syne-tune build one. The syne-tune container can still be used when passed as custom uri and the helper function to geerate the customr uri remains in the file


mseeger:
Sorry, but I am really not sure we need this.

mseeger:
What I mean is having this will create even more confusion.

At present, there are two ways to launch experiments remotely:
- Using `RemoteLauncher`, which builds a container. This is used in some of our experiments
- All the logic implemented in `benchmarking`, for which there is a tutorial and which is battle tested

So, if anybody wants to run many experiments in parallel, not using a container, they can use what is in `benchmarking`, rather than again copy and pasting examples, which we all need to maintain (this includes random seed management, loading data from the right place for plotting, managing meta data, etc).

And if somebody wants to use a container for Syne Tune (which can have advantages, for example once the container is there, experiments get launched a little faster), they can use `RemoteLauncher`.

mseeger:
I am really somewhat puzzled that people do not simply look at what is there in `benchmarking`, and instead insist on reinventing the wheel.

I wrote a tutorial, I gave a talk about it, I honestly do not know what else I should do.

mseeger:
The task I entered in the backlog was about changing the examples which current use `RemoteLauncher`, and make them use the code in `benchmarking/commons`.

jgolebiowski:
I see your point, I agree that the extra example with manual launching of remote tuning is probably not needed. The reasoning there was to explain how the remote launching code (in benchmarking or in RemoteLauncher) is build but maybe tutorials are a better place for this.

Regarding the remote launcher - I think that having it run only with custom container is a little limiting - we should be able to use that functionality with a default sagemaker containers. Why limit the capability of the code that already exists to niche (custom containers) users. When we enable default sagemaker containers with RemoteLauncher, the existing examples need no further changes.

With that in mind, I am happy to add a new example that shows how to use banchmarking tools to run. Is that hat you have in mind?

mseeger:
OK, so my favourite would be this:
- Leave `RemoteLauncher` as it is, namely build a container
- If anybody wants to do remote launching via SM framework, they should use `benchmarking/commons`. There could be extra examples showing how to use this, but I believe the tutorial is enough
- If somebody really wants to do this by hand, not using `benchmarking/commons`, they can have a look at the code there, which is very clean. But I really prefer people just using this, instead of writing their own scripts

mseeger:
Now, about the examples which currently use `RemoteLauncher`. If they can be done using `benchmarking/commons`, I'd prefer that.

Maybe this cannot be done, because benchmarking stuff is too specific. I'd doubt this, because it supports all schedulers and all benchmarks.

But suppose it is felt that this is not pure enough. In that case, we could refactor these examples so they use the PyTorch framework and do not depend on the code in benchmarking.

But to not confuse users, we would point to the benchmarking as a simpler way (even if less "elementary") of doing these things.

mseeger:
What I really would like to avoid is to have another way of doing what is already in `benchmarking/commons`, being rolled out across our examples. This would be duplicating efforts and be real confusing.

jgolebiowski:
Thanks for the context. I understand that we want to gently push customers to use the solution form `benchmarking/commons` as a default which makes sense. My proposal is the following:
1/ Make the curent examples (defaults for running jobs remotely) based on `benchmarking/commons`
2/ add a new example for using `RemoteLauncher` marketing it as an alternative (non-default) way of running remote jobs. In this point I prefer to retain the change to `RemoteLauncher` that makes it compatible with both SM default and custom-build containers. I feel that limiting it to custom containers only just makes it hard to use w.o. much added benefit.

mseeger:
As for 2/, you need to argue with David, he wrote the `RemoteLauncher`.

But note that this code not only uses a container, it also relies on serialization of the Tuner object with all its dependent classes (a closure), which gave us quite a bit of trouble.

The code in `benchmarking/commons`, which is really also from David (but brushed up a lot by myself), circumvents this step as well, but you need two scripts instead of one.

You can therefore think of `benchmarking/commons` as a 2nd generation, and `RemoteLauncher` the first, which is harder to maintain.

I'd personally be most happy if users stopped using `RemoteLauncher` altogether, because of its dependence on dill serialization. So, this is why I am not sure we should invest much in it anymore.

jgolebiowski:
I see, regarding `RemoteLauncher` we should either update it to use SM defaults or deprecate it (at least mark for deprecation) if we dont want to support it. Otherwise testing for custom-build containers is a pain point. Happy to discuss in our next sync.

jgolebiowski:
Following offline discussion the plan is as following:
1. Repurpose this CR to enable default SM containers in RemoteLauncher
2. Followup with a second CR which will update examples to use benchmarking/commons. At the end of the work we will have one example for RemoteLauncher and two examples for benchmrking - adding custom code to benchmarks and running existing benchmarks.

examples/launch_height_sagemaker.py
```diff
@@ -59,6 +60,8 @@
     scheduler = RandomSearch(
         config_space, mode=mode, metric=metric, random_seed=random_seed
     )
+    if "AWS_DEFAULT_REGION" not in os.environ:
+        os.environ["AWS_DEFAULT_REGION"] = "us-west-2"
```
geoalgo, line 64: Is that needed? I think it was working for me but perhaps it is needed in case you dont have a aws config defined?

geoalgo (approved):

examples/launch_height_sagemaker.py
jgolebiowski: Exactly what you mention. I can remove it and add a comment for users to add `AWS_DEFAULT_REGION` to their environment if you think thats better.

PR #507, opened 2023-01-21
Refactoring and renaming of BoTorchSearcher and BoTorch

mseeger:
*Issue #, if available:*

*Description of changes:*
Simplification and renaming of BoTorchSearcher. This PR will be followed by an extension of the developer tutorial, which uses this searcher as example how to wrap external code. The wrapper should therefore be up-2-date with respect to support from base classes.

Also, one deprecated BoTorch function is replaced.


syne_tune/optimizer/schedulers/searchers/botorch/botorch_searcher.py
```diff
@@ -17,34 +17,31 @@
 from syne_tune.try_import import try_import_botorch_message
 
 try:
-    import torch
+    from torch import Tensor, randn_like
     from botorch.models import SingleTaskGP
-    from botorch.fit import fit_gpytorch_model
```
mseeger, line 22: This function is deprecated

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -450,3 +451,84 @@ def _restore_from_state(self, state: dict):
 
     def set_random_state(self, random_state: np.random.RandomState):
         self.random_state = random_state
+
+
+class SearcherWithRandomSeedAndFilterDuplicates(SearcherWithRandomSeed):
```
mseeger, line 456: This is code to be introduced in #487, which will be merged before this one here.

syne_tune/optimizer/schedulers/searchers/botorch/__init__.py
```diff
@@ -16,9 +16,11 @@
 
 try:
     from syne_tune.optimizer.schedulers.searchers.botorch.botorch_searcher import (  # noqa: F401
-        BotorchSearcher,
+        BoTorchSearcher,
+        BotorchSearcher,  # deprecated
```
wesk, line 20: Can `BotorchSearcher` be removed entirely in favor of `BoTorchSearcher`?

wesk: I see -- looks like you are keeping it for backwards compatibility

wesk (approved):

syne_tune/optimizer/schedulers/searchers/botorch/__init__.py
mseeger: Yes, we can remove it after a while

Issue #397, opened 2022-10-21
Unit test for points_to_evaluate in schedulers


mseeger:
Many (or most) of our schedulers support `points_to_evaluate`. A test should cycle over all of them, query initial points, and check they are equal to `points_to_evaluate`

PR #517, opened 2023-01-24
Test points to evaluate

sighellan:
#397 

*Description of changes:*
Test that cycles over schedulers* in [baselines](https://github.com/awslabs/syne-tune/blob/main/syne_tune/optimizer/baselines.py) and checks that initial points match those given in `points_to_evaluate`.

*Only does the ones that are not for transfer learning, though I could look into adding those. 


tst/test_points_to_evaluate.py
```diff
@@ -0,0 +1,136 @@
...
+    (BoTorch, False),
+    (REA, False),
+    (ConstrainedBayesianOptimization, False),
+    (KDE, False),
+]
+
+
+@pytest.mark.timeout(20)
+@pytest.mark.parametrize("Scheduler, mul_fid", SCHEDULERS)
+def test_points_to_evaluate(Scheduler, mul_fid):
```
mseeger, line 68: No capials. Maybe `scheduler_cls`

tst/test_points_to_evaluate.py
```diff
@@ -0,0 +1,136 @@
...
+    (SyncMOBSTER, True),
+    (ASHABORE, True),
+    (BoTorch, False),
+    (REA, False),
+    (ConstrainedBayesianOptimization, False),
+    (KDE, False),
+]
+
+
+@pytest.mark.timeout(20)
```
mseeger, line 66: Is this needed? I thought the time limit is per case

tst/test_points_to_evaluate.py
```diff
@@ -0,0 +1,136 @@
...
+    # Use train_height backend for our tests
+    entry_point = str(
+        Path(__file__).parent.parent
+        / "examples"
+        / "training_scripts"
+        / "height_example"
+        / "train_height.py"
+    )
+    max_steps = 5
+    config_space = {
```
mseeger, line 79: Could you use `height_config_space` imported from `examples.training_scripts.height_example.train_height` ?

tst/test_points_to_evaluate.py
```diff
@@ -0,0 +1,136 @@
...
+        "steps": max_steps,
+        "width": randint(0, 20),
+        "height": randint(-100, 100),
+    }
+    metric = "mean_loss"
```
mseeger, line 84: Ditto, please use `METRIC_ATTR` from same module

tst/test_points_to_evaluate.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # The points we require the schedulers to start by evaluating
+    points_to_evaluate = [
+        {"steps": max_steps, "width": 3, "height": 0},
+        {"steps": max_steps, "width": 13, "height": -20},
+        {"steps": max_steps, "width": 19, "height": 30},
+    ]
+
+    # The scheduler specification varies depending on the type of scheduler
+    if Scheduler is ConstrainedBayesianOptimization:
```
mseeger, line 95: Please use `isinstance` here.

I'd also prefer if you piece together `kwargs` for the scheduler creation, and then just have

`scheduler = scheduler_cls(**kwargs)`

below

tst/test_points_to_evaluate.py
```diff
@@ -0,0 +1,136 @@
...
+            points_to_evaluate=points_to_evaluate,
+        )
+
+    # Set up tuner and run for a few evaluations
+    tuner = Tuner(
+        trial_backend=LocalBackend(entry_point=entry_point),
+        scheduler=scheduler,
+        stop_criterion=StoppingCriterion(max_num_evaluations=5),
+        n_workers=n_workers,
+        sleep_time=0.001,
```
mseeger, line 123: Won't help to make things faster. What matters are sleep time in training functon

sighellan: I tried commenting out the line and reducing the pytest timeout to 5 seconds. Then all tests failed, while with that line they all passed.

tst/test_points_to_evaluate.py
sighellan: It is per case, but they are a bit slow to run so the default 1 sec isn't enough.

tst/test_points_to_evaluate.py
sighellan: Done.

tst/test_points_to_evaluate.py
mseeger: OK

mseeger (approved):
Thanks for this contribution!

PR #520, opened 2023-01-27
run more examples as part of GitHub Actions

wesk:
*Issue #, if available:*

*Description of changes:*


examples/launch_tuning_gluonts.py
```diff
@@ -88,7 +88,7 @@
         config_space, max_t=epochs, resource_attr="epoch_no", mode="min", metric=metric
     )
 
-    max_wallclock_time = 3600 if evaluate_trials_on_sagemaker else 600
+    max_wallclock_time = 3000 if evaluate_trials_on_sagemaker else 600 # wall clock time can be increased to 1 hour for more performance
```
geoalgo, line 91: Why setting a new constant here? Setting 3000 seconds is fine with me but the comment is then not in sync.

examples/launch_huggingface_classification.py
```diff
@@ -77,7 +77,7 @@
         config_space, mode=mode, metric=metric, random_seed=random_seed
     )
 
-    stop_criterion = StoppingCriterion(max_wallclock_time=3600)
+    stop_criterion = StoppingCriterion(max_wallclock_time=3000) # wall clock time can be increased to 1 hour for more performance
```
geoalgo, line 80: Why setting a new constant here? Setting 3000 seconds is fine with me but the comment is then not in sync.

.github/workflows/run-syne-tune.yml
```diff
@@ -0,0 +1,58 @@
...
+          python-version: 3.8
+          cache: 'pip'
+      - name: Install Syne Tune
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -e '.[extra]'
+      - name: Run optional custom command
+        if: ${{ inputs.additional-command != '' }}
+        run: ${{ inputs.additional-command }}
+      - name: Run Syne Tune
```
geoalgo, line 56: The naming `Run Syne Tune` could perhaps be made more precise and self-revealing, perhaps `Run Syne Tune script` is already a bit more telling given that you use script path?

.github/workflows/fast-tests.yml
```diff
@@ -0,0 +1,122 @@
+name: Fast Tests
```
geoalgo, line 1: Could you define "fast"? For instance, "Evaluate fast tests that runs under 5 minutes" as fast can be quite subjective.

.github/workflows/integ-tests.yml
```diff
@@ -0,0 +1,211 @@
...
+
+# TODO set this up to gate new releases of Syne Tune
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+  pull_request: # remove before merging, this is for testing
+    branches:   # remove before merging, this is for testing
+      - main    # remove before merging, this is for testing
```
geoalgo, line 11: Fine for now, but I suppose we will want to limit the frequency of those (for instance weekly in addition to release) as they will be a tad expensive.

.github/workflows/integ-tests.yml
```diff
@@ -0,0 +1,211 @@
...
+  stop_all_training_jobs: 
+    runs-on: ubuntu-latest
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets.PROD_AWS_INTEG_TEST_ROLE_ARN }}
+          role-session-name: integtestsession
+          aws-region: ${{ env.AWS_DEFAULT_REGION }}
+      - name: Stop all left-over training jobs
```
geoalgo, line 31: Great work and great that you think of this case.

.github/workflows/integ-tests.yml
```diff
@@ -0,0 +1,211 @@
...
+        run: |
+          aws sagemaker list-training-jobs --status-equals InProgress > running_jobs.json
```
geoalgo, line 33: This should probably be more inclusive, for instance it should probably also include "Starting".

.github/workflows/integ-tests.yml
```diff
@@ -0,0 +1,211 @@
...
+  launch_moasha_instance_tuning:
+    needs: stop_all_training_jobs
+    uses: ./.github/workflows/run-syne-tune.yml
+    with:
+      script-path: examples/launch_moasha_instance_tuning.py
+      requires-aws-credentials: true
+    secrets: inherit
+
+  launch_nas201_transfer_learning:
+    uses: ./.github/workflows/run-syne-tune.yml
```
geoalgo, line 90: This is fine with me as is, but why not always adding `needs: stop_all_training_jobs`?
It does cost if no job needs to be stopped but it makes it allow to not forget to add this line (which can happen if you think you dont need it whereas the examples actually spawed SM jobs).

.github/workflows/integ-tests.yml
```diff
@@ -0,0 +1,211 @@
...
+    with:
+      script-path: examples/launch_height_sagemaker.py
+      requires-aws-credentials: true
+    secrets: inherit
+
+  launch_height_standalone_scheduler:
+    uses: ./.github/workflows/run-syne-tune.yml
+    with:
+      script-path: examples/launch_height_standalone_scheduler.py
```
geoalgo, line 119: Isnt this one fast?

Ok I see that some of the test appear in both this action and the "fast" action, what is the rationale for running the test twice?

geoalgo (commented):
Very nice work!

examples/launch_tuning_gluonts.py
wesk: This is to ensure that the example finishes in less than one hour. The problem was that the example was failing due to credential expiration (the creds are only valid for one hour). If the experiment runs for an hour, then the runtime goes slightly over an hour because there are a few seconds of setup time before the test reaches the example script. 

The comment is for a future reader who might want to run it for longer

examples/launch_huggingface_classification.py
wesk: Same rationale as previous instance

.github/workflows/fast-tests.yml
wesk: What I really mean is that all tests in this file must satisfy the following two properties: 
1. We want them to run against every PR
2. They do not require AWS credentials (We only run the tests that require credentials on code that has been approved and merged to main)

"Fast" seemed a reasonable way of referring to them, but some possible other names are: 
1. Integ Tests (and we then rename the "Integ Tests" suite to "End-to-end Tests", since that's more of what that other suite is).
2. Lightweight Integ Tests (in contrast with the other "Full integ tests")

If we go with this file becoming "Integ Tests" and the other one becoming "End-to-end Tests", then we can trigger this one with every PR and merge, and trigger the other one with every merge to `main`, and remove the duplicated tests between the two.

.github/workflows/integ-tests.yml
wesk: Before merging this PR, I plan to push one more commit removing these lines. Then, this suite will run 1) on merges to `main` and 2) on-demand if someone with write access to the repository clicks the button to run it from the GitHub UI.

This way, we will know as soon as a bad commit is merged to main. Whereas if we only run it before a release, we might not catch a regression in main for a long time. I think at least until we have a consistent release schedule, it makes sense to start by running them against main and seeing how well it works for us.

.github/workflows/integ-tests.yml
wesk: Good point, there's some duplication of tests between this file and the other one. I think we could do the following: 
1. Rename this file to be "End-to-end tests"
2. Remove all tests that are in the "Fast tests" from this file
3. All remaining tests in this file will always need the `needs: stop_all_training_jobs` line in this file

This way the role of each test suite is more clear

See this earlier comment also for more of what I'm thinking now, based on your comment: https://github.com/awslabs/syne-tune/pull/520#discussion_r1095523249

.github/workflows/integ-tests.yml
wesk: Good point, see comment https://github.com/awslabs/syne-tune/pull/520#discussion_r1095533413 for proposed fix

.github/workflows/integ-tests.yml
wesk: I spent a few minutes and I haven't found the syntax for how to match two statues from the AWS CLI, do you know what it would look like? In my testing, I haven't found it to be a problem, since we have enough of a limit in the account to have some buffer

.github/workflows/fast-tests.yml
geoalgo: Either name is fine, can you add a comment explaining the rationale you just described so that the context is clear for others? (for instance saying that this is a list of tests that dont require cred and runs at all PRs)

.github/workflows/integ-tests.yml
geoalgo: Just checked and it cannot be done indeed.

geoalgo (approved):

.github/workflows/fast-tests.yml
wesk: Done, updated in the next revision

geoalgo (approved):

PR #530, opened 2023-02-03
BaseSearcher for Multi-Objective Methods

aaronkl:
*Description of changes:*

Base class to implement multi-objective searcher, which requires to pass a list of metrics and modes instead of just one. 


syne_tune/optimizer/schedulers/multiobjective/multi_objective_searcher.py
```diff
@@ -0,0 +1,69 @@
...
+    def __init__(
+        self,
+        config_space: dict,
+        metrics: List[str],
+        modes: Optional[List[dict], str] = "min",
+        points_to_evaluate: Optional[List[dict]] = None,
+    ):
+        self.config_space = config_space
+        self._metrics = metrics
+        self._points_to_evaluate = impute_points_to_evaluate(
```
mseeger, line 52: I'd still rather call the superclass constructor. You can then set `_metric` and `_mode` to None if you like.

syne_tune/optimizer/schedulers/multiobjective/multi_objective_searcher.py
```diff
@@ -0,0 +1,69 @@
...
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+class BaseMultiObjectiveSearcher(BaseSearcher):
+    """
+    Base class for searchers that optimize multiple objectives simultaneously.
+    Derivatives of this class need to implement :meth:`~get_config`.
```
mseeger, line 28: No tilde

syne_tune/optimizer/schedulers/multiobjective/multi_objective_searcher.py
```diff
@@ -0,0 +1,69 @@
...
+    impute_points_to_evaluate,
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+class BaseMultiObjectiveSearcher(BaseSearcher):
+    """
+    Base class for searchers that optimize multiple objectives simultaneously.
```
mseeger, line 27: Maybe one could be more specific here. For example, constrained BO makes use of multiple objectives, but optimizes only one of them. Maybe add something like ", where not one of the metrics is the main objective, but we rather aim to find a trade-off`.

syne_tune/optimizer/schedulers/multiobjective/multi_objective_searcher.py
```diff
@@ -0,0 +1,69 @@
...
+        If ``None`` (default), this is mapped to ``[dict()]``, a single default config
+        determined by the midpoint heuristic. If ``[]`` (empty list), no initial
+        configurations are specified.
+    :param modes: Defines for each metric if it should be minimized ("min", default) or maximized
+        ("max")
+    """
+
+    def __init__(
+        self,
+        config_space: dict,
```
mseeger, line 45: `Dict[str, Any]`. We are now being more rigorous. Also `points_to_evaluate`

syne_tune/optimizer/schedulers/multiobjective/multi_objective_searcher.py
```diff
@@ -0,0 +1,69 @@
...
+        metrics: List[str],
+        modes: Optional[List[dict], str] = "min",
+        points_to_evaluate: Optional[List[dict]] = None,
```
mseeger, line 48: `Dict[str, str]`

mseeger (commented):
There is no implementation of this yet, right?

mseeger:
Currently, this abstraction has nothing beyond `BaseSearcher`, except that `metrics` and `modes` replace `metric` and `mode`. Do you foresee anything special that is needed here?

I am saying this because there are some very useful subclasses, namely `SearcherWithRandomSeed` and especially `SearcherWithRandomSeedAndFilterDuplicates`. Since we like to avoid multiple inheritance, a searcher below your new class would not have these features, even though they make sense for multi-objective just as well.

aaronkl:
The goal here is to define a clean abstraction for developers to implement multi-objective methods. The BaseSearcher allows to only pass a single str for `metric` / `mode` instead of a list of metrics. And it's unclear for users what they should pass here.  I am concerned that just deriving from BaseSearcher will make the code hard to parse, because developers might not now whether they should use `_metric` / `_mode` which could either be None or just define a single metric. 

Note that multi-objective is different to constraint optimization because a) all objectives are treated equally and b) the goal is to find a set of configurations instead of a single configuration. I agree that `ConstrainedGPFIFOSearcher` can model different metrics but at, it's core, it only focuses on a single objective and is not designed to explore multi-objective simulatenously. This would require a totally different acquisition function.

syne_tune/optimizer/schedulers/multiobjective/multi_objective_searcher.py
aaronkl: What is the benefit here of calling the superclass constructor? It might be confusing for developers if we pass None to `_metric` and `_mode` here.

aaronkl:
Regarding implementation. No there is no implementation yet, but I will provide one once we agreed on the interface

mseeger:
You should involve @wistuba in this. He gave me valuable advice on things like this.

As I said, there is powerful stuff below `BaseSearcher`, and we'd really like to retain this. At the same time, we don't want to use multiple inheritance.

For example, say you'd like to implement a MO searcher which uses GP surrogate models for several of the objectives, as `ConstrainedGPFIFOSearcher` does. You want to be able to use `GPFIFOSearcher` then, otherwise you'd have to copy all of that.

mseeger:
Maybe we can use a mixin here. But Martin should have a look at this.

syne_tune/optimizer/schedulers/multiobjective/multi_objective_searcher.py
```diff
@@ -0,0 +1,69 @@
...
+    def configure_scheduler(self, scheduler):
+        """
+        Some searchers need to obtain information from the scheduler they are
+        used with, in order to configure themselves.
+        This method has to be called before the searcher can be used.
+
+        :param scheduler: Scheduler the searcher is used with.
+        :type scheduler: :class:`~syne_tune.optimizer.schedulers.TrialScheduler`
+        """
+        if hasattr(scheduler, "metric"):
```
wistuba, line 66: `metrics`

syne_tune/optimizer/schedulers/multiobjective/multi_objective_searcher.py
```diff
@@ -0,0 +1,69 @@
...
+            self._metrics = getattr(scheduler, "metrics")
+        if hasattr(scheduler, "mode"):
```
wistuba, line 68: `modes`

wistuba (commented):
I agree with Matthias regarding the usefulness of this class and the immense problems we may face when splitting the hierarchy at the top level into two.

Maybe an unpopular opinion but what about accepting multiple metrics and modes by the `BaseSearcher`, i.e. `metric: Union[str, List[str]]`? Not sure how much effort that is to account for in the immediate child classes but the leaf child classes may decide to either support str or List[str| only.

aaronkl:
I agree with @wistuba this would be the cleanest solution, but it might require major refactoring of the code

mseeger:
No, that would be quite easy. Currently, `BaseSearcher` has `_metric` and `_mode`.

Either we allow this to be lists or single entries, using a `Union` type. Or, we introduce `_metrics` and `_modes`, and then for all searchers where one of the metrics is primary, `_metric` and `_mode` contain that one.

My slight bias is for the latter, because otherwise you either have to call it `metric` or `metrics`, and both is confusing.

mseeger:
OK, then let us do that

wistuba:
Are you also planning to rename the argument of the constructor to `metrics`? I think we could do it but it would be a massive breaking change. So I wouldn't mind leaving it at `metric`.

wesk:
+1 to avoiding a breaking change! 

> I agree with @wistuba this would be the cleanest solution, but it might require major refactoring of the code

What would be required for this? Can it be done without any breaking changes?

mseeger:
We could also say that if a single metric is needed, but `_metric` contains several ones, then I am picking the first.

As long as this is well documented, we can do anything.

mseeger:
But all in all, I feel there is then probably no need for now for a specific MO searcher base class, unless one which implements some checks or so.

aaronkl:
Ok I will open a new PR that just extends the BaseSearcher class to allow `metric`, `mode` to be also a List

PR #534, opened 2023-02-06
Run 3 tests in E2E mode only; Run `Black` as a linting step

wesk:
*Issue #, if available:*

*Description of changes:*

**Problem:** The `launch_simulated_benchmark.py`, `launch_nas201_transfer_learning.py`, and `launch_nasbench201_simulated.py` tests are failing with the following error: 

```
start: downloading compressed file
Time for downloading compressed file: 0.2871 secs
start: uncompressing and loading
Time for uncompressing and loading: 0.0004 secs
Traceback (most recent call last):
  File "examples/launch_nas201_transfer_learning.py", line 56, in <module>
    bb_dict = load_blackbox(blackbox_name)
  File "/home/runner/work/syne-tune/syne-tune/syne_tune/blackbox_repository/repository.py", line 165, in load_blackbox
    generate_blackbox_recipes[name].generate(s3_root=s3_root)
  File "/home/runner/work/syne-tune/syne-tune/syne_tune/blackbox_repository/conversion_scripts/blackbox_recipe.py", line 50, in generate
    self._generate_on_disk()
  File "/home/runner/work/syne-tune/syne-tune/syne_tune/blackbox_repository/conversion_scripts/scripts/nasbench201_import.py", line 264, in _generate_on_disk
    data = pickle.load(f)
  File "/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/bz2.py", line 172, in peek
    return self._buffer.peek(n)
  File "/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/_compression.py", line 68, in readinto
    data = self.read(len(byte_view))
  File "/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/_compression.py", line 99, in read
    raise EOFError("Compressed file ended before the "
EOFError: Compressed file ended before the end-of-stream marker was reached
Error: Process completed with exit code 1.
```

See job runs: 
* https://github.com/awslabs/syne-tune/actions/runs/4115542151/jobs/7104397493
* https://github.com/awslabs/syne-tune/actions/runs/4115542151/jobs/7104396321

This is likely due to the fact that the examples all have an external dependency (a file from a Google Drive link hosted by a researcher) and now that download is failing, potentially our requests are being throttled.

Link to line where the download happens: https://github.com/awslabs/syne-tune/blob/main/syne_tune/blackbox_repository/conversion_scripts/scripts/nasbench201_import.py#L238

**Solution:** Run this example less frequently from the E2E tests instead of the integ tests. Also, when running in E2E mode, it will download the file from our S3 bucket that we control in our integ test account, rather than the Google Drive link which we do not control. 

----

Also, this PR runs `Black` in validation mode, to ensure all PRs have consistent style.


mseeger:
The real problem here is that these scripts need the blackbox repository. `load_blackbox` first tries to load files locally, then from S3, then tries to recreate from source.

In general, we *don't* want the last thing to happen in any of the tests, except maybe some very special ones that are run manually.
These download scripts can take a really long time, and they are brittle. We want to test Syne Tune under the assumption that the files have been downloaded.

mseeger:
The real solution is not to remove these tests, but to make sure they can download the BB files from the S3 bucket.

In fact, I'd like to have quite a few more such tests, that run schedulers on tabulated blackboxes, so this needs to work.

mseeger (approved):

wesk:
In the End-to-end tests (the file that this PR moves these tests to) they will have access to AWS credentials, and will thus download the BB files from our S3 bucket in the mode that we want here. The difference is that the End-to-end tests won't run with every PR, but before every release. I can look into ways to get S3 access in the integ tests as future work -- the current limitation is that there isn't an easy way to do this that will also work for external collaborators who don't have write access to our repository.

mseeger:
OK. We'd need a bucket with read access for everybody, or at least for everybody through GitHub CI. AutoGluon has this, I think.

PR #537, opened 2023-02-07
lock dependency versions to fix `launch_rl_tuning.py` example

wesk:
*Issue #, if available:*

*Description of changes:*

Problem: `examples/launch_rl_tuning.py` started failing, see for example https://github.com/awslabs/syne-tune/actions/runs/4108215352/jobs/7088647693#step:7:1

Solution: lock to specific dependency versions

### DIFF
--- a/examples/training_scripts/rl_cartpole/requirements.txt
+++ b/examples/training_scripts/rl_cartpole/requirements.txt
@@ -1,7 +1,7 @@
-tensorboardX
+tensorboardX==2.5.1
 opencv-python
-ray[rllib]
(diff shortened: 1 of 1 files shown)


mseeger (approved):
This fixes my issue.
But I'd still be much in favour of throwing launch_rl_tuning out of the list of examples running with every push, because this will happen again.

We should not have tests in there which only fail because some esoteric dependencies change. A person pushing into a PR should not be blocked by having to upgrade esoteric dependencies. This is a waste of time.

PR #543, opened 2023-02-09
Implement DyHPO scheduler

mseeger:
*Issue #, if available:*

*Description of changes:*


syne_tune/optimizer/schedulers/hyperband.py
```diff
@@ -48,10 +52,22 @@
...
+    "cost_promotion": CostPromotionRungSystem,
+    "dyhpo": DyHPORungSystem,
+}
+
+
 _ARGUMENT_KEYS = {
     "resource_attr",
     "grace_period",
     "reduction_factor",
+    "rung_increment",
```
mseeger, line 70: With rung_increment, I am also allowing for linear spacing (which is what DyHPO uses). If reduction_factor is given, that overrules rung_increment,

syne_tune/optimizer/schedulers/searchers/dyhpo/hyperband_dyhpo.py
```diff
@@ -0,0 +1,234 @@
...
+    for promotion decisions to be separate from decisions about new configs, the
+    overall workflow is a bit tricky:
+
+    * In :meth:`FIFOScheduler._suggest`, we first call
+      :code:`promote_trial_id, extra_kwargs = self._promote_trial()`. If
+      ``promote_trial_id != None``, this trial is promoted. Otherwise, we call
+      :code:`config = self.searcher.get_config(**extra_kwargs, trial_id=trial_id)`
+      and start a new trial with this config. In most cases, :meth:`_promote_trial`
+      makes a promotion decision without using the searcher.
+    * Here, we use the fact that information can be passed from
```
mseeger, line 66: I know this does not sound simple. But the structure of `HyperbandScheduler` is pragmatic, trying to keep scheduling decisions separate from search, and the latter is model-based if anything. This works fine and is simple for lots of variants we already have.

Just DyHPO is different, in that the decision on what to resume is folded with what to start new, it is even a common decision. One could of course change the whole design of `HyperbandScheduler` to accomodate that, but that would look complicated for all the other variants.

wistuba (approved):

PR #509, opened 2023-01-23
New examples that cover running custom scripts with benchmarking/commons.

jgolebiowski:
Expanding the list of examples with running custom scripts using benchmarking/commons.

Changes:
- Added three new examples that use benchmarking/commons to run experiments
- Updated examples.rst to refer to new examples
- modified hpo_main_local.py to make benchmark a required argument without a default.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #544, opened 2023-02-09
Including additional examples using benchmarking

jgolebiowski:
Expanding the list of examples with running custom scripts using benchmarking/commons. This is an update of [PR-509](https://github.com/awslabs/syne-tune/pull/509)

Changes:

- Added new examples that use benchmarking/commons to run experiments
- Updated examples.rst to refer to new examples
- modified hpo_main_local.py to make benchmark a required argument without a default.

### DIFF
--- a/benchmarking/commons/utils.py
+++ b/benchmarking/commons/utils.py
@@ -73,7 +73,7 @@ def find_or_create_requirements_txt(
 ) -> Path:
     if requirements_fname is None:
         requirements_fname = "requirements.txt"
-    files = list(entry_point.parent.rglob("requirements*.txt"))
+    files = list(entry_point.parent.glob("requirements*.txt"))
     if len(files) == 0:
         print(
             f"Could not find {entry_point.parent}/requirements*.txt\n"
--- a/examples/launch_custom_benchmark.py
+++ b/examples/launch_custom_benchmark.py
@@ -1,0 +1,113 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+"""
+This example show how to launch a tuning job with a custom benchmark
+"""
+import argparse
+import logging
+from pathlib import Path
+from typing import Dict
+
+from benchmarking.commons.benchmark_definitions import RealBenchmarkDefinition
+from benchmarking.commons.hpo_main_sagemaker import main as main_sagemaker
+from benchmarking.commons.hpo_main_local import main as main_local
+from benchmarking.nursery.launch_sagemaker.baselines import (
+    methods as all_methods,
+    Methods,
+)
+from syne_tune.config_space import randint
+from syne_tune.remote.estimators import (
+    DEFAULT_CPU_INSTANCE_SMALL,
+)
+
+
+def height_benchmark(
+    sagemaker_backend: bool = False, **kwargs
+) -> Dict[str, RealBenchmarkDefinition]:
+    """
+    This function defines the benchmark definition based on our training script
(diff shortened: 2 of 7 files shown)


benchmarking/commons/utils.py
```diff
@@ -73,7 +73,7 @@ def find_or_create_requirements_txt(
 ) -> Path:
     if requirements_fname is None:
         requirements_fname = "requirements.txt"
-    files = list(entry_point.parent.rglob("requirements*.txt"))
+    files = list(entry_point.parent.glob("requirements*.txt"))
```
jgolebiowski, line 76: [Sagemaker frameworks](https://sagemaker.readthedocs.io/en/stable/frameworks/pytorch/using_pytorch.html) pull requirements directly from the directory of the entrypoint therefore we should only look for requirements in the entrypoint dir rather than the whole tree.

mseeger: You are right, I think I misunderstood the docs.

examples/launch_custom_benchmark.py
```diff
@@ -0,0 +1,113 @@
...
+        description="Launch a custom <train_height.py> benchmark using selected backend"
+    )
+    parser.add_argument(
+        "--backend",
+        choices=["local", "sagemaker"],
+        required=True,
+        type=str,
+        help="Backed to use for experiment",
+    )
+    parser.add_argument(
```
mseeger, line 89: You don't need this argument

examples/launch_custom_benchmark_remotely.py
```diff
@@ -0,0 +1,52 @@
...
+    entry_point = Path(__file__).parent / "launch_custom_benchmark.py"
+    single_fidelity_methods = {
+        Methods.RS: all_methods[Methods.RS],
+        Methods.BO: all_methods[Methods.BO],
+    }
+    launch_remote(
+        entry_point=entry_point,
+        methods=single_fidelity_methods,
+        benchmark_definitions=height_benchmark,
+        extra_args=[
```
mseeger, line 43: This does not work

examples/launch_custom_benchmark_remotely.py
```diff
@@ -0,0 +1,52 @@
...
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+"""
+This example show how to launch a tuning job that will orchestrated on Sagemaker rather than on your local machine.
+"""
+
+import logging
+from pathlib import Path
+
+from benchmarking.commons.launch_remote_sagemaker import launch_remote
```
mseeger, line 20: You need to use launch_remote_X and hpo_main_X when you want to run with backend X in ["local", "sagemaker"].

What you do here, either does not work, or only works by accident.

mseeger (changes requested):

examples/launch_custom_benchmark_remotely.py
jgolebiowski: I am not sure what you mean, it does launch the <launch_custom_benchmark.py> benchmark remotely with appropriate backends.

examples/launch_custom_benchmark_remotely.py
jgolebiowski: I see your point with using the specific launch_remote_X options, changed to use the appropriate function to be correct. However, I believe the code works (as it launches the appropriate benchmarks).

The code in  launch_remote_local and launch_remote_remote are very similar with less than 20% code line difference and most of them related to specifying the AWS region. The differences between the two functions do not have much impact when using the custom entrypoint from `launch_custom_benchmark.py` so I wouldn't say it's an accident exactly. I think this is actually a good argument to combine the two functions together to avoid duplication. I will probably try to merge them during the benchmarking entrypoint changes.

On a second point: hpo_main_X are used in `launch_custom_benchmark.py` which is the entrypoint for this benchmark.

examples/launch_custom_benchmark.py
jgolebiowski: Right, this argument is added later on but I override it here with an updated docstring so users of this example are not confused.

examples/launch_custom_benchmark_remotely.py
mseeger: Then, let us do your first change first. I am not OK with examples that go against how I think this stuff should be used. I also see no reason of doing it the way you try to do here.

We can always unify common code, I like that. I did that to quite some extent already. However, running experiments with SM backend is very different from local backend, and blurring the distinction is not so useful, because it is really just there.

jgolebiowski:
Closed to align of the requirements for new examples

PR #548, opened 2023-02-13
Some changes in benchmarking

mseeger:
*Issue #, if available:*

*Description of changes:*
- Increase max_wallclock_time for NB301 benchmark
- parse_known_args -> parse_args (this flags misspelled args)
- scale_max_wallclock_time argument
- main_hpo_sagemaker: warm_pool on by default
- Fix in launch_local, launch_sagemaker
- Output link to docs in commandparser

### DIFF
--- a/benchmarking/commons/benchmark_definitions/yahpo.py
+++ b/benchmarking/commons/benchmark_definitions/yahpo.py
@@ -39,7 +39,7 @@ from benchmarking.commons.benchmark_definitions.lcbench import (
 
 def yahpo_nb301_benchmark(dataset_name):
     return SurrogateBenchmarkDefinition(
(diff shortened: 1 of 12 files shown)


wesk (commented):
Can you elaborate on when someone should use `--scale_max_wallclock_time`? What problem does it solve?

mseeger:
Our benchmarks come with defaults for both `n_workers` and `max_wallclock_time`. Often, `n_workers = 4`, because with `g4dn.12xlarge`, you can nicely do 4 jobs in parallel.

If you want to compare different values of `n_workers` on a range of different benchmarks, you'd want to use larger `max_wallclock_time` for smaller `n_workers`.

wesk (approved):

PR #549, opened 2023-02-13
run launch_pasha_nasbench201.py example

wesk:
*Issue #, if available:*

*Description of changes:*


mseeger (approved):
I am OK, as long as this is only part of end-2-end, run manually before release. It uses BB repo files, so we cannot run it with every push.

wesk:
Agreed, yeah this PR adds it to the end-to-end tests. Just pushed one more commit re-commenting the lines that I was using for manual testing.

mseeger (approved):

PR #553, opened 2023-02-15
Updates to benchmarking and docs

mseeger:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/benchmarking/commons/baselines.py
+++ b/benchmarking/commons/baselines.py
@@ -13,7 +13,7 @@
 from dataclasses import dataclass
 from typing import Dict, Optional, Callable, Any, List
 
-from syne_tune.config_space import ordinal, Categorical
+from syne_tune.config_space import ordinal, Categorical, Domain, Ordinal
 from syne_tune.blackbox_repository.conversion_scripts.scripts.fcnet_import import (
     CONFIGURATION_SPACE,
 )
@@ -117,6 +117,14 @@ def _is_fcnet(config_space: Dict[str, Any]) -> bool:
     return fcnet_keys.issubset(set(config_space.keys()))
 
 
+def _to_be_converted(domain: Domain) -> bool:
+    return (
+        isinstance(domain, Categorical)
+        and not isinstance(domain, Ordinal)
+        and domain.value_type != str
+    )
+
+
 def convert_categorical_to_ordinal_numeric(
     config_space: Dict[str, Any],
     kind: Optional[str],
@@ -140,7 +148,7 @@ def convert_categorical_to_ordinal_numeric(
         return {
             name: (
                 ordinal(domain.categories, kind=kind)
-                if isinstance(domain, Categorical)
+                if _to_be_converted(domain)
                 else domain
             )
             for name, domain in config_space.items()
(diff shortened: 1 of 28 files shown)


syne_tune/optimizer/schedulers/hyperband.py
```diff
@@ -143,7 +143,7 @@ def restart(self, time_stamp: float):
 
 
 class HyperbandScheduler(FIFOScheduler, MultiFidelitySchedulerMixin):
-    """Implements different variants of asynchronous Hyperband
+    r"""Implements different variants of asynchronous Hyperband
```
wesk, line 146: Is the `r` intentional? There are few more in the PR as well

mseeger: Yes, this is for comments with contain `\`, otherwise we are getting warning outputs in the tests.

benchmarking/commons/baselines.py
```diff
@@ -117,6 +117,14 @@ def _is_fcnet(config_space: Dict[str, Any]) -> bool:
     return fcnet_keys.issubset(set(config_space.keys()))
 
 
+def _to_be_converted(domain: Domain) -> bool:
```
mseeger, line 120: All of this is to replace a `choice` definition by an `ordinal` for the fcnet configuration space. The original benchmark has a 6-valued `choice`, but this is not good for BO.

benchmarking/commons/hpo_main_simulator.py
```diff
@@ -161,8 +161,8 @@ def parse_args(
                 name="fcnet_ordinal",
                 type=str,
                 choices=("none", "equal", "nn", "nn-log"),
-                default="none",
-                help="Ordinal encoding for fcnet categorical HPs",
+                default="nn-log",
```
mseeger, line 164: This is a change of defaults. Before, the fcnet `choice` HP would remain. Now, it is by default replaced by `ordinal`.

benchmarking/commons/launch_remote_simulator.py
```diff
@@ -68,7 +68,12 @@ def get_hyperparameters(
         hyperparameters["start_seed"] = args.start_seed
     if args.benchmark is not None:
         hyperparameters["benchmark"] = args.benchmark
-    for k in ("n_workers", "max_wallclock_time", "max_size_data_for_model"):
+    for k in (
+        "n_workers",
+        "max_wallclock_time",
+        "max_size_data_for_model",
+        "fcnet_ordinal",
```
mseeger, line 75: This was overlooked

syne_tune/optimizer/schedulers/searchers/bayesopt/gpautograd/gp_model.py
```diff
@@ -290,10 +290,10 @@ def _recompute_states(self, data: Dict[str, Any]):
         self.likelihood.data_precomputations(data)
         self._states = [self.likelihood.get_posterior_state(data)]
 
-    def get_params(self):
+    def get_params(self) -> Dict[str, Any]:
```
mseeger, line 293: Just some more type annotation

benchmarking/commons/hpo_main_simulator.py
wesk: Would this require any of our customers to change their code? 

This doesn't really matter today since we are still pre-v1.0.0, but thinking ahead to when we do mark our API as stable and want to avoid breaking changes.

mseeger: Except for myself, I don't think anybody is using this right now. It is only now properly documented.

wesk (approved):

PR #555, opened 2023-02-16
Final touches for next release

mseeger:
*Issue #, if available:*

*Description of changes:*


benchmarking/commons/hpo_main_sagemaker.py
```diff
@@ -83,11 +83,11 @@ def parse_args(
             dict(
                 name="warm_pool",
                 type=int,
-                default=1,
+                default=0,
```
mseeger, line 86: Reverting an earlier change. It is probably better to have warm pools switched off by default, because they require a bit of additional effort to get going.

syne_tune/optimizer/schedulers/searchers/__init__.py
```diff
@@ -34,6 +34,9 @@
 __all__ = [
     "BaseSearcher",
     "impute_points_to_evaluate",
+    "SearcherWithRandomSeed",
```
mseeger, line 37: Needed for docs to work properly

wesk (approved):

PR #561, opened 2023-02-20
fix: Refactor examples using train_height in order to avoid bugs like that in launch_height_baselines.py, and fix edge case in GridSearcher

mseeger:
*Issue #, if available:*

*Description of changes:*

- Fix bug in `launch_height_baselines.py`. Refactored all examples to avoid such bugs
- Small change in grid searcher, correct treatment of extreme case

### DIFF
--- a/examples/launch_height_baselines.py
+++ b/examples/launch_height_baselines.py
@@ -21,6 +21,12 @@ from syne_tune.optimizer.baselines import (
 
 from syne_tune import Tuner, StoppingCriterion
 from syne_tune.config_space import randint
+from examples.training_scripts.height_example.train_height import (
+    RESOURCE_ATTR,
+    METRIC_ATTR,
+    METRIC_MODE,
+    MAX_RESOURCE_ATTR,
+)
 from syne_tune.try_import import try_import_gpsearchers_message
 
 
@@ -30,12 +36,9 @@ if __name__ == "__main__":
     random_seed = 31415927
     max_epochs = 100
     n_workers = 4
-    mode = "min"
-    metric = "mean_loss"
-    max_resource_attr = "epochs"
 
     config_space = {
-        max_resource_attr: max_epochs,
+        MAX_RESOURCE_ATTR: max_epochs,
         "width": randint(0, 20),
         "height": randint(-100, 100),
     }
@@ -46,15 +49,15 @@ if __name__ == "__main__":
         / "train_height.py"
     )
 
+    scheduler_kwargs = {
+        "config_space": config_space,
+        "metric": METRIC_ATTR,
+        "mode": METRIC_MODE,
+        "max_resource_attr": MAX_RESOURCE_ATTR,
+    }
     schedulers = [
-        RandomSearch(config_space, metric=metric, mode=mode),
-        ASHA(
-            config_space,
-            metric=metric,
-            max_resource_attr=max_resource_attr,
-            resource_attr="epoch",
-            mode=mode,
-        ),
+        RandomSearch(**scheduler_kwargs),
+        ASHA(**scheduler_kwargs, resource_attr=RESOURCE_ATTR),
     ]
     try:
         from syne_tune.optimizer.baselines import BayesianOptimization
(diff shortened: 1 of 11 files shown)


mseeger:
The bug in `launch_height_baselines.py` was not caught by running the example. I don't exactly know why, it must have terminated the example before even evaluating the functions.

wesk:
Thanks @mseeger . I will look at why running the example failed to catch this.

syne_tune/optimizer/schedulers/searchers/random_grid_searcher.py
```diff
@@ -329,22 +329,19 @@ def _next_candidate_on_grid(self) -> Optional[dict]:
...
+            candidate = dict(
+                zip(self.hp_keys, self.hp_values_combinations[self._next_index])
+            )
+            self._next_index += 1
+            is_not_done = self._all_initial_configs.contains(candidate)
+        if self._allow_duplicates and self._next_index == num_combinations:
+            # Another round through the grid
+            self._next_index = 0
+            self._all_initial_configs = ExclusionList.empty_list(self._hp_ranges)
+        return candidate
```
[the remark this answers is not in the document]
mseeger: good point, I'll add one.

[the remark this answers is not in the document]
mseeger: Thanks! Adding this test made me fix the fix!

syne_tune/optimizer/schedulers/searchers/random_grid_searcher.py
```diff
@@ -329,22 +329,21 @@ def _next_candidate_on_grid(self) -> Optional[dict]:
...
-            if self._allow_duplicates and self._next_index == num_combinations:
-                # Another round through the grid
-                self._next_index = 0
-                self._all_initial_configs = ExclusionList.empty_list(self._hp_ranges)
-            return candidate
-        else:
-            # No more candidates
-            return None
+        candidate = None
+        is_not_done = True
```
wesk, line 333: Two suggestions:
1. `done` is unclear, what does it mean here? Recommend making it more specific, such as `is_search_space_exhausted` (or whatever it's actually checking for)
2. Recommend avoiding the negation in the variable name, as I generally find positive boolean variables more readable, and then negate it using the Python syntax of `not`.

syne_tune/optimizer/schedulers/searchers/random_grid_searcher.py
```diff
@@ -329,22 +329,21 @@ def _next_candidate_on_grid(self) -> Optional[dict]:
...
+        while is_not_done and self._next_index < num_combinations:
+            candidate = dict(
+                zip(self.hp_keys, self.hp_values_combinations[self._next_index])
+            )
+            self._next_index += 1
+            is_not_done = self._all_initial_configs.contains(candidate)
+        if self._allow_duplicates and self._next_index == num_combinations:
+            # Another round through the grid
+            self._next_index = 0
+            self._all_initial_configs = ExclusionList.empty_list(self._hp_ranges)
```
wesk, line 343: I know this is pre-existing logic, but is it correct that the `_all_initial_configs` variable is being changed? Seems counterintuitive to me that this would be modified. Wouldn't it make more sense to leave `_all_initial_configs` as-is, and then over the course of the evaluation, modify some other variable such as `_remaining_configurations`?

syne_tune/optimizer/schedulers/searchers/random_grid_searcher.py
mseeger: Actually, I can get rid of this flag altogether

syne_tune/optimizer/schedulers/searchers/random_grid_searcher.py
mseeger: This is a bit subtle. In general, we enter ``points_to_evaluate`` into ``_all_initial_configs`` as they get suggested. Once this is done, we don't want to suggest these configs again in the first round over the grid.

If we allow for duplicates, there can be a second round over the grid, and there we can suggest the initial configs again (in the order they appear in the grid). This is why I need to reset ``_all_initial_configs`` here.

wesk:
Thanks for the fixes! I would just recommend updating the PR title to follow the new [ConventionalCommits](https://www.conventionalcommits.org/en/v1.0.0/) convention before we merge

wesk (approved):

examples/launch_height_baselines.py
```diff
@@ -88,7 +81,7 @@
         trial_backend = LocalBackend(entry_point=str(entry_point))
 
         stop_criterion = StoppingCriterion(
-            max_wallclock_time=5, min_metric_value={metric: -6.0}
+            max_wallclock_time=20, min_metric_value={METRIC_ATTR: -6.0}
```
wesk, line 84: For posterity: increasing `max_wallclock_time` from `5` to `20` means there is more time for any errors to be thrown. `5` seconds was too short and the example was "succeeding" when in fact, with a slightly longer runtime it would fail, before the fix in this PR (#561)

PR #562, opened 2023-02-21
chore(ci): add code coverage service

wesk:
Track code coverage using the codecov service

----

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,0 +1,3 @@
+ignore:
+  - "docs/*"
+  - "examples/*"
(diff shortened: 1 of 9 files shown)


wesk:
I'd recommend waiting and merging after the release

wesk:
@mseeger / @geoalgo FYI, I noticed that `pytest-cov` makes the existing unit tests run slightly more slowly, such that some now take longer than 1 second and fail. Rather than changing the 1-second limit here: https://github.com/awslabs/syne-tune/blob/main/pytest.ini#L2 this PR adds exceptions for 6 more tests that were crossing the 1-second mark. Let me know if you would recommend an alternative approach.

aaronkl (approved):

PR #569, opened 2023-02-21
refactor: Configure logging tag in Reporter

mseeger:
----

### DIFF
--- a/docs/source/tutorials/basics/basics_backend.rst
+++ b/docs/source/tutorials/basics/basics_backend.rst
@@ -63,7 +63,9 @@ with a SageMaker estimator, which executes the training script. In our example,
 we use the ``PyTorch`` SageMaker framework as a pre-built container for the
 dependencies our training scripts requires. However, any other type of
 `SageMaker estimator <https://sagemaker.readthedocs.io/en/stable/api/training/estimators.html>`_
-can be used here just as well.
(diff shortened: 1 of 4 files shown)


docs/source/tutorials/basics/basics_backend.rst
```diff
@@ -63,7 +63,9 @@ with a SageMaker estimator, which executes the training script. In our example,
 we use the ``PyTorch`` SageMaker framework as a pre-built container for the
 dependencies our training scripts requires. However, any other type of
 `SageMaker estimator <https://sagemaker.readthedocs.io/en/stable/api/training/estimators.html>`_
-can be used here just as well.
+can be used here just as well. Finally, if you include any of the metrics reported
+by your training script in ``metrics_names``, their values are visualized in the
+dashboard for the SageMaker training job.
```
wesk, line 68: How does the SageMaker side know to graph these logs?

mseeger: What do you mean? They are passed as `metric_names` to the estimator, and they are output by the `Reporter`. Then, SageMaker greps the logs.

We also grep the logs on our end (the SM backend does), so to receive new values. David did it this way to stay compatible with SageMaker.

wesk: Got it, thanks!

wesk (approved):

PR #574, opened 2023-02-23
feat: multi-objective REA

aaronkl:
Adds a multi-objective version of regularized evolution.

----

### DIFF
--- a/syne_tune/optimizer/schedulers/searchers/multi_objective_regularized_evolution.py
+++ b/syne_tune/optimizer/schedulers/searchers/multi_objective_regularized_evolution.py
@@ -1,0 +1,103 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+import numpy as np
+
+from collections import deque
+from typing import Optional, List, Union
+
+from syne_tune.optimizer.schedulers.searchers.regularized_evolution import (
+    PopulationElement,
+    RegularizedEvolution,
+)
+
+from syne_tune.optimizer.schedulers.multiobjective.multiobjective_priority import (
+    MOPriority,
+    NonDominatedPriority,
+)
+
+
+class MultiObjectiveRegularizedEvolution(RegularizedEvolution):
+    """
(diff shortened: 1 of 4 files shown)


syne_tune/optimizer/schedulers/searchers/multi_objective_regularized_evolution.py
```diff
@@ -0,0 +1,103 @@
...
+    RegularizedEvolution,
+)
+
+from syne_tune.optimizer.schedulers.multiobjective.multiobjective_priority import (
+    MOPriority,
+    NonDominatedPriority,
+)
+
+
+class MultiObjectiveRegularizedEvolution(RegularizedEvolution):
```
geoalgo, line 29: It seems to me that we could also just allow to use a different sort function in regularized evolution, I mean replace score by multiobjective priorities and compute priorities whenever a decision is made. Let me know what your thoughts on this, the advantage would be less code but if you think it is too convoluted or difficult to do we can also do this approach.

tst/schedulers/test_schedulers_api.py
```diff
@@ -361,11 +369,14 @@ def make_transfer_learning_evaluations(num_evals: int = 10):
...
     trial_ids = range(4)
 
-    if isinstance(scheduler, MOASHA):
+    if isinstance(scheduler, MOASHA) or isinstance(scheduler, MOREA):
         assert scheduler.metric_names() == [metric1, metric2]
     else:
         assert scheduler.metric_names() == [metric1]
-    assert scheduler.metric_mode() == mode
+
+    print(scheduler, mode)
```
geoalgo, line 377: Can you remove this line?
```suggestion
```

syne_tune/optimizer/schedulers/searchers/multi_objective_regularized_evolution.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    Adapts regularized evolution algorithm by Real et al. to the multi-objective setting. Elements in the
+    populations are scored via non-dominated sorting. Parents are sampled from the population based on
```
geoalgo, line 32:
```suggestion
    populations are scored via a multi-objective priority that is set to non-dominated sort by default. Parents are sampled from the population based on
```

geoalgo (commented):
LGTM, I just have a question whether we can provide an option to just use a custom score in REA.

syne_tune/optimizer/schedulers/searchers/multi_objective_regularized_evolution.py
geoalgo: To be clear, this is non blocking and I am fine merging as is. In my opinion, having a custom sort would be slightly better: there would be less code and I think it is quite easy to explain to a user that the algorithm can be customized with different criterion, some of which are multi-objective (you could also have a subclass which just instantiate a specific non-dominated sort as the priority function).

We can also do it in two steps, i.e. merge now and do the custom sort, later as you prefer.

aaronkl: Ok then let's do as you suggested

geoalgo (approved):

Issue #554, opened 2023-02-15
Allow for list-valued hyperparameters


mseeger:
**Is your feature request related to a problem? Please describe.**

The number of features could vary in the model and based on that total hyperparameter to tune would also change. This change in feature leads to change to in config space and input script. For example, when there are two features in the model, the config space would look like:

Feature_list=[‘X1’, ‘X2’]
```
config_space={
   degree_X1: randint(1,3),
   lambda_X1: randint(1,1000),
   knots_X1: randint(1, 10),
   degree_X2: randint(1,3),
   lambda_X2: randint(1,1000),
    knots_X2: randint(1, 10),
}
```
but when feature list changes to three variables for an instance, the config space will need nine hyperparameters.

**Describe the solution you'd like**

t would be great if we could use a list of hyperparameter in config space. This would save us from creating wrapper to 1) generate new config and 2) a new input file to accept the updated hyperparameters. Essentially, having something like the below would be helpful-
 
Config_space={ degree:[randint(1,3), randint(1,3)],                        #a list of hyperparameters, each value in list corresponds to degree of each variable in feature list
                                Lambda:[ randint(1,1000), randint(1,1000)],       #a list of hyperparameters, each value in list corresponds to lambda value of each variable in feature list
                                Knots:[randint(1,10), randint(1,10)],                     #a list of hyperparameters, each value in list corresponds to knot value of each variable in feature list
                                Feature_list:[‘X1’, ‘X2’]}                                           # a constant list passed to input script

**Describe alternatives you've considered**

We can replicate each feature by hand, as in the example above, but this requires code changes.

mseeger:
The idea would be that we allow for list-valued hyperparameters in a config space. Each component can be a `Domain` or a fixed value.

Internally, this would be flattened when it comes to searchers and encodings.

This can certainly be done, but we should warn folks that high-D HPO is often failing if no special structure is present and assumed by the surrogate model.

mseeger:
One technical blocker is that command line arguments for SageMaker training jobs are limited in length. What is possible is to pass parameters to a script as JSON file `config.json`. We could maybe provide some tooling to load this file inside a training script.

This is not a simple change.

mseeger:
https://github.com/awslabs/syne-tune/blob/main/benchmarking/nursery/tuning_with_hyperparameter_file

mseeger:
Example of how to read config from JSON file. This works for the local backend only at the moment. I can try and support this for the SageMaker backend as well.

[...]

PR #577, opened 2023-02-23
feature: Support passing config as JSON to trial evaluation script

mseeger:
Supports passing config as JSON file to training script, in local and SageMaker backend.

This needed some small fixes of `SageMakerBackend` as well.

The new example serves as test, because it fails if the arguments are not passed properly.

Issue: #554 

----


examples/training_scripts/height_example/train_height_config_json.py
```diff
@@ -0,0 +1,98 @@
...
+
+from syne_tune import Reporter
+from syne_tune.config_space import randint
+from benchmarking.utils import add_config_json_to_argparse, load_config_json
+
+
+report = Reporter()
+
+
+RESOURCE_ATTR = "epoch"
```
mseeger, line 26: Cannot import them from `train_height.py`, because `examples` is not a dependence of the SM estimator (I tried).

benchmarking/utils/config_as_json.py
```diff
@@ -0,0 +1,39 @@
...
+def add_config_json_to_argparse(parser: argparse.ArgumentParser):
+    """
+    To be called for the argument parser in the endpoint script.
+
+    :param parser: Parser to add extra arguments to
+    """
+    parser.add_argument(f"--{ST_CONFIG_JSON_FNAME_ARG}", type=str)
+
+
+def load_config_json(args) -> Dict[str, Any]:
```
jgolebiowski, line 29: I think we can make it more general and have it accept `Dict[str, Any]`. This way it can be called with `method(vars(args))` so not much chnages but it also can be used to expand any config in Dict format with the data from json

syne_tune/backend/sagemaker_backend/sagemaker_backend.py
```diff
@@ -194,9 +203,62 @@ def _checkpoint_s3_uri_for_trial(self, trial_id: int) -> str:
...
         return f"{res_path}/{str(trial_id)}/checkpoints/"
 
+    def _config_json_filename(self, trial_id: int, with_path: bool) -> str:
+        fname = CONFIG_JSON_FILENAME + f"_{trial_id}.json"
+        if with_path and self.source_dir is not None:
+            return str(Path(self.source_dir) / fname)
+        else:
+            return fname
+
+    def _hyperparameters_from_config(
```
jgolebiowski, line 213: We should split this function into two as it accomplishes two things: 1/ translate config into hyperparams and 2/ store config into json if `self.pass_args_as_json`. We should have two functions a/ `_hyperparameters_from_config` which just does the translation and b/ `_store_config_in_json` which will take care of the writing. We can call them from `_schedule` as 

```
hyperparameters = self._hyperparameters_from_config(trial_id, config)
if self.pass_args_as_json:
    self._store_config_in_json(trial_id, config)
```

jgolebiowski (commented):
Very nice change, it seems like we can add this feature without much changes as we are already storing config files. I have requested a small change but a larger question is: I see that a lot of new code is not covered by tests, can you describe how you validated the changes, was it only with examples? Following that, should we consider adding tests for this feature?

mseeger:
Hi Jacek, the new example serves as test, because it checks internally whether the list and dict components are properly transmitted. It is run for local backend as part of our tests. It is also run as end-2-end test with SageMaker backend, and I confirmed this works.

In fact, I had to change/fix a few things in SageMaker backend to make this work

benchmarking/utils/config_as_json.py
mseeger: OK

.github/workflows/end-to-end-tests.yml
```diff
@@ -118,3 +118,11 @@ jobs:
       script-path: examples/launch_pasha_nasbench201.py
       requires-aws-credentials: true
     secrets: inherit
+
+  launch_height_config_json_sagemaker:
```
mseeger, line 122: Serves as test for SageMaker backend. This is run as end-2-end test

jgolebiowski:
Sure makes sense, we have integration tests but I was wondering about unit tests for individual components. I suppose if integ tests cover all the functionalities we can skip the unit tests for now.

syne_tune/backend/sagemaker_backend/sagemaker_backend.py
mseeger: OK, I split it into two. However, both 1/ and 2/ are only non-trivial (and connected) if `self.pass_args_as_json`.

Basically, the point is distributing the config between CL arguments (i.e., hyperparameters returned) and what is written as JSON.

It is simpler this way, i.e. passing default args with names "ST_..." *always* on the command line, because there are other helper functions for parsing these args. These args always have simple types (int, float, str), so it does not matter

jgolebiowski (approved):
LGTM, nice new feature.

jgolebiowski (approved):
No chnage from previous review.

PR #579, opened 2023-02-25
fix: Fix of BOHB and SyncBOHB

mseeger:
Conditions in `KernelDensityEstimator` for whether KDE models can be fit are typically violated in `MultiFidelityKernelDensityEstimator`, so this did the same as ASHA. Here, I am changing the choice of `r_acq` to depend on the "can fit" decision

----

### DIFF
--- a/syne_tune/optimizer/schedulers/searchers/kde/kde_searcher.py
+++ b/syne_tune/optimizer/schedulers/searchers/kde/kde_searcher.py
@@ -10,7 +10,7 @@
 # on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
-from typing import Optional, List, Dict, Any
+from typing import Optional, List, Dict, Any, Tuple
 import logging
 import numpy as np
 import statsmodels.api as sm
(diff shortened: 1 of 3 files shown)


mseeger:
I reran my experiments on our 13 surrogate benchmarks. BOHB is now quite different from ASHA, and often better.

syne_tune/optimizer/schedulers/searchers/kde/kde_searcher.py
```diff
@@ -353,30 +354,27 @@ def acquisition_function(x):
 
         return suggestion
 
-    def _train_kde(self, train_data, train_targets):
-        if train_data.shape[0] < self.num_min_data_points:
+    def _good_data_size(self, data_shape: Tuple[int, int]) -> Optional[int]:
+        num_data, num_features = data_shape
+        n_good = max(self.num_min_data_points, (self.top_n_percent * num_data) // 100)
+        if min(n_good, num_data - n_good) <= num_features:
```
aaronkl, line 360: I am not sure I follow here. I assume num_data - n_godd are the number of data point for the KDE that models the poorly performing configuration? Why do both have to be larger or equal than num_features? Also you might want to change the name of the function to reflect that you check the correct sizes for both models?

syne_tune/optimizer/schedulers/searchers/kde/kde_searcher.py
```diff
@@ -353,30 +354,27 @@ def acquisition_function(x):
...
         idx = np.argsort(train_targets)
-
         train_data_good = train_data[idx[:n_good]]
-        train_data_bad = train_data[idx[n_good : n_good + n_bad]]
-
-        if train_data_good.shape[0] <= train_data_good.shape[1]:
-            return None
-        if train_data_bad.shape[0] <= train_data_bad.shape[1]:
-            return None
+        train_data_bad = train_data[idx[n_good:]]
```
aaronkl, line 375: Why do we not check anymore if n_bad is greater than self.num_min_data_points?

syne_tune/optimizer/schedulers/searchers/kde/kde_searcher.py
mseeger: The current code is doing these checks, I did not change that

syne_tune/optimizer/schedulers/searchers/kde/kde_searcher.py
mseeger: This is done in the `_good_data_size` method

wesk:
I just triggered a re-run of the tests, let's see if this is transient or not

mseeger:
Thanks, it was a hickup

aaronkl (dismissed):

aaronkl (approved):

PR #580, opened 2023-02-26
refactor: Use longer descriptive names for result directories

mseeger:
In the current code, benchmake name and seed are not contained in the directory name for the results, but just in metadata.json. A more descriptive filename makes it easier to pick out or remove specific results.

----

### DIFF
--- a/benchmarking/commons/hpo_main_common.py
+++ b/benchmarking/commons/hpo_main_common.py
@@ -115,6 +115,12 @@ def parse_args(
             "where A = n_workers and B = benchmark.n_workers"
         ),
     )
+    parser.add_argument(
+        "--use_long_tuner_name_prefix",
+        type=int,
+        default=1,
+        help="Use descriptive tuner name prefix for storing results?",
+    )
     if extra_args is not None:
         extra_args = copy.deepcopy(extra_args)
         for kwargs in extra_args:
@@ -126,6 +132,7 @@ def parse_args(
     args = parser.parse_args()
     args.save_tuner = bool(args.save_tuner)
     args.scale_max_wallclock_time = bool(args.scale_max_wallclock_time)
+    args.use_long_tuner_name_prefix = bool(args.use_long_tuner_name_prefix)
     seeds = list(range(args.start_seed, args.num_seeds))
     method_names = [args.method] if args.method is not None else list(methods.keys())
     return args, method_names, seeds
--- a/benchmarking/commons/hpo_main_local.py
+++ b/benchmarking/commons/hpo_main_local.py
@@ -158,11 +158,14 @@ def create_objects_for_tuner(
         benchmark=benchmark,
         extra_args=None if extra_args is None else extra_metadata(args, extra_args),
     )
+    tuner_name = args.experiment_tag
+    if args.use_long_tuner_name_prefix:
+        tuner_name += f"-{args.benchmark}-{seed}"
     return dict(
         scheduler=scheduler,
         stop_criterion=stop_criterion,
         n_workers=benchmark.n_workers,
-        tuner_name=args.experiment_tag,
+        tuner_name=tuner_name,
         metadata=metadata,
         save_tuner=args.save_tuner,
     )
--- a/docs/source/tutorials/benchmarking/bm_simulator.rst
+++ b/docs/source/tutorials/benchmarking/bm_simulator.rst
@@ -162,6 +162,11 @@ This call runs a number of experiments sequentially on the local machine:
   where ``A = n_workers``, ``B = benchmark.n_workers``. This means we run for
   longer if ``n_workers < benchmark.n_workers``, but keep
   ``benchmark.max_wallclock_time`` the same otherwise.
+* ``use_long_tuner_name_prefix``: If 1, results for an experiment are written
+  to a directory whose prefix is
+  :code:`f"{experiment_tag}-{benchmark_name}-{seed}"`, followed by a postfix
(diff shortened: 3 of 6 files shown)


benchmarking/commons/hpo_main_common.py
```diff
@@ -115,6 +115,12 @@ def parse_args(
             "where A = n_workers and B = benchmark.n_workers"
         ),
     )
+    parser.add_argument(
+        "--use_long_tuner_name_prefix",
+        type=int,
+        default=1,
+        help="Use descriptive tuner name prefix for storing results?",
+    )
```
wesk, line 123: This would be a breaking change. Do we really want to set this as the new default behavior? 

(Maybe we do, but we should weigh the cost for existing customers versus the ease-of-use benefits for future customers. Also, we're not yet at version 1.0.0, but once we say we're stable, we'll want to strive for backward compatibility.)

benchmarking/commons/hpo_main_local.py
```diff
@@ -158,11 +158,14 @@ def create_objects_for_tuner(
         benchmark=benchmark,
         extra_args=None if extra_args is None else extra_metadata(args, extra_args),
     )
+    tuner_name = args.experiment_tag
+    if args.use_long_tuner_name_prefix:
+        tuner_name += f"-{args.benchmark}-{seed}"
```
wesk, line 163: Rather than passing this additional metadata inside `tuner_name`, why not create an explicit separate variable, something like `tuner_output_path`?

benchmarking/commons/hpo_main_common.py
```diff
@@ -126,6 +132,7 @@ def parse_args(
     args = parser.parse_args()
     args.save_tuner = bool(args.save_tuner)
     args.scale_max_wallclock_time = bool(args.scale_max_wallclock_time)
+    args.use_long_tuner_name_prefix = bool(args.use_long_tuner_name_prefix)
```
wesk, line 135: The key `use_long_tuner_name_prefix` is somewhat unclear in my opinion.

What if we instead create a flag like `custom-tuner-name-prefix`, which would be false by default. We would allow customers to pass their own prefixes and override our default value if they choose. 

As for what the default value should be, it could either: 1) be the "long" prefix that you propose in this CR (this would be a breaking change) or 2) be the status-quo prefix.

docs/source/tutorials/benchmarking/bm_simulator.rst
```diff
@@ -162,6 +162,11 @@ This call runs a number of experiments sequentially on the local machine:
   where ``A = n_workers``, ``B = benchmark.n_workers``. This means we run for
   longer if ``n_workers < benchmark.n_workers``, but keep
   ``benchmark.max_wallclock_time`` the same otherwise.
+* ``use_long_tuner_name_prefix``: If 1, results for an experiment are written
+  to a directory whose prefix is
+  :code:`f"{experiment_tag}-{benchmark_name}-{seed}"`, followed by a postfix
```
wesk, line 167: Are seeds limited to three characters? 

Also, how does one set the experiment tag?

benchmarking/commons/hpo_main_common.py
mseeger: I do not think this is really a breaking change. Results are stored in directories with prefix `tuner_name`. The only thing this PR is doing, is to use a longer value for `tuner_name`, when this is used with `benchmarking`.

I can set the default to 0, but the reason for this change is  pragmatic. Say you run experiments with several benchmarks, and suddenly you figure they need to be redone for one of the benchmarks. You need to remove the
result files in S3, but you cannot identify them from the directory names.

With this change, since benchmark name is part of the `tuner_name`, you can easily remove them using the S3 console. There is no drawback, except longer directory names.

benchmarking/commons/hpo_main_local.py
mseeger: `tuner_name` is an argument of `Tuner`, it is the prefix of the result files written by the tuner.

benchmarking/commons/hpo_main_common.py
mseeger: But the prefix is a combination of other information, it is not a fixed string you can pass here.

docs/source/tutorials/benchmarking/bm_simulator.rst
mseeger: No, why? Seeds are numbers starting from 0.

The `Tuner` writes results (and checkpoints) to a directory whose name has prefix `tuner_path`, and it appends a datetime plus a 3-digit hash, so things are always unique.

This PR is about choosing what `tuner_name` should be. Currently, it is just `experiment_tag`, which is typically a short string you choose. I run comparative experiments where 1000 or even 10000 experiments share the same `experiment_tag`, they differ by method, seed, benchmark, etc.

To tell these apart, you need to read the `metadata.json` files for each of them and filter according to values for seed and benchmark in there.

So, a longer `tuner_name` is redundant, because the info is in the metadata, but it simplifies basic things like re-run parts of a large experiment.

benchmarking/commons/hpo_main_common.py
wesk: Spoke offline, makes sense, scope is also just limited to benchmarking

docs/source/tutorials/benchmarking/bm_simulator.rst
wesk: Thanks for the clarification, makes sense!

wesk (approved):

PR #583, opened 2023-03-02
refactor: Cleanup of benchmark examples

mseeger:
Refactoring of 2 of our benchmark scripts, which makes it easier to understand them. I did this so I can show them in tutorials.

----

### DIFF
--- a/benchmarking/training_scripts/resnet_cifar10/resnet_cifar10.py
+++ b/benchmarking/training_scripts/resnet_cifar10/resnet_cifar10.py
@@ -35,6 +35,8 @@ METRIC_NAME = "objective"
 
 RESOURCE_ATTR = "epoch"
 
+MAX_RESOURCE_ATTR = "epochs"
+
 ELAPSED_TIME_ATTR = "elapsed_time"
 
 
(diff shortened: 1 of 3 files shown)


benchmarking/training_scripts/resnet_cifar10/resnet_cifar10.py
```diff
@@ -178,7 +165,13 @@ def objective(config):
         sampler=valid_sampler,
         pin_memory=True,
     )
+    return train_loader, valid_loader
 
+
+def _create_training_objects(config):
+    num_gpus = config.get("num_gpus")
+    if num_gpus is None:
+        num_gpus = 1
```
wesk, line 174: What are the possible scenarios where `num_gpus is None`?

benchmarking/training_scripts/resnet_cifar10/resnet_cifar10.py
```diff
@@ -189,33 +182,52 @@ def objective(config):
...
     )
     scheduler = torch.optim.lr_scheduler.MultiStepLR(
         optimizer, milestones=milestones, gamma=0.1
     )
+    return model, optimizer, scheduler
+
 
-    # Checkpointing
+def objective(config):
+    torch.manual_seed(np.random.randint(10000))
```
wesk, line 197: Should we set a manual seed here?

wesk (commented):
Just two clarification questions

benchmarking/training_scripts/resnet_cifar10/resnet_cifar10.py
mseeger: If the parameter is not given. We could deal with this in argparse, just saw that

benchmarking/training_scripts/resnet_cifar10/resnet_cifar10.py
mseeger: This is not my code. I think Aaron wrote it.
We control seeds of everything on our side, but not what happens in the training functions.

mseeger: We definitely should not use the same seed, because that means that initial weights and batch ordering are the same for each evaluation, that would not be good.

mseeger: In the end, we cannot assume that we can control this for some user code, so let us not try to do it.

wesk (approved):

PR #584, opened 2023-03-06
call init of class

aaronkl:
Call the init method of the current class not of the base class.

----

### DIFF
--- a/syne_tune/callbacks/tensorboard_callback.py
+++ b/syne_tune/callbacks/tensorboard_callback.py
@@ -163,7 +163,7 @@ class TensorboardCallback(TunerCallback):
         return state
 
     def __setstate__(self, state):
-        super().__init__(
+        self.__init__(
             ignore_metrics=state["ignore_metrics"],
             target_metric=state["target_metric"],
             mode="min" if state["metric_sign"] == 1 else "max",


wesk:
Would you please update the PR title with the conventionalcommits prefix, looks like `fix: ` here? Do we have adequate test coverage for this class of issue more generally?

aaronkl:
The error appears when we re-loaded a Tuner object from a tuner.dill file. It seems that we don't have a test coverage for that

wesk:
Are there any tests we can add to prevent this type of error from occurring in the future?

syne_tune/callbacks/tensorboard_callback.py
```diff
@@ -163,7 +163,7 @@ def __getstate__(self):
         return state
 
     def __setstate__(self, state):
-        super().__init__(
+        self.__init__(
```
mseeger, line 166: I am not familiar with how to code `__setstate__`. But what is the difference here? A comment would be nice, because since most of us do not understand this "Python magic", we just copy and paste.

I am very suspicious about calling the constructor for an already existing class (what else is `self`), so what I did in the scheduler and searcher hierarchy is to define method that make more sense to me.

I note there is also `syne_tune.blackbox_repository.BlackboxRepositoryBackend` which has the same code, could you change that as well?

mseeger (changes requested):

mseeger:
Background: In the past, we used to rely on serializing the `Tuner` object and all its dependent objects as a `dill` file, and load it on an instance. For example, I think `RemoteLauncher` may still do this.

The new tooling uses SageMaker estimators, so this `dill` serialization is not really needed anymore, AFAIK. But sure, somebody could write a test. Just would have to test a lots of combinations.

I've never been a fan of this reliance on serialization on our side. This is just a large burden, because many members of a class cannot just be default-serialized, because they have funny types or are very large.

syne_tune/callbacks/tensorboard_callback.py
mseeger: OK, the code in `BlackboxRepositoryBackend` is correct, it is supposed to call the super-class constructor.

mseeger:
@wesk What we could do is to end all the tests which run experiments in our examples or forthcoming integration test by saving and restoring the dill file for the complete tuner. This would give us coverage at least over all experiments we are anyway testing.

In order to catch things like what @aaronkl found here, we would need to do more, because here the error was in some callback that is not normally used.

PR #585, opened 2023-03-06
fix: Provide different state converter for DyHPO, which fails with ex…

mseeger:
…isting one

Fixes a major issue with the DyHPO implementation. The current one uses dataset filtering to keep the GP computations limited. This filtering was specialized to MOBSTER, Hyper-Tune, but does not make sense for DyHPO (see code comments). Replaced with two-step filtering.

This gets much improved results on nas201, but about similar ones on lcbench. DyHPO gets worse on nas301, but this seems an outlier, given that the filtering is obviously nonsense for DyHPO.
----


syne_tune/optimizer/schedulers/searchers/__init__.py
```diff
@@ -34,8 +34,8 @@
 __all__ = [
     "BaseSearcher",
     "impute_points_to_evaluate",
-    "SearcherWithRandomSeed",
-    "SearcherWithRandomSeedAndFilterDuplicates",
+    "StochasticSearcher",
```
mseeger, line 37: Was forgotten in Aaron's PR

syne_tune/optimizer/schedulers/searchers/bayesopt/models/subsample_state.py
```diff
@@ -0,0 +1,325 @@
...
+        ), "Call set_random_state before first usage"
+        return cap_size_tuning_job_state(
+            state=state, max_size=self.max_size, random_state=self._random_state
+        )
+
+    def set_random_state(self, random_state: RandomState):
+        self._random_state = random_state
+
+
+def _sparsify_at_most_one_per_trial_and_group(
```
mseeger, line 180: New code starts from here. Above was just moved

syne_tune/optimizer/schedulers/utils/successive_halving.py
```diff
@@ -0,0 +1,87 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mseeger, line 1: Just code moved from `hyperband.py`

wistuba (approved):

PR #592, opened 2023-03-10
fix: Further improvements to logic syncing checkpoints to S3 or not

mseeger:
As noted by Aaron, the previous solution was not transparent and maybe incomplete.

----

### DIFF
--- a/syne_tune/backend/sagemaker_backend/sagemaker_utils.py
+++ b/syne_tune/backend/sagemaker_backend/sagemaker_utils.py
@@ -28,6 +28,7 @@ from sagemaker.estimator import Framework
 
 import syne_tune
 from syne_tune.backend.trial_status import TrialResult
+from syne_tune.backend.trial_backend import TrialBackend
 from syne_tune.report import retrieve
 from syne_tune.util import experiment_path, random_string, s3_experiment_path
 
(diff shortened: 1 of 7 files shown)


syne_tune/backend/sagemaker_backend/sagemaker_utils.py
```diff
@@ -446,3 +447,25 @@ def delete_action(s3_client, bucket: str, object_key: str) -> Optional[str]:
...
+
+    .. note::
+       This has to be called *after* the :class:`~syne_tune.Tuner` object has
+       been created, because the latter sets the backend path as well.
+
+    :param trial_backend: Backend in question
+    """
+    path = Path("/opt/ml/input/data/")
+    path.mkdir(parents=True, exist_ok=True)
+    trial_backend.set_path(results_root=str(path))
```
aaronkl, line 471: Just to make sure, this does not change the path for results.csv.zip and metadata.json , right?

mseeger: No, it does not. It is a bit tricky. When `tuner` gets constructed, it both sets `tuner.tuner_path`, where tuning results are written, but also it sets `tuner.trial_backend.set_path(...)` to the same path, and this is where the backend writes info per trial, such as checkpoints and logs.

Here, only the latter is changed.

I tested this, results are written as before.

mseeger: And there is also a reason why we are not writing just to "~/", or something returned by `tempfile`. It seems there is not so much space there for SM training jobs, but there is more space at "/opt/ml/input/data/". The documentation on this stuff is non-existing, though. You should run it with a large Hugging Face model and check that things work.

mseeger: BTW: When `delete_checkpoints=True` for the backend (I think this is default), then we remove checkpoints once a trial is stopped or terminates (but NOT if it is paused). This can save some space when checkpoints are large.

On the other hand, for ASHA promotion, most trials are paused until the end, so it does not help much.

mseeger: You should check what happens for PBT: Are we removing checkpoints which are no longer needed? If not, we should re-think this, because our checkpoints will only ever get larger.

aaronkl (dismissed):

aaronkl (approved):

PR #594, opened 2023-03-10
fix: Make reference anonymous to avoid Sphinx warnings

mseeger:
In rst, links `text <...>`_ (one underscore) are supposed to have unique text. It is more useful to use anonymous links, of the form `text <...>`__ (two underscores), this does not give warnings about duplicate text.

----

### DIFF
--- a/docs/source/benchmarking/benchmark_dehb.rst
+++ b/docs/source/benchmarking/benchmark_dehb.rst
@@ -2,7 +2,7 @@ Code in benchmarking/nursery/benchmark_dehb
 ===========================================
 
 Comparison of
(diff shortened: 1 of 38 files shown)


wesk:
This won't change the way docs render, correct? It will just remove warnings that currently show up when the docs build?

wesk (approved):
I noticed that all the files that are modified are `.rst` files, except for `syne_tune/optimizer/schedulers/searchers/bayesopt/gpautograd/gluon.py`. Should we really add the underscores to lines 805 and 917 of the .py file?

Approving as-is since this is a nit

mseeger:
We want to use anynomous links everywhere, also in .py files.

PR #597, opened 2023-03-13
fix: Fixes to benchmarking with simulator backend

mseeger:
Without this fix, benchmarking does not work properly for simulator backend.

Also some fixes of docstrings, and extended default `max_wallclock_time` for `fcnet` benchmark.

----

### DIFF
--- a/benchmarking/commons/hpo_main_simulator.py
+++ b/benchmarking/commons/hpo_main_simulator.py
@@ -45,13 +45,13 @@ from syne_tune.stopping_criterion import StoppingCriterion
 from syne_tune.tuner import Tuner
 from syne_tune.util import sanitize_sagemaker_name
 
+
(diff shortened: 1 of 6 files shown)


benchmarking/commons/hpo_main_simulator.py
```diff
@@ -340,7 +344,7 @@ def start_benchmark_simulated_backend(
             )
         tuner_name = experiment_tag
         if configuration.use_long_tuner_name_prefix:
-            tuner_name += f"-{sanitize_sagemaker_name(configuration.benchmark)}-{seed}"
+            tuner_name += f"-{sanitize_sagemaker_name(benchmark_name)}-{seed}"
```
wesk, line 347: Is this line the fix? Are there any other places where we might also have done this?

wesk (approved):

PR #599, opened 2023-03-14
refactor: Make all literalinclude in docs independent of line numbers

mseeger:
Robustifies showing code snippets in the docs. Since ranges are selected not based on line numbers, but on how they start and end, this is more robust to code changes.

----

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -242,7 +242,7 @@ checkpointing enabled:
 .. literalinclude:: ../../examples/training_scripts/checkpoint_example/checkpoint_example.py
    :name: checkpoint_example_script
    :caption: examples/training_scripts/checkpoint_example/checkpoint_example.py
-   :lines: 13-
+   :start-after: # permissions and limitations under the License.
 
 When using the SageMaker backend, we use the
 `SageMaker checkpoint mechanism <https://docs.aws.amazon.com/sagemaker/latest/dg/model-checkpoints.html>`__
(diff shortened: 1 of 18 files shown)


docs/source/faq.rst
```diff
@@ -722,15 +722,17 @@ backend:
 
 .. literalinclude:: ../../examples/launch_height_config_json.py
    :caption: examples/launch_height_config_json.py
-   :lines: 72-75
+   :start-after: if not use_sagemaker_backend:
+   :end-before: else:
```
wesk, line 726: This is better than the raw line numbers, but also might be somewhat brittle. Just interesting. Maybe in the future we could add special comments or something to lines of interest for docs, and start/end on those (but that might also open up new challenges)

docs/source/tutorials/basics/basics_setup.rst
```diff
@@ -95,7 +95,8 @@ tutorial:
 
 .. literalinclude:: code/hpo_main.py
    :caption: hpo_main.py: Configuration space
-   :lines: 25-37
+   :start-at: from syne_tune.config_space import randint, uniform, loguniform
```
wesk, line 98: A risk would be if we import more things on that line

wesk: But I think it's fine given that it's docs code: https://github.com/awslabs/syne-tune/blob/main/docs/source/tutorials/basics/code/hpo_main.py, so if people change it, they will reasonably expect the docs to change, and reviewers should check for this.

wesk (approved):
Nice!

PR #600, opened 2023-03-15
Make sure that checkpoints in PBT are removed once they are no longer…

mseeger:
… needed

This should solve the issue about checkpoints not being removed. I just return STOP instead of PAUSE. See the code comment why I think this is fine.
I also implemented early checkpoint removal for the synchronous schedulers, which requires the scheduler to flag paused trials for which the checkpoint is no longer needed.

----

### DIFF
--- a/syne_tune/callbacks/remove_checkpoints_callback.py
+++ b/syne_tune/callbacks/remove_checkpoints_callback.py
@@ -1,0 +1,31 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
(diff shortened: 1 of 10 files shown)


mseeger:
I should also check what happens for synchronous multi-fidelity methods.
The potential issues with promo-based ASHA remain, because checkpoints of paused trials are never removed before the end of the experiment.

mseeger:
Converting this to draft. I'd still like to resoved the situation for synchronous MF schedulers

mseeger:
OK, this should go in, since it fixed the checkpoint situation for PBT and the synchronous schedulers. Pre-emptive removal for ASHA promotion is more complicated, I'll do it later.

PR #604, opened 2023-03-20
Keep rung levels sorted in HyperbandScheduler

### LINKED ISSUE #377: Keep rungs sorted in HyperbandScheduler
Currently, whenever a decision is made at a rung level (about whether to stop a trial, or promote it), the metric values stored at the rung are sorted again. For example:

https://github.com/awslabs/syne-tune/blob/main/syne_tune/optimizer/schedulers/hyperband_stopping.py#L29

This can be wasteful if a rung level gets populated with a lot of trials, which can happen at the lower rungs.

The goal for this issue would be to maintain the rung levels in sorted order, so the percentile values can be determined in O(1).

Note that these decisions are part of `RungSystem` subclasses, and it would be good to modify all of them, as long as this makes sense.

mseeger:
The current code recomputes the quantile for every inserting into a rung. This means that n insertions cost `O(n^2 log n)`, which can be a problem for large n. The new code keeps rung enntries sorted, using a data structure where insertions are `O(log n)`, so n insertions cost `O(n log n)`.

The PR also cleans things up. This is needed for another idea for clever checkpoint removal.

Closes #377

----

### DIFF
--- a/benchmarking/nursery/benchmark_dehb/requirements.txt
+++ b/benchmarking/nursery/benchmark_dehb/requirements.txt
@@ -1,2 +1,3 @@
 syne-tune[gpsearchers,kde,blackbox-repository,yahpo,aws]
 tqdm
+sortedcontainers  # Remove in version > 0.4.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ dill>=0.3.6
 pandas
 # only needed for python version lower than 3.8
 typing_extensions
+sortedcontainers
--- a/syne_tune/optimizer/schedulers/hyperband_pasha.py
+++ b/syne_tune/optimizer/schedulers/hyperband_pasha.py
@@ -81,19 +81,21 @@ class PASHARungSystem(PromotionRungSystem):
             self._rungs[-self.current_rung_idx],
             self._rungs[-self.current_rung_idx + 1],
         ]:
-            if rung.data != {}:
-                trial_ids = rung.data.keys()
-                values = []
-                for trial_id in trial_ids:
-                    values.append(rung.data[trial_id][0])
-                # order specifies where the value should be placed in the sorted list
-                values_order = np.array(values).argsort()
-                # calling argsort on the order will give us the ranking
-                values_ranking = values_order.argsort()
+            if len(rung) > 0:
+                # Note that entries in ``rung.data`` are already sorted
+                trial_ids, values = zip(
+                    *[(x.trial_id, x.metric_val) for x in rung.data]
+                )
+                # Note: To retain the exact logic of older code (which used
+                # ``numpy.argsort``, we assume the ranks for increasing metric
+                # values, which is the opposite ordering to ``rung.data`` if
+                # ``self._mode == "max"``
+                if self._mode == "min":
+                    values_ranking = list(range(len(trial_ids)))
+                else:
+                    values_ranking = list(range(len(trial_ids) - 1, -1, -1))
                 ranking = list(zip(trial_ids, values_ranking, values))
-
                 rankings.append(ranking)
-
         return rankings
 
     def _get_sorted_top_rungs(self, rankings):
@@ -111,11 +113,7 @@ class PASHARungSystem(PromotionRungSystem):
             lambda e: e[0] in top_rung_keys, rankings[1]
         )
         # if we try to maximize the objective, we need to reverse the ranking
-        if self._mode == "max":
-            reverse = True
-        else:
-            reverse = False
-
+        reverse = self._mode == "max"
         sorted_top_rung = sorted(rankings[0], key=lambda e: e[1], reverse=reverse)
         sorted_previous_rung = sorted(
             corresponding_previous_rung_trials, key=lambda e: e[1], reverse=reverse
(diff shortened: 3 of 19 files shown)


benchmarking/nursery/benchmark_dehb/requirements.txt
```diff
@@ -1,2 +1,3 @@
 syne-tune[gpsearchers,kde,blackbox-repository,yahpo,aws]
 tqdm
+sortedcontainers  # Remove in version > 0.4.1
```
mseeger, line 3: The reason for these temp changes is that syne-tune dependencies are installed from the latest PyPI release. Since `sortedcontainers` is a new dependence, it is not picked up from there.

requirements.txt
```diff
@@ -3,3 +3,4 @@ dill>=0.3.6
 pandas
 # only needed for python version lower than 3.8
 typing_extensions
+sortedcontainers
```
mseeger, line 6: Small (and very useful) extra dependence in core

syne_tune/optimizer/schedulers/searchers/dyhpo/dyhpo_searcher.py
```diff
@@ -92,9 +101,64 @@ def _get_config_modelbased(
         self._debug_log = None  # No more debug logging for rest of ``get_config``
         return {INTERNAL_KEY: "dummy"}
 
+    def _extended_configs_from_paused(
```
mseeger, line 104: This is a small refactoring, because the old code was hard to read.

syne_tune/optimizer/schedulers/hyperband_pasha.py
```diff
@@ -81,19 +81,21 @@ def _get_top_two_rungs_rankings(self):
...
-            if rung.data != {}:
-                trial_ids = rung.data.keys()
-                values = []
-                for trial_id in trial_ids:
-                    values.append(rung.data[trial_id][0])
-                # order specifies where the value should be placed in the sorted list
-                values_order = np.array(values).argsort()
-                # calling argsort on the order will give us the ranking
-                values_ranking = values_order.argsort()
+            if len(rung) > 0:
```
wistuba, line 84: nit: you could drop `>0`

syne_tune/optimizer/schedulers/hyperband_pasha.py
```diff
@@ -227,18 +225,17 @@ def _update_epsilon(self):
...
-        ]
-
-        if result[self._resource_attr] not in self.epoch_to_trials:
-            self.epoch_to_trials[result[self._resource_attr]] = set()
-        self.epoch_to_trials[result[self._resource_attr]].add(trial_id)
-
-        if result[self._resource_attr] > self.current_max_epoch:
-            self.current_max_epoch = result[self._resource_attr]
+            self.per_epoch_results[trial_id] = {resource: metric_val}
+        else:
```
wistuba, line 232: nit: init with empty dict and dropping else seems more readable to me

syne_tune/optimizer/schedulers/hyperband_pasha.py
```diff
@@ -81,19 +81,21 @@ def _get_top_two_rungs_rankings(self):
...
+                    *[(x.trial_id, x.metric_val) for x in rung.data]
+                )
+                # Note: To retain the exact logic of older code (which used
+                # ``numpy.argsort``, we assume the ranks for increasing metric
+                # values, which is the opposite ordering to ``rung.data`` if
+                # ``self._mode == "max"``
+                if self._mode == "min":
+                    values_ranking = list(range(len(trial_ids)))
+                else:
+                    values_ranking = list(range(len(trial_ids) - 1, -1, -1))
```
wistuba, line 96: most likely this entire function could be dropped now given that things are sorted already

610v4nn1 (changes requested):
There are changes to PASHA but it is not clear if it is being tested as part of the integration tests. Can you check that before merging?

mseeger:
@610v4nn1 AFAIK, there are no integration tests for PASHA. This is the responsibility of the contributor. Short tests can be inserted into `.github/workflows/integ_tests.yml`, longer ones into `.github/workflows/end-to-end-tests.yml`.

mseeger:
I see that `examples/launch_pasha_nasbench201.py` is run as end-to-end-tests. We can run them before this is merged, sure.

mseeger:
I just ran `examples/launch_pasha_nasbench201.py` locally.

After the change here:

```
4 trials running, 253 finished (0 until the end), 1.83s wallclock-time

metric_valid_error: best 0.5637999773025513 for trial-id 186
```

Before the change here:

```
4 trials running, 253 finished (0 until the end), 1.92s wallclock-time

metric_valid_error: best 0.5637999773025513 for trial-id 186
```

So this looks good.

mseeger:
If you have some experiments done with PASHA, you could rerun some of them and check.

mseeger:
I re-ran experiments with ASHA and MOBSTER. For ASHA, they are on top of each other (old, new). For MOBSTER, they are very close.

syne_tune/optimizer/schedulers/searchers/dyhpo/dyhpo_searcher.py
jgolebiowski: Nice change!

jgolebiowski (commented):
This looks like a good change. I like the switch to Rungs a lot, makes the abstraction cleaner. With that in mind I am not that experienced in the Hyperband implementation so I might have missed something, I can approve it but we can ask another reviewer to take a look to be certain.

mseeger:
In fact, also Martin looked at it.

jgolebiowski (approved):
LGTM!

PR #605, opened 2023-03-20
feat: automatic termination criterion

aaronkl:
Implements the automatic termination criterion proposed by Makarova et al.

----


syne_tune/stopping_criterions/automatic_termination_criterion.py
```diff
@@ -0,0 +1,159 @@
...
+            evaluations = evaluations.reshape((-1, 1))
+
+        gp = SingleTaskGP(observations, evaluations)
+        mll = ExactMarginalLogLikelihood(gp.likelihood, gp)
+        fit_gpytorch_mll(mll, max_attempts=1)
+        acq = UpperConfidenceBound(
+            model=gp,
+            beta=self._beta,
+        )
+        _, max_ucb = optimize_acqf(
```
geoalgo, line 142: We should try catch this one. In case cholesky fails (which will be frequent with 3 restarts), we should not stop.

syne_tune/stopping_criterions/automatic_termination_criterion.py
```diff
@@ -0,0 +1,159 @@
...
+    def __init__(
+        self,
+        config_space: Dict[str, Any],
+        metric: str,
+        threshold: float,
+        mode: str = "min",
+        beta: float = 1.0,
+        seed: int = None,
+        warm_up: int = 10,
+        topq: float = 0.5,
```
geoalgo, line 68: You should probably assert that the value in ]0,1[, the user may pass you an integer for instance by mistake.

syne_tune/stopping_criterions/automatic_termination_criterion.py
```diff
@@ -0,0 +1,159 @@
...
+    and Matthias Seeger and Cedric Archambeau
+    First Conference on Automated Machine Learning (Main Track)
+    2022
+
+    :param config_space: Configuration space for evaluation function
+    :param metric: The metric to be monitored.
+    :param mode: The mode to select the top results ("min" or "max")
+    :param threshold: The threshold on the regret. We stop the optimization process when it's unlikely that the
+    expected decrease in regret compared to the global optimum is higher than this threshold.
+    :param beta: Multiplier on the standard deviation to compute the upper and lower confidence bound.
```
geoalgo, line 53: Why would `beta` be different than 1?

syne_tune/stopping_criterions/automatic_termination_criterion.py
mseeger: All of such stuff is already there. You can replicate all this by using BOTorch, or you can just use the internal code that is already there.

mseeger:
Some folks using Syne Tune cannot afford BOTorch dependencies, they will not be able to use this. Maybe, one should consider re-implementing this so that internal BO code of Syne Tune is being used.

syne_tune/stopping_criterions/automatic_termination_criterion.py
aaronkl: Theoretically you want to adapt `beta` to guarantee convergence of the optimization process (see GP-UCB paper). I agree that in practice it is almost always set to 1.

syne_tune/stopping_criterions/automatic_termination_criterion.py
```diff
@@ -0,0 +1,159 @@
...
+    Automatic Termination for Hyperparameter Optimization
+    Anastasia Makarova and Huibin Shen and Valerio Perrone and Aaron Klein and Jean Baptiste Faddoul and Andreas Krause
+    and Matthias Seeger and Cedric Archambeau
+    First Conference on Automated Machine Learning (Main Track)
+    2022
+
+    :param config_space: Configuration space for evaluation function
+    :param metric: The metric to be monitored.
+    :param mode: The mode to select the top results ("min" or "max")
+    :param threshold: The threshold on the regret. We stop the optimization process when it's unlikely that the
```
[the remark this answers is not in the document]
aaronkl: Thanks for the suggestions. I agree that ranks would be much easier to set. How would you reparametrize this?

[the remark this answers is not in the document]
geoalgo: I was thinking to apply rank normalization on the values but now I see that it would not be trivial to set the stopping criterion this way either.

tst/test_automatic_termination_criterion.py
```diff
@@ -0,0 +1,40 @@
...
+# permissions and limitations under the License.
+# from syne_tune.stopping_criterions.automatic_termination_criterion import (
+#     AutomaticTerminationCriterion,
+# )
+# from syne_tune.backend.trial_status import Trial, Status
+# from syne_tune.tuning_status import TuningStatus
+# from syne_tune.config_space import uniform
+
+#
+# def test_automatic_termination():
```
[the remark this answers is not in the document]
aaronkl: This would mean we have to add botorch to the requirements for the unit tests (extra). I am happy to do that, but I would send another PR for that

geoalgo (approved):

PR #606, opened 2023-03-21
Move utils from benchmarking to syne_tune, and ensure in docs that be…

mseeger:
Makes sure that:

- docs say that for anything using `benchmarking`, including using training code or benchmark defs from there, you need to have installed ST from source
- helper functions to write training scripts move to `syne_tune.utils`, so they can be used by people who have `pip` installed us

----


mseeger:
I am also running end-to-end-tests to verify it still works

benchmarking/utils/__init__.py
```diff
@@ -10,28 +10,10 @@
...
-    load_config_json,
-)
 
 __all__ = [
-    "add_checkpointing_to_argparse",
-    "resume_from_checkpointed_model",
-    "checkpoint_model_at_rung_level",
-    "pytorch_load_save_functions",
-    "parse_bool",
     "get_cost_model_for_batch_size",
```
wesk, line 18: Quick code search shows that the only code referencing `get_cost_model_for_batch_size` is commented out. Is it specific to benchmarking, or is it also a general utility? Wondering if it makes sense to also move or leave as-is.

docs/source/tutorials/basics/basics_setup.rst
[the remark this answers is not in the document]
wesk: Something else we could do is have the docs directly reference the code in the `examples/` folder, rather than the `docs/` folder, so we avoid two competing sources of truth for example code. So that the examples that we show in the docs are one-and-the-same with what the docs reference. And customers could directly run those files, rather than copy-pasting from the docs into their own files. 

If we do this, we can then add in the basics_randomsearch file (https://syne-tune.readthedocs.io/en/latest/tutorials/basics/basics_randomsearch.html) something like the following instructions:

```
To run the trial using random search:
1. (pre-requisite step) Clone Syne Tune with `git clone git@github.com:awslabs/syne-tune.git`
2. Run the following command: ` python examples/hpo_main.py --method RS`
```

And the same for ASHA-STOP, etc:

```
To run the trial using ASHA-STOP:
1. Run the following command: ` python examples/hpo_main.py --method ASHA-STOP`

```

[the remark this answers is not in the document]
wesk: Some of this feedback goes beyond the scope of your PR here, so I'd be happy to implement these suggestions in a separate PR that I raise myself, to keep the scope of this PR smaller. Let me know what you think!

docs/source/tutorials/basics/README.rst
```diff
@@ -5,6 +5,11 @@ This tutorial provides a first overview of Syne Tune. You will learn about the
 most important concepts of automated hyperparameter tuning, and how to make it
 work for your setup.
 
+.. note::
+   This tutorial makes use of the *benchmarking* framework, you need to have
+   `installed Syne Tune from sources <getting_started.html#installation>`__
```
wesk, line 10:
```suggestion
   `installed Syne Tune from source <getting_started.html#installation>`__
```

docs/source/examples.rst
```diff
@@ -23,6 +23,8 @@ Fine-Tuning Hugging Face Model for Sentiment Classification
 
 **Requirements**:
 
+* Needs ``benchmarking`` framework, which requires Syne Tune to be installed
```
wesk, line 26:
```suggestion
* Needs Syne Tune's ``benchmarking`` framework, which requires Syne Tune to be installed
```

docs/source/examples.rst
```diff
@@ -119,6 +123,8 @@ Joint Tuning of Instance Type and Hyperparameters using MOASHA
...
```
wesk, line 126:
```suggestion
* Needs  Syne Tune's ``benchmarking`` framework, which requires Syne Tune to be installed
```

docs/source/examples.rst
```diff
@@ -161,10 +167,12 @@ PASHA: Efficient HPO and NAS with Progressive Resource Allocation
...
    :start-after: # permissions and limitations under the License.
 
-This script uses the simulator backend to run an experiment on NASBench-201.
-It takes only a few seconds to run, but it needs ``nasbench201`` blackbox
-to be downloaded and preprocessed, which can take a while when done
-for the first time.
+**Requirements**:
+
+* Needs ``benchmarking`` framework, which requires Syne Tune to be installed
+  `from sources <getting_started.html#installation>`__.
```
wesk, line 173: Same suggestion as the other instances:

```suggestion
* Needs  Syne Tune's``benchmarking`` framework, which requires Syne Tune to be installed
  `from source <getting_started.html#installation>`__.
```

wesk (commented):
Some minor phrasing suggestions, otherwise looks good. Also, I have an idea for a more substantial docs refactor that could be done as a separate PR.

mseeger:
OK, end-to-end tests run through

benchmarking/utils/__init__.py
mseeger: Yeah, that is why I left it here. I implemented cost models a while ago, for a certain project. This is useful code, but right now it is not used.

docs/source/tutorials/basics/basics_setup.rst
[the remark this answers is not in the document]
mseeger: I think we should allow for code in the docs as well. Code in `examples/` stands on its own, while code in `docs/` only makes sense in the context of the tutorial.

I think the tutorial already says how to run this code. And it is not enough to just clone the repo, you also need to create a virtual env, as is explained in `getting_started`

wesk (approved):

PR #611, opened 2023-03-30
feat: Plotting functions to analyse multi-objective experiments.

jgolebiowski:
1. New function in experiments plot the hypervolume error for MO experiment
2. New function in multiobjective searchers computes the hypervolume
3. Expanded the current Experiment plotting functions to work with multi-objective optimization problems
4. Added requirements-moo to import pymoo needed for hypervolume computation

----

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,6 +80,7 @@ setup(
         "aws": required_aws,
         "yahpo": required_yahpo,
         "dev": required_dev,
+        "moo": required_moo,
         "extra": required_extra,
     },
     install_requires=required_core,


syne_tune/experiments.py
```diff
@@ -61,30 +64,87 @@ def creation_date(self):
         """
         return datetime.fromtimestamp(self.metadata[ST_TUNER_CREATION_TIMESTAMP])
 
-    def plot(self, **plt_kwargs):
+    def plot_hypervolume(
+        self, reference_points: np.ndarray = None, figure_path: str = None, **plt_kwargs
+    ):
+        """Plot best hyervolume value as function of wallclock time
+
+        :param reference_points: Reference points for hypervolume calculations.
```
mseeger, line 72: nit: If `:param:` in docstring is multi-line, just use one tab for 2-last line

syne_tune/experiments.py
```diff
@@ -93,20 +153,35 @@ def metric_names(self) -> List[str]:
...
         :return: Configuration corresponding to best metric value
         """
-        metric_names = self.metric_names()
-        metric_mode = self.metric_mode()
-
-        if len(metric_names) > 1:
-            logging.warning(
-                "Several metrics exists so the best is not defined, this will "
-                f"return the best w.r.t. the first metric of {metric_names}."
+        if isinstance(metric, str):
```
mseeger, line 163: This codeblock (allowing `metric` to be str or int, and map to index and name) looks more generally useful. Method?

mseeger: Including determining `metric_mode`

syne_tune/optimizer/schedulers/multiobjective/utils.py
```diff
@@ -0,0 +1,50 @@
...
+    reference_points: np.ndarray = None,
+    return_progress: bool = False,
+):
+    """
+    Compute the hypervolume of all results based on reference points
+
+    :param results_df: Array with experiment results ordered by time with shape (npoints, ndimensions)
+    :param reference_points: Reference points for hypervolume calculations.
+                             If None, the maximum values of each metric is used.
+    :param return_progress: If True, returns an array with hypervolumes given by an increasing range of points.
```
mseeger, line 37: This almost calls for another function, something like `cumulative_hypervolume`. The behaviour is very different in these two cases, as is the runtime.

syne_tune/optimizer/schedulers/multiobjective/utils.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+EPSILON = 1e-6
+
+
+def hypervolume(
+    results_array: np.ndarray,
+    reference_points: np.ndarray = None,
+    return_progress: bool = False,
+):
```
mseeger, line 30: Return type. Would be easier if this were two functions

syne_tune/optimizer/schedulers/multiobjective/utils.py
```diff
@@ -0,0 +1,50 @@
...
+    if reference_points is None:
+        reference_points = results_array.max(axis=0) * (1 + EPSILON) + EPSILON
+    indicator_fn = HV(ref_point=reference_points)
+
+    if not return_progress:
+        return indicator_fn(results_array)
+
+    hypervolume_indicator = np.zeros(shape=len(results_array))
+    for idx in range(len(results_array)):
+        hypervolume_indicator[idx] = indicator_fn(results_array[0:idx])
```
mseeger, line 49: What does this give for the empty array, or for only one point?

syne_tune/optimizer/schedulers/multiobjective/utils.py
```diff
@@ -0,0 +1,50 @@
...
+def hypervolume(
+    results_array: np.ndarray,
+    reference_points: np.ndarray = None,
+    return_progress: bool = False,
+):
+    """
+    Compute the hypervolume of all results based on reference points
+
+    :param results_df: Array with experiment results ordered by time with shape (npoints, ndimensions)
+    :param reference_points: Reference points for hypervolume calculations.
```
mseeger, line 35: Would this not rather be `reference_point`? It is a single datapoint, right?

mseeger:
Additional comment (not for this PR): We now have the option to report metrics to CloudWatch when using remote tuning. This would be a neat contender. Check `syne_tune.remote.remote_metrics_callback`.

syne_tune/experiments.py
```diff
@@ -61,30 +64,87 @@ def creation_date(self):
...
+
+        :param reference_points: Reference points for hypervolume calculations.
+                                 If None, the maximum values of each metric is used.
+        :param figure_path: If specified, defines the path where the figure will be saved.
+                            If None, the figure is shown
+        :param plt_kwargs: Arguments to :func:`matplotlib.pyplot.plot`
+        """
+        import matplotlib.pyplot as plt
+
+        metric_names = self.metric_names()
```
geoalgo, line 80: You would very likely want to customize the metrics that can be passed here, ie to be able to pass `metric_names` as argument and use the full list by default (as you often want to optimize only a subset of available metrics, for instance in the benchmark in blackbox repository which contains many metrics).

syne_tune/experiments.py
jgolebiowski: OK makes sense

syne_tune/experiments.py
jgolebiowski: sure I can extract that

syne_tune/optimizer/schedulers/multiobjective/utils.py
jgolebiowski: For empty array it will return an empty array, for one element it returns zero. I can make it explicit in the docstring

requirements-moo.txt
```diff
@@ -0,0 +1 @@
+pymoo>=0.6.0
```
wesk, line 1: We already have at least one multi-objective method in Syne Tune, right? If that's true, under what circumstances do customers need this dependency?

mseeger: I think we discussed this yesterday. The current MO methods in Syne Tune (mostly MO-ASHA) do not require dependencies beyond the core. But now, Jacek is owning this part, and he is introducing new tooling. I foresee there will be extra methods, also schedulers, which require these. For now, he only contributes a plotting functionality.

mseeger:
Sure, this was just a remark. Anyway, we still need somebody to figure out a nice-looking CloudWatch setup, before this makes sense.

tst/schedulers/multiobjective/test_utils.py
```diff
@@ -0,0 +1,32 @@
...
+    hypervolume,
+    hypervolume_cumulative,
+)
+
+
+def test_hypervolume():
+    ref_point = np.array([2, 2])
+    points = np.array([[1, 0], [0.5, 0.5], [0, 1], [1.5, 0.75]])
+    hv = hypervolume(results_array=points, reference_point=ref_point)
+    assert hv == 3.25
```
wesk, line 25: Just for the sake of verifying that this new function works correctly, how do we know that the magic number of 3.25 is correct here? Can I compute this example in my head, or is there an even simpler example that could also be used to verify the computation?

syne_tune/optimizer/schedulers/multiobjective/utils.py
```diff
@@ -0,0 +1,63 @@
...
+try:
+    from pymoo.indicators.hv import HV
+except ImportError:
+    print(try_import_moo_message())
+
+
+EPSILON = 1e-6
+
+
+def hypervolume(
```
wesk, line 26: Can you expand on what this function is doing? Most of the heavy-lifting is coming from pymoo's HV function, but what does it mean when reference_point is None?

syne_tune/optimizer/schedulers/multiobjective/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    results_array: np.ndarray,
+    reference_point: np.ndarray = None,
+) -> float:
+    """
+    Compute the hypervolume of all results based on reference points
+
+    :param results_df: Array with experiment results ordered by time with shape (npoints, ndimensions)
```
wesk, line 33:
```suggestion
    :param results_array: Array with experiment results ordered by time with shape (npoints, ndimensions)
```

syne_tune/optimizer/schedulers/multiobjective/utils.py
```diff
@@ -0,0 +1,63 @@
...
+) -> np.ndarray:
+    """
+    Compute the cumulative hypervolume of all results based on reference points
+    Returns an array with hypervolumes given by an increasing range of points.
+    ``return_array[idx] = hypervolume(results_array[0: idx])``
+
+    :param results_df: Array with experiment results ordered by time with shape (npoints, ndimensions).
+    :param reference_point: Reference points for hypervolume calculations.
+        If None, the maximum values of each metric is used.
+    :return: Cumulative hypervolume array with dimensions (npoints, )
```
wesk, line 55: Is `(npoints, )` the correct dimensions?

tst/schedulers/multiobjective/test_utils.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def test_hypervolume_progress():
+    ref_point = np.array([2, 2])
+    points = np.array([[1, 0], [0.5, 0.5], [0, 1], [1.5, 0.75]])
+    hv = hypervolume_cumulative(results_array=points, reference_point=ref_point)
+    assert np.allclose(hv, np.array([0.0, 2.0, 2.75, 3.25]))
```
wesk, line 32: Is the `allclose` assertion because it's floating-point and thus non-deterministic? In the `test_hypervolume` test, why is it then able to be a strict equals check `hv == 3.25`?

tst/schedulers/multiobjective/test_utils.py
jgolebiowski: I suppose you can compute this in your head but using a piece of paper is simpler :) I think this example strikes the right balance between simplicity and not being trivial but I can add a simpler case.

syne_tune/optimizer/schedulers/multiobjective/utils.py
jgolebiowski: It is computing hypervolume indicator for the provided point and reference. Its really just a convinienece wrapper around pymoo. The thinking is that if we want to replace the library for something else like pygmo we can just replace the implementation here rather than scount for it in the code. 

Re second question updated the docstring to be clearer

syne_tune/optimizer/schedulers/multiobjective/utils.py
jgolebiowski: Yes, its a one-dime array wiht the length equal to number of reference points.

tst/schedulers/multiobjective/test_utils.py
jgolebiowski: This is for floating point precision. Updated the other test too, good catch.

setup.py
```diff
@@ -38,6 +38,7 @@ def load_benchmark_requirements():
 required_benchmarks = load_benchmark_requirements()
 required_dev = load_requirements("requirements-dev.txt")
 required_aws = load_requirements("requirements-aws.txt")
+required_moo = load_requirements("requirements-moo.txt")
```
mseeger, line 41: Please also extend `extras_require` in `setup`

mseeger (dismissed):
Just one final change in `setup.py`, then I am fine.

wesk (approved):

PR #612, opened 2023-04-01
feature: Allow for customized extra results to be written to results.csv.zip

mseeger:
At the moment, we write time-stamped results to ``results.csv.zip`` (a dataframe), and the serialized tuner to ``tuner.dill``.
This PR allows to (optionally) write extra results to ``results.csv.zip``. These are appended as new columns.

An example use case is given.

----


.github/workflows/integ-tests.yml
```diff
@@ -111,7 +111,7 @@ jobs:
     with:
       script-path: examples/launch_transfer_learning_example.py
 
-  launch_height_simple:
```
mseeger, line 114: Note that ``launch_height_simple.py`` is not an executable Python script, so does not make sense to have it here. It is also duplicated by others, like ``launch_height_baselines``.

benchmarking/commons/hpo_main_local.py
```diff
@@ -172,25 +169,26 @@ def create_objects_for_tuner(
...
+    tuner_kwargs["callbacks"] = callbacks
     return tuner_kwargs
 
 
 def start_benchmark_local_backend(
     configuration: ConfigDict,
     methods: MethodDefinitions,
     benchmark_definitions: RealBenchmarkDefinitions,
-    post_processing: Optional[PostProcessingType] = None,
+    final_results: Optional[FinalResultsComposer] = None,
```
mseeger, line 191: Replaces ``post_processing``. Instead of running some code at the end, we extract some extra results which are stored. This is more useful.

benchmarking/commons/utils.py
```diff
@@ -15,6 +15,11 @@
 import numpy as np
 
 from syne_tune.util import s3_experiment_path
+from syne_tune.constants import (
+    ST_METADATA_FILENAME,
```
mseeger, line 19: This is cleaning things up. In the old code, file names were hardcoded. Now, they are in ``constants.py``

.github/workflows/integ-tests.yml
wesk: I just ran `python examples/launch_height_simple.py` and got the following output: 

```
% python examples/launch_height_simple.py
--------------------
Resource summary (last result is reported):
 trial_id     status  iter  width  height  epochs  epoch  mean_loss  worker-time
        0    Stopped    41     10      10     100   41.0   1.243902     4.237480
        1    Stopped    41     13      11     100   41.0   1.288679     4.234309
        2  Completed   100     10       1     100  100.0   0.200000    10.316471
        3    Stopped    41      2      10     100   41.0   2.111111     4.232994
        4    Stopped    41     13      14     100   41.0   1.588679     4.220124
        5    Stopped    41     12      16     100   41.0   1.804082     4.220239
        6    Stopped    41      1       3     100   41.0   2.300000     4.216881
        7    Stopped    41     17       5     100   41.0   0.644928     4.233574
        8    Stopped    41     14       8     100   41.0   0.975439     4.229480
        9    Stopped    41      6       7     100   41.0   1.100000     4.222577
       10    Stopped    41     19      13     100   41.0   1.429870     4.238196
       11    Stopped    90     16       2     100   90.0   0.269735     9.291159
       12    Stopped    41     17       7     100   41.0   0.844928     4.223391
       13    Stopped    41     18       5     100   41.0   0.636986     4.221838
       14    Stopped    41      1      11     100   41.0   3.100000     4.227698
       15    Stopped    41     20      19     100   41.0   2.023457     4.226891
       16    Stopped    41     18      17     100   41.0   1.836986     4.225258
       17 InProgress     0     20       2     100      -          -            -
       18 InProgress     0     20      17     100      -          -            -
       19 InProgress     0     15      16     100      -          -            -
       20 InProgress     0     17      16     100      -          -            -
4 trials running, 17 finished (1 until the end), 30.31s wallclock-time

mean_loss: best 0.2 for trial-id 2

```

So I believe it is executable. 

I'm fine with removing it if it's redundant / covered by other tests, just to my knowledge, it is indeed executable.

benchmarking/commons/hpo_main_local.py
wesk: Are there any known uses of customers using `post_processing`? 

I agree that `final_results` is a great new feature -- just wondering before we remove `post_processing` if there are any use-cases that will break from the removal.

Especially since this is in `benchmarking`, I think it's very low-risk to update.

.github/workflows/integ-tests.yml
mseeger: It is just in examples because it is included in our docs. If it runs, I'll keep it in there.

benchmarking/commons/hpo_main_local.py
mseeger: Good point. I don't think anybody except me uses this right now.

syne_tune/constants.py
```diff
@@ -72,6 +72,20 @@
 """Name for ``upload_dir`` in ``RemoteTuner``"""  # pylint: disable=W0105
 
 
+# File names
+
+ST_RESULTS_DATAFRAME_FILENAME = "results.csv.zip"
```
wesk, line 77: Would it make sense to add the fact that this contains intermediate results, to disambiguate from the final results JSON? 

Also, to check my understanding, this is written always, whereas the final results JSON is not always written? 

Question: at the end of tuning, would this `results.csv.zip` contain the same info as the final results JSON?

syne_tune/results_callback.py
```diff
@@ -0,0 +1,158 @@
...
+    Default implementation of :class:`~TunerCallback` which records all
+    reported results, and allows to store them as CSV file.
+
+    :param add_wallclock_time: If ``True``, wallclock time since call of
+        ``on_tuning_start`` is stored as
+        :const:`~syne_tune.constants.ST_TUNER_TIME`.
+    :param final_results_composer: Optional. If given, this is called in
+        :meth:`on_tuning_end`, and the resulting dictionary is written as
+        ``{tuner.tuner_path}/{ST_FINAL_RESULTS_FILENAME}``.
+    :param store_final_results_intermediately: If ``True`` and
```
wesk, line 67: Do you mean `intermediately` or perhaps `intermittently` or `immediately`?

As a customer, I might be confused, because are they really "final" results if the tuning is still in-progress? Are they not "intermediate" results in some sense?

syne_tune/constants.py
mseeger: No, the difference is not when these two are written (they are both written intermediately, so if things crash, you have something until then), but `results.csv` is a dataframe with a time index, so contains records of events during the experiment (e.g., when certain reports were obtained, ...). `final-results.json` is a set of results not indexed by time.

syne_tune/results_callback.py
```diff
@@ -0,0 +1,158 @@
...
+            self._start_time_stamp = perf_counter()
+        if self._final_results_composer is not None:
+            self._tuner = tuner
+
+    def _write_final_results(self):
+        if self._final_results_composer is not None:
+            final_results = self._final_results_composer(self._tuner)
+            if final_results is not None:
+                path = self._tuner.tuner_path / ST_FINAL_RESULTS_FILENAME
+                dump_json_with_numpy(final_results, path)
```
wesk, line 151: How does this new method relate to this existing way of getting the final hyperparameters: https://github.com/awslabs/syne-tune/blob/main/examples/launch_plot_results.py#L69

```
tuning_experiment = load_experiment(tuner.name)
print(tuning_experiment)
print(f"best result found: {tuning_experiment.best_config()}")
```

Is this `_final_results_composer` needed if it's already possible to call `.best_config()`? Could we not re-use what `.best_config()` is doing?

syne_tune/results_callback.py
mseeger: I'll change it to "periodically". Note we also store `results.csv.zip` periodically, so if an experiment crashes after 10 hours, you still have something.

tst/schedulers/test_searchers.py
```diff
@@ -66,7 +66,7 @@
 
 # Does not contain ASHABORE, because >10 secs on CI
 # TODO: Dig more, why is ASHABORE more expensive than BORE here?
-@pytest.mark.timeout(10)
+@pytest.mark.timeout(12)
```
wesk, line 69: Looks like writing the files to disk makes the test run longer? Is this with intermediate checkpoints enabled, or is this just writing final output to disk?

mseeger: I am not sure. I just increased this because I had this test failing. I do not think this has anything to do with the change here, to be honest. The change here is not activated in this test, right? Final results are only written when you activate the feature.

syne_tune/constants.py
wesk: Thanks for the clarification! Sounds good.

syne_tune/results_callback.py
mseeger: Final results are not best parameters. They can be anything. In the example I have here, they are statistics recorded by the scheduler, so nothing that would even be in `results.csv.zip`.

In the next PR, I am using this to record performance statistics for checkpoint removal. Again, this is not stored in `results.csv`.

These final results are optional. If you can obtain your answer from the `results.csv`, you should do that. But often, this is not possible.

tst/schedulers/test_searchers.py
wesk: We can bump the limit, but if it was passing reliably before this PR, and starts timing out with this PR, then it would be useful to know why.

Not necessarily a blocker but something interesting

syne_tune/results_callback.py
mseeger: You have a good point, I'll do that

mseeger:
I will rework this, so information is appended to the `results.csv` dataframe. Thanks Wes, this is indeed a better solution.

mseeger:
OK, this is totally rewritten. The extra results are now appended to the default ones, by extending the dataframe by additional columns. This has the advantage of not needing another file, and of getting time-stamped results.

The disadvantage is we cannot write out final results of complex structure (e.g., lists or dicts). But that is OK.

examples/launch_height_extra_results.py
```diff
@@ -0,0 +1,97 @@
...
+        return scheduler.terminator._rung_systems[0].summary_schedule_records()
+
+    def keys(self) -> List[str]:
+        return DyHPORungSystem.summary_schedule_keys()
+
+
+if __name__ == "__main__":
+    logging.getLogger().setLevel(logging.INFO)
+
+    random_seed = 31415927
```
wesk, line 47: nice seed choice 🥧

wesk (approved):
Very intuitive, looks good!

PR #614, opened 2023-04-04
feat: Speculative early checkpoint removal for async multi-fidelity

mseeger:
This features a new technique to remove checkpoints early for ASHA, MOBSTER and the like. The rough idea is to rank paused trials by expected cost of having to train from scratch when resuming them (which is the case when the checkpoint is missing). This needs a prediction for the probability of getting resumed in the remainder of experiment time.

There is a note on the method, which I can share with the reviewer.

Motivated by Aaron's problems of disk running full when tuning large Hugging Face models.

----


syne_tune/callbacks/hyperband_remove_checkpoints_callback.py
```diff
@@ -0,0 +1,522 @@
...
+    def num_total(self):
+        return self._num_total
+
+    def posterior_mean(self) -> float:
+        a_posterior = self._beta_a + self._num_one
+        aplusb_posterior = self._beta_size + self._num_total
+        return a_posterior / aplusb_posterior
+
+
+class HyperbandRemoveCheckpointsCommon(TunerCallback):
```
mseeger, line 74: The method is implemented as a callback, which runs alongside the tuner.

syne_tune/callbacks/hyperband_remove_checkpoints_callback.py
```diff
@@ -0,0 +1,522 @@
...
+    """
+    Common base class for :class:`HyperbandRemoveCheckpointsCallback` and
+    :class:`HyperbandRemoveCheckpointsBaselineCallback`.
+    """
+
+    def __init__(
+        self,
+        max_num_checkpoints: int,
```
mseeger, line 82: The user only really has to set this one. The other args are filled in by looking at the scheduler and the stopping rule

syne_tune/callbacks/hyperband_remove_checkpoints_callback.py
```diff
@@ -0,0 +1,522 @@
...
+        # which were nevertheless resumed
+        self._trials_resumed_without_checkpoint = []
+        self._num_checkpoints_removed = 0
+        self._num_trials_resumed = 0
+
+    @property
+    def num_checkpoints_removed(self) -> int:
+        return self._num_checkpoints_removed
+
+    def _filter_paused_trials(
```
mseeger, line 131: Only want paused trials which still have a checkpoint

syne_tune/tuner_callback.py
```diff
@@ -101,3 +101,17 @@ def on_tuning_sleep(self, sleep_time: float):
         :param sleep_time: Time (in secs) for which tuner has just slept
         """
         pass
+
+    def on_start_trial(self, trial: Trial):
```
mseeger, line 105: These callback methods were missing before

PR #617, opened 2023-04-05
chore: Moved scheduler metadata generation

jgolebiowski:
Moved scheduler metadata generation from Tuner to the scheduler. Now different schedulers can override this metod to provide their own metadata.
----

### DIFF
--- a/syne_tune/optimizer/scheduler.py
+++ b/syne_tune/optimizer/scheduler.py
@@ -15,7 +15,12 @@ from typing import Optional, List, Dict, Any, Union
 import logging
 
 from syne_tune.backend.trial_status import Trial
-from syne_tune.config_space import non_constant_hyperparameter_keys, cast_config_values
+from syne_tune.config_space import (
+    non_constant_hyperparameter_keys,
+    cast_config_values,
+    config_space_to_json_dict,
+)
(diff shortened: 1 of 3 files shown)


syne_tune/tuner.py
```diff
@@ -682,4 +674,4 @@ def _default_callback():
         """
         :return: Default callback to store results
         """
-        return StoreResultsCallback()
+        return StoreResultsCallback()
```
jgolebiowski, line 677: This is just adding a newline after https://github.com/awslabs/syne-tune/blob/main/syne_tune/tuner.py#L685 from mainline

syne_tune/optimizer/scheduler.py
```diff
@@ -296,3 +301,17 @@ def trials_checkpoints_can_be_removed(self) -> List[int]:
         :return: IDs of paused trials for which checkpoints can be removed
         """
         return []  # Default is not to participate in early checkpoint removal
+
+    def metadata(self) -> Dict[str, Any]:
```
mseeger, line 305: Thanks for this clean-up

mseeger (dismissed):

mseeger (approved):

PR #619, opened 2023-04-06
feat: simple linear scalarization scheduler

jgolebiowski:
Added a simple linear scalarization scheduler that combines multiple objectives into one using weighted mean and optimizes the resulting metric with any singe-objective FIFO optimizer.

----

### DIFF
--- a/syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
+++ b/syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
@@ -132,6 +132,13 @@ class LinearScalarizedScheduler(TrialScheduler):
         """
         return self.base_scheduler.on_trial_remove(trial)
 
+    def trials_checkpoints_can_be_removed(self) -> List[int]:
+        """
+        See the docstring of the chosen base_scheduler for details
+        :return: IDs of paused trials for which checkpoints can be removed
+        """
+        return self.base_scheduler.trials_checkpoints_can_be_removed()
+
     def metric_names(self) -> List[str]:
         """
         :return: List of metric names.
--- a/syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
+++ b/syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
@@ -113,11 +113,11 @@ class LinearScalarizedScheduler(TrialScheduler):
         mo_results = np.array([result[item] for item in self.metric_names()])
         return np.sum(np.multiply(mo_results, self.scalarization_weights))
 
-    def suggest(self, trial_id: int) -> Optional[TrialSuggestion]:
-        """Returns a suggestion for a new trial, or one to be resumed
+    def _suggest(self, trial_id: int) -> Optional[TrialSuggestion]:
+        """Implements ``suggest``, except for basic postprocessing of config.
         See the docstring of the chosen base_scheduler for details
         """
-        return self.base_scheduler.suggest(trial_id)
+        return self.base_scheduler._suggest(trial_id)
 
     def on_trial_add(self, trial: Trial):
         """Called when a new trial is added to the trial runner.
--- a/syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
+++ b/syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
@@ -13,13 +13,13 @@
 import logging
 from collections.abc import Iterable
 from itertools import groupby
-from typing import Dict, Any, List, Union
+from typing import Dict, Any, List, Union, Callable, Optional
 
 import numpy as np
 
 from syne_tune.backend.trial_status import Trial
-from syne_tune.optimizer.schedulers import FIFOScheduler
-from syne_tune.optimizer.schedulers.fifo import _to_list
+from syne_tune.optimizer.scheduler import TrialScheduler, TrialSuggestion
+from syne_tune.optimizer.schedulers.fifo import FIFOScheduler
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
```diff
@@ -0,0 +1,160 @@
...
+import numpy as np
+
+from syne_tune.backend.trial_status import Trial
+from syne_tune.optimizer.scheduler import TrialScheduler, TrialSuggestion
+from syne_tune.optimizer.schedulers.fifo import FIFOScheduler
+
+logger = logging.getLogger(__name__)
+
+
+def _all_equal(iterable: Iterable) -> bool:
```
mseeger, line 27: Looks pretty cryptic. Why not something like
`a = lst[0]; all(x == a for x in lst[1:])`

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
```diff
@@ -0,0 +1,160 @@
...
+        super(LinearScalarizedScheduler, self).__init__(config_space)
+        if scalarization_weights is None:
+            scalarization_weights = np.ones(shape=len(metric))
+        self.scalarization_weights = np.asarray(scalarization_weights)
+
+        self.metric = metric
+        self.mode = mode
+
+        self.single_objective_metric = f"scalarized_{'_'.join(metric)}"
+        assert len(mode) >= 1, "At least one mode must be provided"
```
mseeger, line 73: I'd instead insist on `len(metric) > 1`, otherwise this class should not be used.

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
```diff
@@ -0,0 +1,160 @@
...
+    The scalarized single objective is named: 'scalarized_<metric1>_<metric2>_..._<metricN>'
+
+
+    :param base_scheduler: Class for the single-objective scheduler to be used on the scalarized objective.
+        It will be initialized inside this Scheduler
+    :param config_space: Configuration space for evaluation function
+    :param metric: Names of metrics to optimize
+    :param mode: Modes of metrics to optimize (min or max). All must be matching.
+    :param scalarization_weights: Weights used to scalarize objectives, if None an array of 1s is used
+    :param kwargs: Additional arguments to base_scheduler
```
mseeger, line 48: Call this `base_scheduler_kwargs` then

mseeger: And say which arguments you are settting here, so should not be in `base_scheduler_kwargs`

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
```diff
@@ -0,0 +1,160 @@
...
+    single_objective_metric: str
+    base_scheduler: TrialScheduler
+
+    def __init__(
+        self,
+        config_space: Dict[str, Any],
+        metric: List[str],
+        mode: Union[List[str], str] = "min",
+        scalarization_weights: Union[np.ndarray, List[float]] = None,
+        base_scheduler: Callable[[Any], TrialScheduler] = FIFOScheduler,
```
mseeger, line 61: Call this something else, like `base_scheduler_maker` or so.

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
```diff
@@ -0,0 +1,160 @@
...
+            mode=single_objective_mode,
+            **kwargs,
+        )
+
+    def _scalarized_metric(self, result: Dict[str, Any]) -> float:
+        if isinstance(self.base_scheduler, FIFOScheduler):
+            FIFOScheduler._check_keys_of_result(result, self.metric_names())
+
+        mo_results = np.array([result[item] for item in self.metric_names()])
+        logger.debug(
```
mseeger, line 92: Not a useful log statement. Best remove. Otherwise, output the values

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
```diff
@@ -0,0 +1,160 @@
...
+            f"Scalarizing {self.metric_names()} into {self.single_objective_metric} using linear combination"
+        )
+        return np.sum(np.multiply(mo_results, self.scalarization_weights))
+
+    def suggest(self, trial_id: int) -> Optional[TrialSuggestion]:
```
mseeger, line 97: Maybe better to overwrite `_suggest`

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
```diff
@@ -0,0 +1,160 @@
...
+        """Called when a trial has failed.
+        See the docstring of the chosen base_scheduler for details
+        """
+        return self.base_scheduler.on_trial_error(trial)
+
+    def on_trial_result(self, trial: Trial, result: Dict[str, Any]) -> str:
+        """Called on each intermediate result reported by a trial.
+        See the docstring of the chosen base_scheduler for details
+        """
+        result[self.single_objective_metric] = self._scalarized_metric(result)
```
mseeger, line 119: Please make a copy:

`result = dict(result, **{self.single_objective_metric: ...})`

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
```diff
@@ -0,0 +1,160 @@
...
+        return self.base_scheduler.on_trial_result(trial, result)
+
+    def on_trial_complete(self, trial: Trial, result: Dict[str, Any]):
+        """Notification for the completion of trial.
+        See the docstring of the chosen base_scheduler for details
+        """
+        result[self.single_objective_metric] = self._scalarized_metric(result)
```
mseeger, line 126: Copy

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
```diff
@@ -0,0 +1,160 @@
...
+        :return: List of metric names.
+        """
+        return self.metric
+
+    def metric_mode(self) -> Union[str, List[str]]:
+        """
+        :return: "min" if target metric is minimized, otherwise "max".
+        """
+        return self.mode
+
```
mseeger, line 146: Missing `trials_checkpoints_can_be_removed`

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
```diff
@@ -0,0 +1,160 @@
...
+    def metadata(self) -> Dict[str, Any]:
+        """
+        :return: Metadata of the scheduler
+        """
+        return {
+            **super(LinearScalarizedScheduler, self).metadata(),
+            "scalarized_metric": self.single_objective_metric,
+        }
+
+    def is_multiobjective_scheduler(self) -> bool:
```
mseeger, line 156: I'd prefer this to become a mixin, instead of an additional method for `TrialScheduler`

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
```diff
@@ -0,0 +1,167 @@
...
+    ):
+        super(LinearScalarizedScheduler, self).__init__(config_space)
+        if scalarization_weights is None:
+            scalarization_weights = np.ones(shape=len(metric))
+        self.scalarization_weights = np.asarray(scalarization_weights)
+
+        self.metric = metric
+        self.mode = mode
+
+        self.single_objective_metric = f"scalarized_{'_'.join(metric)}"
```
jgolebiowski, line 72: See if we should consider an edge-case where this is too long of a string.

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
jgolebiowski: The implementation below returns True for lists with length 0 and works for any iterable, also those that do not support indexing. It might be overkill for this application but its probably good ot have more versatile tooling in the library.

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
jgolebiowski: See L135

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
jgolebiowski: Makes sense. I will add this in a folowup review since its largely a refactor of the base TrialScheduler, not inclusion of this feature. I dont want to conflate the two

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
jgolebiowski: Why do you think so? Both should result in the same behaviour if `base_scheduler` only overrides `_suggest` but in case `base_scheduler` overrides `sugges` we want to call it (`suggest`) to ensure desired behaviour.

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
mseeger: There is `metric_names` on L135 (?)

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
mseeger: Oh yes, please limit the length. I think there is a length limit in SageMaker, maybe David knows.

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
jgolebiowski: That was for previous commit, it is in L160 now.

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
jgolebiowski: I think it can be helpful if the names are short. See limiting code below.

syne_tune/optimizer/schedulers/multiobjective/linear_scalarizer.py
```diff
@@ -0,0 +1,192 @@
...
+        See the docstring of the chosen base_scheduler for details
+        """
+        local_results = {
+            self.single_objective_metric: self._scalarized_metric(result),
+            **result,
+        }
+        return self.base_scheduler.on_trial_complete(trial, local_results)
+
+    def on_trial_remove(self, trial: Trial):
+        """Called to remove trial.
```
mseeger, line 155: It would be OK for me not to add comments when it is just obvious

mseeger (approved):

PR #635, opened 2023-04-20
feat: LocalBackend allows for >1 GPU per trial

### LINKED ISSUE #632: Flexibility in the number of GPUs when using LocalBackend
The `LocalBackend`, through the flag `rotate_gpus` supports allocating either 1 GPU per trial or all the GPUs. It would be great to have the flexibility to choose `num_gpus_per_trial` or equivalent flag to indicate the number of GPUs each trial requires. 

Eg: A node with 4 GPUs and two trials. The first trial can use GPUs 1 & 2 and second 3 &4. PyTorch shouldn't have a problem a `torch.device("cuda:0" or "1")` when the appropriate env var `CUDA_VISIBLE_DEVICES` is set.

mseeger:
Closes #632 

----


benchmarking/nursery/test_num_gpus/requirements-synetune.txt
wesk, line 1: Do we need the new requirements.txt?

benchmarking/nursery/test_num_gpus/launch_remote.py
wesk, line 1: The files under `benchmarking/nursery/test_num_gpus/` -- are they for testing? Would customers use them?

benchmarking/nursery/test_num_gpus/requirements-synetune.txt
mseeger: `requirements.txt` is combined from this file (for Syne Tune) and `requirements.txt` for the training script.

benchmarking/nursery/test_num_gpus/launch_remote.py
mseeger: We could remove these. I tend to leave example code in `benchmark/nursery`, essentially the code I am using to test this. Maybe this is not good practice.

docs/source/tutorials/benchmarking/bm_local.rst
```diff
@@ -80,6 +80,13 @@ GPU with PyTorch being installed):
   `here <../multifidelity/mf_async_model.html#controlling-mobster-computations>`__
   and
   `here <../basics/basics_bayesopt.html#speeding-up-decision-making>`__.
+* ``num_gpus_per_trial``: If you run on an instance with more than one GPU,
+  you can prescribe how many GPUs should be allocated to each trial. The default
+  is 1. Note that if the product of ``n_workers`` and ``num_gpus_per_trial`` is
+  larger than the number of GPUs on the instance, trials will be delayed.
+* ``delete_checkpoints``: If 1, checkpoints of trials are removed whenever they
+  are not needed anymore. The default is 0, in that all checkpoints are
+  retained.
```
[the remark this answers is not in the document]
mseeger: No. By default, all checkpoints are retrained. If this flag is set, we remove checkpoints which are for sure not needed anymore. For speculative CP removal, you need to pass additional parameters.

The thing about `delete_checkpoints` is that it is an argument of the backend (not the scheduler), and the benchmarking formalism does not allow to customize arguments to the backend. There are only few backends, and all arguments relevant for them should be available anyway.

benchmarking/nursery/test_num_gpus/launch_remote.py
mseeger: This is touching a more general problem. So far, we keep examples in `examples/`, which is independent of `benchmarking`. But anything involving remote launching or SageMaker backend is hard to put together from scratch, and it is much simpler to use `benchmarking`.

Also, `benchmarking/nursery` is getting misused. We should split it into (a) curated examples, and (b) the original nursery, where anything goes.

benchmarking/commons/baselines.py
```diff
@@ -51,6 +51,8 @@ class MethodArguments:
         arguments
     :param restrict_configurations: Only for simulator backend. If given, the
         scheduler is restricted to suggest configs from this list only
+    :param num_gpus_per_trial: Only for local backend and GPU training. Number
+        of GPUs assigned to a trial. Defaults to 1
```
wesk, line 55: Would there be any downsides if we instead default to the maximum available?

benchmarking/nursery/test_num_gpus/launch_remote.py
wesk: > But anything involving remote launching or SageMaker backend is hard to put together from scratch, and it is much simpler to use benchmarking.

Interesting! But we have SageMaker examples, for example `examples/launch_height_sagemaker.py`. Do we have any remote launching examples? (I see `examples/launch_height_sagemaker_remotely.py` but that's both sagemaker and remote, what about just remote?)

What makes it simpler to use `benchmarking`? (Maybe out-of-scope for this PR, but interesting for me to follow-up on separately.)

> Also, benchmarking/nursery is getting misused. We should split it into (a) curated examples, and (b) the original nursery, where anything goes.

In what way(s) is it being misused? (probably also out-of-scope for this PR).

mseeger: If you look at these examples, they need to do lots of "plumbing" by hand. Exactly what the benchmarking formalism saves you doing.

mseeger: The original idea of `benchmarking/nursery` was to use it for experimental setups. In that sense, `benchmarking/examples` would be where such a setup could be "graduated".

benchmarking/commons/baselines.py
mseeger: Yes! There is a trade-off between this and `n_workers`. If you use all GPUs for one trial, you can only have 1 worker.

In fact, I myself rather set this here to 1 and use the GPUs for different workers. This is the default.

wesk (approved):

PR #637, opened 2023-04-20
refactor: Changes to benchmarking formalism

mseeger:
This is mainly renaming `map_extra_args` to `map_method_args`, and also makes sure this mapping is executed just before the creation of the scheduler, which allows the mapping function to modify all arguments (for example, also the config space).

----

### DIFF
--- a/benchmarking/commons/hpo_main_local.py
+++ b/benchmarking/commons/hpo_main_local.py
@@ -134,9 +134,6 @@ def create_objects_for_tuner(
             },
         }
 
-    if map_method_args is not None:
-        method_kwargs = map_method_args(configuration, method, method_kwargs)
-
     method_kwargs.update(
         dict(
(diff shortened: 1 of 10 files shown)


benchmarking/commons/hpo_main_local.py
```diff
@@ -277,25 +276,27 @@ def main(
     methods: MethodDefinitions,
     benchmark_definitions: RealBenchmarkDefinitions,
     extra_args: Optional[ExtraArgsType] = None,
-    map_extra_args: Optional[MapMethodArgsType] = None,
+    map_method_args: Optional[MapMethodArgsType] = None,
```
mseeger, line 279: The renaming is fixing something you started

jgolebiowski (approved):
LGTM, thanks for the change.

PR #640, opened 2023-04-28
refactor: Simplified Bayesian optimization code

mseeger:
Number of simplifications of Bayesian optimization code:

* Removed profiling code
* BayesianOptimizationSearcher works with all surrogate models, not just GP
* ExclusionList more generally useful
* Moving code to where it belongs, delete some files

----


syne_tune/optimizer/schedulers/searchers/model_based_searcher.py
```diff
@@ -487,3 +479,331 @@ def _assign_random_searcher(self):
...
+    else:
+        acquisition_class, acquisition_kwargs = unwrap_acquisition_class_and_kwargs(
+            acquisition_class
+        )
+        return acquisition_class(
+            model, active_metric=active_metric, **acquisition_kwargs
+        )
+
+
+class BayesianOptimizationSearcher(ModelBasedSearcher):
```
jgolebiowski, line 505: Its is kind of hard to compare line-by-line. Could you summarize what changed between this implementation and the one we have in GPFIFOSeacher? It seems that its mostly the same but I want to make sure I havent missed anything.

jgolebiowski (commented):
Looks like a very good change, added one clarifying question but it should be considered good to go.

syne_tune/optimizer/schedulers/searchers/model_based_searcher.py
mseeger: All code was moved to `BayesianOptimizationSearcher`, except for the constructor and the method to restore from inner state.

Construction of our complex searchers is always done by a factory, because there is a lot of code (and complexity) which is shared.

If you come up with a new one, we also need to make sure the construction is easy for the user.

jgolebiowski (approved):
LGTM

PR #641, opened 2023-04-28
feat: add example to optimize hyperparams of HF models on SWAG

aaronkl:
Adds an example how to optimize hyperparameters for fine-tuning transformer models from the HuggingFace hub on the SWAG Q/A dataset. The example also shows how to visualize the optimization process. Future PR will extend this example to include more sophisticated techniques (e.g checkpointing) to accelerate the HPO process.

----

### DIFF
--- a/benchmarking/nursery/fine_tuning_hugging_face/run_hpo.py
+++ b/benchmarking/nursery/fine_tuning_hugging_face/run_hpo.py
@@ -11,12 +11,13 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 import logging
+import matplotlib.pyplot as plt
 
 from sagemaker.huggingface import HuggingFace
 
 from syne_tune import Tuner, StoppingCriterion
 from syne_tune.backend import LocalBackend, SageMakerBackend
-from syne_tune.config_space import randint, loguniform, uniform
+from syne_tune.config_space import loguniform, uniform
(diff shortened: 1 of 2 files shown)


mseeger:
Given this goes to `nursery`, can keep it this way.

It would, however, be more useful if the training code went to `benchmarking/training_scripts/`, and the default config space definition to `benchmarking/commons/benchmark_definitions/`, and also included in `real_benchmark_definitions.py` there. This allows people to run it locally/remotely, with local/SM backend.

I can also do that later.

mseeger:
There is another example in `fine_tuning_transformer_glue`, which we (I?) also need to convert into a benchmark. I put that on my list.

mseeger:
I could convert both of them in one go.

mseeger:
@aaronkl Would you be up for writing a tutorial, something like "How to tune Hugging Face models with Syne Tune"?

benchmarking/nursery/fine_tuning_hugging_face/run_hpo.py
```diff
@@ -0,0 +1,137 @@
...
+)
+tuner.run()
+
+
+experiment = load_experiment(tuner.name)
+results = experiment.results
+metric_name = experiment.metadata["metric_names"][0]
+
+best_config = experiment.best_config()
+
```
mseeger, line 102: Comment on what this plot shows?

PR #644, opened 2023-05-02
fix: Random seed initialisation in scheduler_searcher.py.

sighellan:
Similar to #608 .

int32 is limited to 2**31-1. On windows this was resulting in a program crash when running `scheduler = BORE(conf_space, metric, random_seed=None)`. The problem did not occur when specifying the `random_seed`.

```
File c:\[...]\syne_tune\optimizer\schedulers\scheduler_searcher.py:45, in TrialSchedulerWithSearcher.__init__(self, config_space, **kwargs)
     43 random_seed = kwargs.get("random_seed")
     44 if random_seed is None:
---> 45     random_seed = np.random.randint(0, 2**32)
     46 logger.info(f"Master random_seed = {random_seed}")
     47 self.random_seed_generator = RandomSeedGenerator(random_seed)

File mtrand.pyx:746, in numpy.random.mtrand.RandomState.randint()

File _bounded_integers.pyx:1336, in numpy.random._bounded_integers._rand_int32()

ValueError: high is out of bounds for int32
```

----

### DIFF
--- a/syne_tune/optimizer/schedulers/scheduler_searcher.py
+++ b/syne_tune/optimizer/schedulers/scheduler_searcher.py
@@ -42,7 +42,7 @@ class TrialSchedulerWithSearcher(TrialScheduler):
         # Generator for random seeds
         random_seed = kwargs.get("random_seed")
         if random_seed is None:
-            random_seed = np.random.randint(0, 2**32)
+            random_seed = np.random.randint(0, 2**31 - 1)
         logger.info(f"Master random_seed = {random_seed}")
         self.random_seed_generator = RandomSeedGenerator(random_seed)
 


syne_tune/optimizer/schedulers/scheduler_searcher.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, config_space: Dict[str, Any], **kwargs):
         # Generator for random seeds
         random_seed = kwargs.get("random_seed")
         if random_seed is None:
-            random_seed = np.random.randint(0, 2**32)
+            random_seed = np.random.randint(0, 2**31 - 1)
```
mseeger, line 45: Thanks!

I'd actually prefer to have a function like this in `syne_tune/optimizer/random_seeds.py`:

```
def generate_random_seed(random_state=np.random) -> int:
   return random_state.randint(0, 2**31 - 1)
```

Then, use this to implement both `RandomSeedGenerator.__call__` and sample the master random seed elsewhere.

mseeger: Then, we have the logic in one place.

mseeger (changes requested):

syne_tune/optimizer/schedulers/scheduler_searcher.py
mseeger: I've also submitted #645 with some further fixes along these lines.

mseeger:
The CI is failing, because the commit message should start with "fix:". It currently starts with "Fix:"

wesk:
Validation is failing because your commit title does not have the `fix: ` prefix. The way the validation currently behaves is that at least one commit must have a valid ConventionalCommits prefix (in addition to the PR title). To fix this, you can either 1) push a new commit with a valid prefix to your branch, or 2) amend and force-push the commit to your branch. 

Or, option 3), if we merge https://github.com/awslabs/syne-tune/pull/646 first, then we should be able to merge your PR as-is

mseeger:
@sighellan I could also do the fix in #646 and we close this one. Up to you.

mseeger:
Thanks @sighellan for this catch! I am providing a fix in #645 , which also cleans things up a bit, so this won't happen again.

sighellan:
Sounds good! I agree that #645  is a neater fix.

PR #645, opened 2023-05-02
Fix: Fixing limits for random seed

mseeger:
Further fix about limits for random seed

----


mseeger:
Relates to #644

mseeger:
OK, this is now containing the fix from #644 , and cleans things up so this does not happen again

wesk (approved):

PR #647, opened 2023-05-03
fix: pin fastparquet version

geoalgo:
Closes #642 by setting fastparquet version. The library was updated and is not compatible anymore with LCBench dataset. If we want to update to more recent version, we could update the processing done to lcbench.

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #648, opened 2023-05-03
chore: run two additional examples as tests

wesk:
Related to https://github.com/awslabs/syne-tune/pull/647. Runs two missing examples as tests.

----

### DIFF
--- a/.github/workflows/end-to-end-tests.yml
+++ b/.github/workflows/end-to-end-tests.yml
@@ -158,6 +158,14 @@ jobs:
       requires-aws-credentials: true
     secrets: inherit
 
+  launch_lcbench_simulated:
+    needs: stop_all_training_jobs
+    uses: ./.github/workflows/run-syne-tune.yml
+    with:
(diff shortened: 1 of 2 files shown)


.github/workflows/end-to-end-tests.yml
```diff
@@ -158,6 +158,14 @@ jobs:
       requires-aws-credentials: true
     secrets: inherit
 
+  launch_lcbench_simulated:
```
geoalgo, line 161: I wonder if we could just run all examples by default, and just specific the ones we want to run on sagemaker.
Also, one thing that seems clear in addition to adding examples potentially manually is that we may want to trim them, right now there are 35 examples which is probably too much.

geoalgo (approved):

.github/workflows/end-to-end-tests.yml
wesk: This is a good suggestion. Some refactoring/cleanup will be required in order to automatically identify which examples need sagemaker versus which do not, so planning to do this in a separate PR.

PR #652, opened 2023-05-05
feat: Plotting tools for result of comparative study

mseeger:
This is the code for plotting comparative results.

* Creating reverse index by scanning metadata files
* Load results and aggregate curves over seeds, and different ways
* Comparative plots
* Learning curve plots
* Update of documentation; example how to use this

----


benchmarking/nursery/benchmark_hypertune/plot_results.py
```diff
@@ -0,0 +1,95 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mseeger, line 1: Just a use case example

docs/source/tutorials/benchmarking/bm_plotting.rst
```diff
@@ -0,0 +1,171 @@
+Visualization of Results
```
mseeger, line 1: Just docs

syne_tune/experiments/aggregate_results.py
```diff
@@ -0,0 +1,151 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mseeger, line 1: Code for aggregating results over seeds, as curves over time

syne_tune/experiments/plotting.py
```diff
@@ -0,0 +1,664 @@
...
+    This class loads, processes, and plots results of a comparative study,
+    combining several experiments for different methods, seeds, and
+    benchmarks (optional). Note that an experiment corresponds to one run
+    of HPO, resulting in files :const:`~syne_tune.constants.ST_METADATA_FILENAME`
+    for metadata, and :const:`~syne_tune.constants.ST_RESULTS_DATAFRAME_FILENAME`
+    for time-stamped results.
+
+    There is one comparative plot per benchmark (aggregation of results
+    across benchmarks are not supported here). Results are grouped by
+    setup (which usually equates to method), and then summary statistics are
```
aaronkl, line 256: `setup` might be confusing here. Maybe trajectory or just method?

syne_tune/experiments/results_utils.py
```diff
@@ -10,14 +10,364 @@
...
+            benchmark_name = _get_benchmark_name(benchmark_key, metadata, tuner_path)
+            seed_value = None if seed_key is None else int(metadata[seed_key])
+            try:
+                # Extract ``setup_name``
+                map = (
+                    metadata_to_setup[benchmark_name]
+                    if is_map_dict
+                    else metadata_to_setup
+                )
+                setup_name = map(metadata)
```
aaronkl, line 278: I don't understand what setup_name here exactly is? Is that just a identifier of a single run for one method?

syne_tune/experiments/plotting.py
```diff
@@ -0,0 +1,664 @@
...
+                ax.grid(True)
+            if subplot_no in legend_no:
+                ax.legend()
+        plt.show()
+        return fig, axs
+
+    def plot(
+        self,
+        benchmark_name: Optional[str] = None,
+        plot_params: Optional[PlotParameters] = None,
```
aaronkl, line 606: We do we have to pass plot_params here when we already pass it in the __init__ function?

syne_tune/experiments/plotting.py
```diff
@@ -0,0 +1,664 @@
...
+
+    * Less than ``num_runs`` experiments found. Experiments failed, or files
+      were not properly synced.
+    * More than ``num_runs`` experiments found. This happens if initial
+      experiments for the study failed, but ended up writing results. This can
+      be fixed by either removing the result files, or by using
+      ``datetime_bounds`` (since initial failed experiments ran first).
+
+    Result files have the path
+    ``f"{experiment_path()}{ename}/{patt}/{ename}-*/"``, where ``path`` is from
```
aaronkl, line 284: This assumes that all runs of one method are in some subfolder and the runs of another method in another subfolder, correct?

benchmarking/examples/benchmark_hypertune/plot_results.py
```diff
@@ -0,0 +1,95 @@
...
+    # First: nas201-cifar100
+    benchmark_name = "nas201-cifar100"
+    benchmark = benchmark_definitions[benchmark_name]
+    # These parameters overwrite those given at construction
+    plot_params = PlotParameters(
+        metric=benchmark.metric,
+        mode=benchmark.mode,
+        ylim=(0.265, 0.31),
+    )
+    results.plot(
```
aaronkl, line 77: Can I access the figure if I want to change some details in the plot?

syne_tune/experiments/plotting.py
```diff
@@ -0,0 +1,664 @@
...
+        metadata_subplot_level: bool = False,
+        download_from_s3: bool = False,
+        s3_bucket: Optional[str] = None,
+    ):
+        if download_from_s3:
+            assert (
+                with_subdirs is not None
+            ), "Cannot download files from S3 if with_subdirs=None"
+            download_result_files_from_s3(experiment_names, s3_bucket)
+        result = create_index_for_result_files(
```
aaronkl, line 353: How expensive is it to built this index? Would it make sense to dump this object, such that I can reload it when I run my plotting script multiple times, for example, because I adjust the parameters of plotting?

aaronkl (changes requested):

syne_tune/experiments/plotting.py
```diff
@@ -0,0 +1,664 @@
...
+
+DEFAULT_XLABEL = "wall-clock time (secs)"
+
+
+@dataclass
+class PlotParameters:
+    """
+    Parameters specifying the figure.
+
+    In plots, lower is better. An original metric value ``metric_val`` is
```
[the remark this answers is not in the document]
mseeger: OK, I make this optional. Personally I find it confusing when in the same paper, for some benchmarks, larger is better, for others, smaller is better.

syne_tune/experiments/plotting.py
```diff
@@ -0,0 +1,664 @@
...
+                "xlim",
+                "ylim",
+                "metric_multiplier",
+                "tick_params",
+                "aggregate_mode",
+                "dpi",
+                "grid",
+            ],
+        )
+        default_values = [
```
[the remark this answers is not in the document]
mseeger: This is because of this merging. You can specify *some* arguments when calling `plot`, the remaining ones should be taken from the default params passed at construction of `ComparativeResults`. If I use default values here, I cannot check whether you set something or not.

syne_tune/experiments/results_utils.py
mseeger: A setup is a bit abstract, it translates to essentially a curve of a certain color. The curves in the same plots are for different setups, but the same setup can appear in different subplots.

Often, setup equates to "HPO method", but sometimes it does not. You can also use plotting, running the same method on different variations (e.g., n_workers; config space), and then a setup is not a method.

One example would be the plots you did, where you compare ASHA with 2, 4, 8, 16 workers. The setup would be n_workers then.

syne_tune/experiments/plotting.py
mseeger: You do not, but you can. You can specify benchmark-dependent params here, in particular `ylim` and `metric`.

syne_tune/experiments/plotting.py
mseeger: Depends a bit. If a method is cheap to run (e.g., ASHA in simulation), then results land in the same folder for all seeds. For expensive ones, you have separate subfolders per seed.

benchmarking/examples/benchmark_hypertune/plot_results.py
mseeger: How would this work? Is there a way to export the code and data for a plot?

syne_tune/experiments/plotting.py
mseeger: Good point. In my experience so far, this is quite cheap. But we can also support storing it.

syne_tune/experiments/plotting.py
mseeger: Method would not be good.

benchmarking/examples/benchmark_hypertune/plot_results.py
aaronkl: well you could just return the figure handle no?

aaronkl (approved):

PR #660, opened 2023-05-08
feat: implementation of NSGA-2

aaronkl:
Wrapper for the pymoo implementation of NSGA-2, an evolutionary algorithm for multi-objective optimization.

----


syne_tune/optimizer/baselines.py
```diff
@@ -778,6 +779,50 @@ def __init__(
         )
 
 
+class NSGA2(FIFOScheduler):
```
geoalgo, line 782: Could you add tests?

syne_tune/optimizer/baselines.py
```diff
@@ -778,6 +779,50 @@ def __init__(
...
+    """
+
+    See :class:`~syne_tune.optimizer.schedulers.searchers.RandomSearcher`
+    for ``kwargs["search_options"]`` parameters.
+
+    :param config_space: Configuration space for evaluation function
+    :param metric: Name of metric to optimize
+    :param population_size: See
```
geoalgo, line 790: Can you add a new comment here to describe the population? We probably dont want to cross reference in this case as those are different methods which my also have different defaults. 

Also a minor point: I do not think we need to put default in the comment (it adds a risk to have them async with the code).

syne_tune/optimizer/baselines.py
```diff
@@ -778,6 +779,50 @@ def __init__(
...
+        :class:`~syne_tune.optimizer.schedulers.searchers.RegularizedEvolution`.
+        Defaults to 100
+    :param sample_size: See
```
geoalgo, line 793: Is it used?

syne_tune/optimizer/baselines.py
```diff
@@ -778,6 +779,50 @@ def __init__(
...
+        :class:`~syne_tune.optimizer.schedulers.searchers.RegularizedEvolution`.
+        Defaults to 10
+    :param random_seed: Random seed, optional
```
geoalgo, line 796: The comment can be removed, it does not add any information.

syne_tune/optimizer/baselines.py
```diff
@@ -778,6 +779,50 @@ def __init__(
...
+    :param kwargs: Additional arguments to
+        :class:`~syne_tune.optimizer.schedulers.FIFOScheduler`
+    """
+
+    def __init__(
+        self,
+        config_space: Dict[str, Any],
+        metric: List[str],
+        mode: Union[List[str], str] = "min",
+        population_size: int = 100,
```
geoalgo, line 806: Regarding this default, the value seems high to me. Do you think a smaller value like 20 would work better on "average"? (for instance if we were to run in our benchmarks I suppose 20 might be better?)

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
```diff
@@ -0,0 +1,181 @@
...
+)
+from pymoo.core.variable import Real, Choice
+from pymoo.core.variable import Integer as PyMOOInteger
+
+
+class MultiObjectiveMixedVariableProblem(Problem):
+    def __init__(self, n_obj, config_space, **kwargs):
+        vars = {}
+
+        for hp_name, hp in config_space.items():
```
geoalgo, line 37: We should probably assert the type if we support only a subset of all options.

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
```diff
@@ -0,0 +1,181 @@
...
+    MixedVariableMating,
+    MixedVariableSampling,
+    MixedVariableDuplicateElimination,
+)
+from pymoo.core.variable import Real, Choice
+from pymoo.core.variable import Integer as PyMOOInteger
+
+
+class MultiObjectiveMixedVariableProblem(Problem):
+    def __init__(self, n_obj, config_space, **kwargs):
```
geoalgo, line 34: Can you add type hints?

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
```diff
@@ -0,0 +1,181 @@
...
+                if (
+                    not isinstance(hp, Categorical)
+                    and not isinstance(hp, Integer)
+                    and not isinstance(hp, Float)
+                    and not isinstance(hp, FiniteRange)
+                ):
+                    raise Exception(
+                        f"Type {type(hp)} for hyperparameter {hp_name} "
+                        f"is not support for NSGA-2."
+                    )
```
geoalgo, line 109:
```suggestion
                assert type(hp) in [Categorical, Integer, Float, FiniteRange], f"Type {type(hp)} for hyperparameter {hp_name} not supported."
```

syne_tune/optimizer/baselines.py
aaronkl: Good question. I followed pymoo's default, but it indeed might be a bit high for our benchmarks. I'll set it to 20

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
```diff
@@ -0,0 +1,177 @@
...
+from pymoo.algorithms.moo.nsga2 import NSGA2
+from pymoo.core.problem import Problem
+from pymoo.core.evaluator import Evaluator
+from pymoo.problems.static import StaticProblem
+from pymoo.core.mixed import (
+    MixedVariableMating,
+    MixedVariableSampling,
+    MixedVariableDuplicateElimination,
+)
+from pymoo.core.variable import Real, Choice
```
mseeger: Map them to `PyMOOReal`, `PyMOOChoice`, or to `PyMOOFloat`, `PyMOOCategorical`.

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
```diff
@@ -0,0 +1,177 @@
...
+
+class MultiObjectiveMixedVariableProblem(Problem):
+    def __init__(self, n_obj: int, config_space: Dict[str, Any], **kwargs):
+        vars = {}
+
+        for hp_name, hp in config_space.items():
+            if isinstance(hp, Categorical):
+                vars[hp_name] = Choice(options=hp.categories)
+            elif isinstance(hp, Integer):
+                vars[hp_name] = PyMOOInteger(bounds=(hp.lower, hp.upper))
```
mseeger, line 46: I think this should be `hp.upper - 1`, no? For our `Integer` type, the upper bound is inclusive.

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
```diff
@@ -0,0 +1,177 @@
...
+            elif isinstance(hp, FiniteRange):
+                vars[hp_name] = PyMOOInteger(bounds=(0, hp.size - 1))
+            elif isinstance(hp, Float):
+                vars[hp_name] = Real(bounds=(hp.lower, hp.upper))
+
```
mseeger, line 51: We have more types. Best to throw exception here.

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
```diff
@@ -0,0 +1,177 @@
...
+        super().__init__(vars=vars, n_obj=n_obj, n_ieq_constr=0, **kwargs)
+
+
+class NSGA2Searcher(StochasticSearcher):
+    """
+    This is a wrapper around the NSGA-2 [1] implementation of pymoo [2].
+
+    [1] K. Deb, A. Pratap, S. Agarwal, and T. Meyarivan.
```
mseeger, line 59: Prefix these lines with "|".

See `syne_tune/optimizer/schedulers/searchers/bayesopt/models/meanstd_acqfunc_impl.py` for example

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
```diff
@@ -0,0 +1,177 @@
...
+        ("max"). In the case of multi-objective optimization, mode can be a list defining for
+        each metric if it is minimized or maximized
+    :param population_size: Size of the population
+    """
+
+    def __init__(
+        self,
+        config_space: Dict[str, Any],
+        metric: List[str],
+        mode: Union[List[str], str],
```
mseeger, line 88: Defaults to "min"

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
```diff
@@ -0,0 +1,177 @@
...
+                    Float,
+                    FiniteRange,
+                ], f"Type {type(hp)} for hyperparameter {hp_name} not supported."
+
+        self.problem = MultiObjectiveMixedVariableProblem(
+            config_space=config_space, n_var=len(self.hp_names), n_obj=len(metric)
+        )
+        self.algorithm = NSGA2(
+            pop_size=population_size,
+            sampling=MixedVariableSampling(),
```
mseeger, line 117: Is there any chance we can control the random seed of these components? If so, please try to do that, and use `self.random_seed`.

Otherwise, we lose the feature that results are repeatable, which we should retain for methods that just do random sampling.

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def _update(self, trial_id: str, config: dict, result: dict):
+        observed_metrics = {}
+        for mode, metric in zip(self._mode, self._metric):
+            value = result[metric]
+            if mode == "max":
+                value *= -1
+            observed_metrics[metric] = value
+
+        self.observed_values.append(list(observed_metrics.values()))
```
mseeger, line 139: You sure you don't want to retain the dict?

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
```diff
@@ -0,0 +1,177 @@
...
+        :return: New configuration. The searcher may return None if a new
+            configuration cannot be suggested. In this case, the tuning will
+            stop. This happens if searchers never suggest the same config more
+            than once, and all configs in the (finite) search space are
+            exhausted.
+        """
+
+        if self.current_individual >= len(self.current_population):
+            raise Exception(
+                "It seems that some configurations are sill pending, while querying a new configuration."
```
mseeger: This is a bit sad. Isn't there an easy way to make this work?

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
```diff
@@ -0,0 +1,177 @@
...
+        )
+        self.algorithm.setup(
+            problem=self.problem, termination=("n_eval", 2**32 - 1), verbose=False
+        )
+
+        self.current_population = self.algorithm.ask()
+        self.current_individual = 0
+        self.observed_values = []
+
+    def _update(self, trial_id: str, config: dict, result: dict):
```
mseeger, line 131: You don't use `trial_id` or `config` here, this is a bit odd. You cannot just collect metrics, you also need to collect from which trial they come.

In case you use `config`, be careful that it needs to be transformed for `FiniteRange`.

mseeger (changes requested):

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
aaronkl: We have to pass a numpy array to pymoo

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
aaronkl: I guess the best option would be to have a synchronous FIFO scheduler. We could implement some hack in there, but that would lead to a different algorithm and bias the results

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
aaronkl: Since we only support the num_workers = 1 case, trials will come in sequentially.

aaronkl: Actually, let me change this to a dict. This should also work then, when we have a synchronous fifo scheduler.

mseeger (approved):

PR #662, opened 2023-05-11
feat: Allow to skip initial jobs in remote launching

mseeger:
Simple change to allow a launch command to be re-run if it failed after having launched a number of jobs already.
Together with Jacek's forthcoming PR, this will improve remote launching for customers.

----

### DIFF
--- a/benchmarking/commons/launch_remote_local.py
+++ b/benchmarking/commons/launch_remote_local.py
@@ -32,7 +32,10 @@ from benchmarking.commons.hpo_main_local import (
     get_benchmark,
     LOCAL_BACKEND_EXTRA_PARAMETERS,
 )
-from benchmarking.commons.launch_remote_common import sagemaker_estimator_args
+from benchmarking.commons.launch_remote_common import (
+    sagemaker_estimator_args,
+    append_extra_arguments_launch_remote,
+)
 from benchmarking.commons.utils import (
     filter_none,
     message_sync_from_s3,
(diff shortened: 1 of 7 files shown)


benchmarking/commons/launch_remote_local.py
```diff
@@ -132,7 +135,10 @@ def launch_remote(
         command line arguments
     :param extra_args: Extra arguments for command line parser, optional
     """
-    configuration = config_from_argparse(extra_args, LOCAL_BACKEND_EXTRA_PARAMETERS)
+    configuration = config_from_argparse(
+        extra_args=append_extra_arguments_launch_remote(extra_args),
+        benchmark_specific_args=LOCAL_BACKEND_EXTRA_PARAMETERS,
+    )
```
jgolebiowski, line 141: It appears that append_extra_arguments_launch_remote is added every time this function is called. If that is the case we should instead include it into LOCAL_BACKEND_EXTRA_PARAMETERS and other appropriate constants used throughout benchmarking.

If this is an argument for all benchmarking jobs, we should instead add it to `ConfigDict.__base_cl_parameters`

jgolebiowski (changes requested):

benchmarking/commons/launch_remote_local.py
mseeger: No careful. This is an argument only for `launch_remote`, but not for `hpo_main`. Typically, these two have the same arguments, so it is easy to move from one to the other.

jgolebiowski: I meant that we should create a new constant `REMOTE_LAUNCHING_EXTRA_PARAMETERS` and define it as 
```
REMOTE_LAUNCHING_EXTRA_PARAMETERS = LOCAL_BACKEND_EXTRA_PARAMETERS + <new param goes here>
```

And do the same in other places it is appropriate. All extra parameter are defined in constants like this so having this one defined in a function can be really confusing since it is not fundamentally different from others. I can see how my original explanation was too cryptic, sorry about that.

jgolebiowski (dismissed):
LGTM

jgolebiowski (approved):
🚢 it

PR #663, opened 2023-05-11
fix: Fix import bug, and also new workflow to install Syne Tune with …

mseeger:
…gpsearchers dependencies only

* Fixes import bug which precludes running Syne Tune with restricted dependencies
* Introduce new workflow `run-syne-tune-gpsearchers-only.yml`, and make sure that `examples/launch_standalone_bayesian_optimization.py` runs with this one (would have caught the bug)

----


mseeger:
I'll check which of the examples should also run with minimal dependencies, and change the workflow there as well. Also makes the CI run faster.

.github/workflows/run-syne-tune-gpsearchers-only.yml
wesk, line 1: The `.github/workflows/run-syne-tune.yml` is a re-usable workflow, and you can pass in custom `inputs` at the top of the file. 

Rather than creating this new re-usable workflow, I'd recommend creating a new option in the original file which specifies the mode that you want to set.

Something perhaps along the lines of:
```
on:
  workflow_call:
    inputs:
      gpsearchers-only:
        required: false
        type: boolean
```

And then update the logic to branch based on that flag.

wesk (changes requested):
Requesting re-use of the existing `run-syne-tune.yml` workflow

.github/workflows/run-syne-tune.yml
```diff
@@ -19,6 +19,9 @@ on:
       timeout-minutes:
         required: false
         type: number
+      extras-require:
+        required: false
+        type: string
```
wesk, line 24: Nice, thanks!

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
```diff
@@ -36,26 +36,35 @@
...
+                        vars[hp_name] = PyMOOChoice(options=hp.categories)
+                    elif isinstance(hp, Integer):
+                        vars[hp_name] = PyMOOInteger(bounds=(hp.lower, hp.upper - 1))
+                    elif isinstance(hp, FiniteRange):
+                        vars[hp_name] = PyMOOInteger(bounds=(0, hp.size - 1))
+                    elif isinstance(hp, Float):
+                        vars[hp_name] = PyMOOReal(bounds=(hp.lower, hp.upper))
+                    else:
+                        raise Exception(
+                            f"Type {type(hp)} of hyperparameter {hp_name} is not supported!"
```
wesk, line 60: The current change of nesting this in a class makes sense to fix the import issue.

But all these `isinstance()` checks added in https://github.com/awslabs/syne-tune/pull/660 feel somehow brittle to me, swapping in all these PyMOO datatypes here. I wonder if there's a cleaner way to use MOO, or if using MOO necessarily introduces this complexity. 

Why does PyMOO even create those special pymoo datatypes?

wesk (approved):

wesk:
Thanks for improving the testing as well to use narrower imports, this is really good!

syne_tune/optimizer/schedulers/multiobjective/nsga2_searcher.py
mseeger: Well, fair enough, we also create our own datatypes!

mseeger: I think this is fine. As long as we remember to keep this stuff in functions, so it is not imported when not really needed. And the new integ tests will make sure this does not happen

PR #664, opened 2023-05-11
chore: re-introduce check-standalone-bayesian-optimization.yml

wesk:
Re-add `check-standalone-bayesian-optimization.yml` workflow

----

### DIFF
--- a/.github/workflows/check-standalone-bayesian-optimization.yml
+++ b/.github/workflows/check-standalone-bayesian-optimization.yml
@@ -21,6 +21,6 @@ jobs:
       run: |
         python -m pip install -r requirements.txt --upgrade pip
         python -m pip install -r requirements-gpsearchers.txt
-    - name: Check that stand-alone bayesian-optimation example runs with only core and gpsearchers dependencies
+    - name: Check that stand-alone bayesian-optimization example runs with only core and gpsearchers dependencies
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/check-standalone-bayesian-optimization.yml
```diff
@@ -0,0 +1,26 @@
+name: Check instantiation of bayesian optimization on python 3.8
```
mseeger, line 1: Then you can remove it from `integ-tests.yml`

.github/workflows/check-standalone-bayesian-optimization.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
+        python -m pip install -r requirements-gpsearchers.txt --upgrade pip
```
mseeger, line 22: You also need `requirements.txt`

mseeger (changes requested):

.github/workflows/check-standalone-bayesian-optimization.yml
mseeger: This is what we call `core`, this is always required.

mseeger (approved):

PR #665, opened 2023-05-12
chore: add initial multiobjective benchmarking logic

wesk:
Add initial multiobjective benchmarking logic. This is to support the ongoing development and evaluation of multiobjective methods in Syne Tune. 

This is based on the approach taken in `benchmarking/nursery/benchmark_automl`.

----


benchmarking/nursery/benchmark_multiobjective/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+This folder shows one way to run quick experiments running different scheduler on different benchmarks and plot
+ results once they are done.
+
+To run all experiments, you can run the following:
+
+
+```bash
+pip install -r benchmarking/nursery/benchmark_multiobjective/requirements.txt
+python benchmarking/nursery/benchmark_multiobjective/benchmark_main.py --experiment_tag "my-new-experiment" --num_seeds 2
```
mseeger, line 11: Why not call this `hpo_main.py`?

And you could also add remote launching, so people don't need to worry about requirements, or running things locally.

benchmarking/nursery/benchmark_multiobjective/baselines.py
```diff
@@ -0,0 +1,53 @@
...
+from syne_tune.optimizer.schedulers.searchers import StochasticSearcher, StochasticAndFilterDuplicatesSearcher
+
+NUM_TO_SAMPLE = 1000
+
+
+def MOREABench(method_arguments: MethodArguments, **kwargs):
+    kwargs = _baseline_kwargs(method_arguments, kwargs)
+    for name, space in kwargs["config_space"].items():
+        if isinstance(space, Float) or isinstance(space, Integer):
+            space = Categorical(categories=space.sample(size=NUM_TO_SAMPLE))
```
mseeger, line 34: Why are you doing this? Can the method not deal with types other than `choice`?

If you run this config space into something involving BO, you'll get an enormous input vector, because of 1-hot encoding. It is much better to use the `Ordinal` type, where you need to sort the numbers first.

mseeger: Another problem is that for `Integer`, you may sample duplicates.

mseeger: I'd first check why `MOREA` does not support all types we have. This should not be a problem at all.

benchmarking/nursery/benchmark_multiobjective/baselines.py
```diff
@@ -0,0 +1,53 @@
...
+        kwargs["config_space"][name] = space
+    return MOREA(**kwargs)
+
+
+def LSOBOBench(method_arguments: MethodArguments, **kwargs):
```
mseeger, line 39: It would be better to (a) make `LinearScalarizedScheduler` and `MOREA` available in `optimizer/baselines.py`, and then insert this here into `benchmarking/commons/default_baselines.py`

benchmarking/nursery/benchmark_multiobjective/baselines.py
mseeger: I checked: `MOREA` is in `optimizer/baselines.py`, and it supports all types, so there is no need to change the config space at all.

benchmarking/nursery/benchmark_multiobjective/benchmark_definitions.py
```diff
@@ -0,0 +1,41 @@
...
+    fcnet_benchmark_definitions,
+)
+
+
+def nas201_mo_benchmark(dataset_name):
+    return SurrogateBenchmarkDefinition(
+        max_wallclock_time=6 * 3600,
+        n_workers=4,
+        elapsed_time_attr="metric_elapsed_time",
+        metric=["metric_valid_error", "metric_params"],
```
mseeger, line 24: `metric` and `mode` are not supposed to be lists in this type.

Let us be a bit more careful here, and change some things in the def of `SurrogateBenchmarkDefinition`, for example the types becoming union's, and also add a check to `__post_init__` that the lists are the same size, etc.

benchmarking/nursery/benchmark_multiobjective/requirements.txt
```diff
@@ -0,0 +1,12 @@
+syne-tune[extra]
```
mseeger, line 1: This is not needed here.

benchmarking/nursery/benchmark_multiobjective/run_experiment.py
```diff
@@ -0,0 +1,117 @@
+from typing import Callable
```
mseeger, line 1: OK, this is a V0, but why cannot we use the generic code that is already there?

The only difference I see is that there are >1 `metric` and `mode`. We could extend the generic case to deal with that.

mseeger (changes requested):

benchmarking/nursery/benchmark_multiobjective/baselines.py
jgolebiowski: This is a problem with MOREA, it crashes when `Float` is used. This is a hack around it but indeed should be merged inside MOREA directly, I will look into this.

wesk:
Thanks for the feedback @mseeger, moving this PR back to `draft` status while I incorporate the feedback

benchmarking/nursery/benchmark_multiobjective/run_experiment.py
jgolebiowski: I agree. The plan is to make this v0 work while being explicit (spelling out all the operations) to see what needs to be done for MO benchmarks. Then v1 will refactor this code to use `commons` while making any changes to commons to make it possible.

benchmarking/nursery/benchmark_multiobjective/baselines.py
mseeger: I'd honestly prefer if this gets fixed instead. When looking at the code, there is not reason for it not working with `Float`.

benchmarking/nursery/benchmark_multiobjective/requirements.txt
wesk: Removed

benchmarking/nursery/benchmark_multiobjective/benchmark_definitions.py
wesk: Thanks!

benchmarking/nursery/benchmark_multiobjective/README.md
wesk: Renamed to `hpo_main.py`. Agree remote launching would be useful, can add in a future PR

benchmarking/nursery/benchmark_multiobjective/baselines.py
wesk: Updated, would you verify that it now follows the structure you expect?

benchmarking/nursery/benchmark_multiobjective/hpo_main.py
```diff
@@ -0,0 +1,101 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mseeger, line 1: This is not needed. Have a look at `examples/benchmark_hypertune/hpo_main.py`.

mseeger (changes requested):

benchmarking/nursery/benchmark_multiobjective/baselines.py
wesk: @jgolebiowski created https://github.com/awslabs/syne-tune/issues/671 to track the issue

benchmarking/nursery/benchmark_multiobjective/run_experiment.py
wesk: Updating PR to re-use existing logic in commons

benchmarking/nursery/benchmark_multiobjective/requirements.txt
```diff
@@ -0,0 +1,11 @@
+tqdm
```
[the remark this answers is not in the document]
wesk: Removed

benchmarking/nursery/benchmark_multiobjective/run_experiment.py
```diff
@@ -0,0 +1,116 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
wesk: Updated

benchmarking/commons/default_baselines.py
```diff
@@ -36,6 +37,10 @@
     SyncBOHB as _SyncBOHB,
     DEHB as _DEHB,
     SyncMOBSTER as _SyncMOBSTER,
+    MOREA as _MOREA,
+)
+from syne_tune.optimizer.schedulers.multiobjective.linear_scalarizer import (
```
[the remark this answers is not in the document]
wesk: I will follow-up with you to understand how to best make this change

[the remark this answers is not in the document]
mseeger: it seems that `LinearScalarizedScheduler` can be created with the same default args as `BayesianOptimization`, so there should be an entry in `baselines`.

But that can be done later.

benchmarking/commons/hpo_main_simulator.py
```diff
@@ -305,7 +305,7 @@ def start_benchmark_simulated_backend(
                     datasets=benchmark.datasets,
                 ),
             )
-        search_options = dict(debug_log=configuration.verbose)
+        search_options = dict(debug_log=False)
```
mseeger, line 308: No!

tst/benchmarking/benchmark_commons/test_hpo_main_local.py
```diff
@@ -0,0 +1,81 @@
...
+    MOREABench,
+    LSOBOBench,
+)
+from benchmarking.nursery.benchmark_multiobjective.benchmark_definitions import (
+    nas201_mo_benchmark,
+)
+from benchmarking.nursery.benchmark_multiobjective.hpo_main import main
+
+
+class HPOMainLocalTests(unittest.TestCase):
```
mseeger, line 28: This seems a good effort. We should take this as a start for testing benchmark examples, and more general for integration tests!

mseeger (changes requested):
Sorry, but these `verbose=False` should really be removed.

benchmarking/commons/hpo_main_simulator.py
wesk: Good catch, updated

benchmarking/nursery/benchmark_multiobjective/hpo_main.py
```diff
@@ -0,0 +1,21 @@
...
+#
```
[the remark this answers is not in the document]
wesk: Added it!

mseeger:
I don't see the recent changes

wesk:
> I don't see the recent changes

Me neither, that's weird. I force-pushed, but I'm still not seeing it show up. I will close this PR and move to https://github.com/awslabs/syne-tune/pull/674 instead

PR #666, opened 2023-05-12
refactor: Allow benchmark to be multi-objective

mseeger:
----


jgolebiowski (approved):
Thats a very nice change, definitwly helps with requesting multiple metrics from a benchmark!

wesk (approved):

PR #667, opened 2023-05-14
fix: Bugfix in benchmarking

mseeger:
Replace recursive glob by normal glob

----

### DIFF
--- a/benchmarking/commons/utils.py
+++ b/benchmarking/commons/utils.py
@@ -74,7 +74,7 @@ def find_or_create_requirements_txt(
 ) -> Path:
     if requirements_fname is None:
         requirements_fname = "requirements.txt"
-    files = list(entry_point.parent.rglob("requirements*.txt"))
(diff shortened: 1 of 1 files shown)


benchmarking/commons/utils.py
```diff
@@ -74,7 +74,7 @@ def find_or_create_requirements_txt(
 ) -> Path:
     if requirements_fname is None:
         requirements_fname = "requirements.txt"
-    files = list(entry_point.parent.rglob("requirements*.txt"))
+    files = list(entry_point.parent.glob("requirements*.txt"))
```
wesk, line 77: What was the issue that this fixes?

mseeger: This is supposed to find a file "requirements*.txt" in the current path, but not in subdirectories.

mseeger: In general, I was using `rglob` at several places when I should have used `glob`. This is fixed

wesk (approved):

PR #669, opened 2023-05-15
feat: New surrogate models API

jgolebiowski:
Implementation of the new surrogate models API as discussed in https://quip-amazon.com/ndjCA2vLNLB4/Surrogate-models-in-Syne-Tune

----


syne_tune/optimizer/schedulers/searchers/bayesopt/contributed/estimator.py
```diff
@@ -0,0 +1,102 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mseeger, line 1: Can you think of a different name than `contributed`?

This sounds like we discourage contributions on the BO side.

syne_tune/optimizer/schedulers/searchers/bayesopt/models/estimator.py
```diff
@@ -156,14 +156,14 @@ def transform_state_to_data(
     if state.pending_evaluations:
         # In this case, y becomes a matrix, where the observed values are
         # broadcast
-        cand_lst = [
-            hp_ranges.to_ndarray(config) for config in state.pending_configurations()
-        ]
+        cand_lst = []
```
mseeger: This change needs some comment, I am not sure this works out.

syne_tune/optimizer/schedulers/searchers/bayesopt/models/estimator.py
```diff
@@ -174,31 +174,3 @@ def transform_state_to_data(
         targets = np.vstack([targets * np.ones((1, num_fantasy_samples))] + fanta_lst)
         features = np.vstack([features] + cand_lst)
     return TransformedData(features, targets, mean, std)
-
-
-class EstimatorFromTransformedData(Estimator):
```
mseeger, line 179: Why remove this one? I thought this is something we wanted to have

syne_tune/optimizer/schedulers/searchers/bayesopt/contributed/predictor.py
```diff
@@ -0,0 +1,121 @@
...
+        statistics.
+
+        :param input: Single input point :math:`x`, shape ``(d,)``
+        :param head_gradients: See above
+        :return: Gradient :math:`\nabla f(x)`
+        """
+        raise NotImplementedError()
+
+
+class ContributedPredictorWrapper(BasePredictor):
```
mseeger, line 66: I think this one should move to `models/contributed_estimator.py` (except that "contributed" should become something else).

This is like for GP: The wrappers are in `bayesopt/models` (both estimator and predictor in the same wrapper code file), the code is in `bayesopt/gpautograd`

syne_tune/optimizer/schedulers/searchers/bayesopt/contributed/estimator.py
```diff
@@ -0,0 +1,102 @@
...
+    @property
+    def normalize_targets(self) -> bool:
+        """
+        :return: Should targets in ``state`` be normalized before calling
+            :meth:`fit`?
+        """
+        return False
+
+
+class ContributedEstimatorWrapper(Estimator):
```
mseeger, line 58: I think this should also move to `models/contributed_estimator.py`, and why not inherit from `EstimatorFromTransformedData` ?

mseeger (changes requested):

syne_tune/optimizer/schedulers/searchers/bayesopt/models/estimator.py
jgolebiowski: This was commited by mistake, already removed in the new commit.

syne_tune/optimizer/schedulers/searchers/bayesopt/models/estimator.py
jgolebiowski: EstimatorWrapper now implements this function, its a rename + move rather than a removal.

syne_tune/optimizer/schedulers/searchers/bayesopt/contributed/predictor.py
jgolebiowski: Ok makes sense, will move.

mseeger: And note that your code in `bayesopt/contributed` need not adhere to the estimator / predictor split, but you can keep everything in a single class. The split into two is only required one level up, in the wrappers.

That's what I did for the GP code, and also what we discussed.

syne_tune/optimizer/schedulers/searchers/bayesopt/contributed/estimator.py
jgolebiowski: This is just a rename of `EstimatorFromTransformedData`. I wanted to make estimator/predictor consistant with each other.

mseeger:
OK, I think my main comment would be that only the wrapper code (in `bayesopt/models`) needs to follow predictor / estimator split, and not the code in the subdirectory.

syne_tune/optimizer/schedulers/searchers/bayesopt/contributed/predictor.py
jgolebiowski: Thats a good point but I wanted to keep them separated to make it more streamlined. This way we rely on the decoupled view through all the levels and I think it will be easier to follow for newer contributors. I know that I was confused by couplig/decoupling in GP code the first time I saw it.

syne_tune/optimizer/schedulers/searchers/bayesopt/models/sklearn_predictor.py
```diff
@@ -0,0 +1,83 @@
...
+        super().__init__(state, active_metric)
+        self.contributed_predictor = contributed_predictor
+
+    def predict(self, inputs: np.ndarray) -> List[Dict[str, np.ndarray]]:
+        """
+        Returns signals which are statistics of the predictive distribution at
+        input points ``inputs``. By default:
+
+        * "mean": Predictive means.
+        - "std": Predictive stddevs, shape ``(n,)``
```
mseeger, line 48: * instead of -

mseeger: But you could also drop this comment, because it is kind of just what the superclass has.

mseeger (approved):
Just a small comment change.

syne_tune/optimizer/schedulers/searchers/bayesopt/models/sklearn_predictor.py
jgolebiowski: Good point, I wated to change it for both but dropped it. Adjusted this and the base class now.

mseeger (approved):

PR #670, opened 2023-05-15
fix: Fix bug in benchmarking where default values were int instead of…

mseeger:
… bool

----


jgolebiowski (approved):
Nice catch, thanks!

PR #673, opened 2023-05-16
feat: Backoff decorator for jobs scheduling

jgolebiowski:
Simple backoff decorator which can be used for scheduling sagemaker jobs. Once a function is decorated with it, it will run normally but wait on a specific error (e.g. ResourceShareLimitExceededException) and retry after some time. This can be used to leave a scheduling loop running in the background and have it allocate new jobs when instances become available.

----


syne_tune/backend/scheduling.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mseeger, line 1: I think a better place for this would be `syne_tune/remote`, no?

mseeger (commented):
Once this is in, I'd like to integrate it with benchmarking (I can do that)

syne_tune/backend/scheduling.py
jgolebiowski: Good idea, moved.

jgolebiowski:
> Once this is in, I'd like to integrate it with benchmarking (I can do that)

That would be great, if you can do this then its perfect for me. Otherwise we can have a quick chat since I am not yet certain where is the best place to integrate.

PR #676, opened 2023-05-16
feat: Try again in remote launching when ResourceLimitExceeded is caught

mseeger:
I could not directly use your decorater. Maybe there is a better solution.

----

### DIFF
--- a/benchmarking/commons/launch_remote_common.py
+++ b/benchmarking/commons/launch_remote_common.py
@@ -12,6 +12,10 @@
 # permissions and limitations under the License.
 from pathlib import Path
 from typing import Optional, Dict, Any
+import time
+
+from sagemaker.estimator import EstimatorBase
+from botocore.exceptions import ClientError
 
 import benchmarking
 import syne_tune
@@ -67,3 +71,43 @@ def sagemaker_estimator_args(
         if benchmark.estimator_kwargs is not None:
             sm_args.update(benchmark.estimator_kwargs)
     return sm_args
+
+
+ERROR_NAME = "ResourceLimitExceeded"
+
+
+def fit_sagemaker_estimator(
+    backoff_wait_time: int,
+    estimator: EstimatorBase,
+    ntimes_resource_wait: int = 100,
+    **kwargs,
+):
+    """
+    Runs ``estimator.fit(**kwargs)``. If ``backoff_wait_time > 0``, we make sure
+    that if ``fit`` fails with ``ClientError`` of type "ResourceLimitExceeded",
+    we wait for ``backoff_wait_time`` seconds and try again (up to
+    ``ntimes_resource_wait`` times).
+
+    If ``backoff_wait_time <= 0``, the call of ``fit`` is not wrapped.
+
+    :param backoff_wait_time: See above.
+    :param estimator: SageMaker estimator to call ``fit`` for
+    :param ntimes_resource_wait: Maximum number of retries
+    :param kwargs: Arguments for ``estimator.fit``
+    """
+    if backoff_wait_time <= 0:
+        return estimator.fit(**kwargs)
+
+    for _ in range(ntimes_resource_wait):
+        try:
+            return estimator.fit(**kwargs)
+        except ClientError as ex:
+            if ERROR_NAME not in str(ex):
+                raise ex
+
+        print(
+            f"botocore.exceptions.ClientError[{ERROR_NAME}] detected "
+            f"when calling estimator.fit. Waiting "
+            f"{backoff_wait_time / 60:.2f} minutes before retrying"
+        )
+        time.sleep(backoff_wait_time)
--- a/syne_tune/remote/scheduling.py
+++ b/syne_tune/remote/scheduling.py
@@ -14,6 +14,8 @@ import functools
 import logging
 import time
 
+logger = logging.getLogger(__name__)
+
 
 def backoff(errorname: str, ntimes_resource_wait: int = 100, length2sleep: float = 600):
     """
@@ -30,7 +32,7 @@ def backoff(errorname: str, ntimes_resource_wait: int = 100, length2sleep: float
                     if not e.__class__.__name__ == errorname:
                         raise (e)
 
-                logging.info(
+                logger.info(
                     f"{errorname} detected when calling <{some_function.__name__}>, waiting {length2sleep / 60} minutes before retring"
                 )
                 time.sleep(length2sleep)
--- a/tst/benchmarking/benchmark_commons/test_hpo_main_local.py
+++ b/tst/benchmarking/benchmark_commons/test_hpo_main_local.py
@@ -12,6 +12,7 @@
 # permissions and limitations under the License.
 import unittest
 from unittest.mock import MagicMock, patch
+
 from benchmarking.commons.default_baselines import RandomSearch
 from benchmarking.commons.hpo_main_common import ConfigDict
 from benchmarking.nursery.benchmark_multiobjective.baselines import (
@@ -66,7 +67,7 @@ class HPOMainLocalTests(unittest.TestCase):
             config_space
         )
 
-        seeds = [1, 2]
+        seeds = [0, 1]
 
         mock_config_from_argparse.return_value = ConfigDict.from_dict(
             {
--- a/benchmarking/commons/hpo_main_common.py
+++ b/benchmarking/commons/hpo_main_common.py
@@ -131,6 +131,16 @@ class ConfigDict:
             default=False,
             help=f"Internal argument, do not use",
         ),
+        Parameter(
+            name="estimator_fit_backoff_wait_time",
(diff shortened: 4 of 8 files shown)


tst/benchmarking/benchmark_commons/test_hpo_main_local.py
```diff
@@ -25,6 +27,7 @@
 from benchmarking.nursery.benchmark_multiobjective.hpo_main import main
 
 
+@pytest.mark.skip("NEEDS FIXING")
```
wesk, line 30: The test started failing after https://github.com/awslabs/syne-tune/pull/672 was merged.

It's a new test that I just added, I am checking now to see if we can keep it enabled or if we really need to skip it.

benchmarking/commons/launch_remote_common.py
```diff
@@ -75,5 +78,40 @@ def sagemaker_estimator_args(
...
+    @backoff_boto_clienterror(
+        errorname="ResourceLimitExceeded", length2sleep=backoff_wait_time
+    )
+    def fit_sagemaker_estimator_with_backoff(estimator: EstimatorBase, **kwargs):
+        estimator.fit(**kwargs)
+
+    if backoff_wait_time > 0:
+        fit_sagemaker_estimator_with_backoff(estimator, **kwargs)
+    else:
+        estimator.fit(**kwargs)
```
jgolebiowski, line 117: I tihnk this is abusing decorators a little. I would try to bake the logic in directly here or otherwise use something along the lines of:

```
  if backoff_wait_time > 0:
      backoff_boto_clienterror(errorname="ResourceLimitExceeded", length2sleep=backoff_wait_time)(estimator.fit(**kwargs))
  else:
      estimator.fit(**kwargs)
```

jgolebiowski (changes requested):
Small changes, most importantly I think we should combine the two decorators into one

syne_tune/remote/scheduling.py
```diff
@@ -39,3 +43,39 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return errorcatch
+
+
+def backoff_boto_clienterror(
```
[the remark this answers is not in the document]
mseeger: I don't know how to do that. I don't know what `e.__class__.__name__ would be for `botocore.exceptions.ClientError`. Also, there are different `ClientError`, they differ only by error message.

I find this matching on class name quite brittle.

syne_tune/remote/scheduling.py
```diff
@@ -39,3 +43,39 @@ def wrapper(*args, **kwargs):
...
+        print(try_import_aws_message())
+
+    def errorcatch(some_function):
+        @functools.wraps(some_function)
+        def wrapper(*args, **kwargs):
+            for idx in range(ntimes_resource_wait):
+                try:
+                    return some_function(*args, **kwargs)
+                except ClientError as ex:
+                    if errorname not in str(ex):
```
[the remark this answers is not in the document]
mseeger: Oh wait, this is wrong! I only want to catch ClientError

benchmarking/commons/launch_remote_common.py
mseeger: I don't know this stuff well

mseeger: You know now what we need, maybe you have a proposal

mseeger: We need something that triggers on `botocore.exceptions.ClientError`, and only if the error string contains "ResourceLimitExceeded". Any other exception, or `ClientError` with other message, should pass through

mseeger:
@jgolebiowski I think we should aim to close this. The situation is this:

* We have a specific need, namely that a certain exception is thrown, of a certain subtype, when you run out of quota. Having this in benchmarking (or elsewhere in Syne Tune) would be great
* I do not have the skills to connect your solution to this need. I don't know how stable it is to match class names of exceptions, in particular if they are from some external dependency. Your solution also does not allow dependence on the exception message

Solutions:
* Either we replace your general case with the specific case we need at the moment (I don't really see a broader one)
* Or you make sure your general case works for the specific one, hopefully not making it a lot more complex

mseeger:
How about a refactoring of your code like this?

```
a-b
```

benchmarking/commons/launch_remote_common.py
jgolebiowski: I would suggest to bake the decorator logic directly into this function, for example as:
```
def fit_sagemaker_estimator(backoff_wait_time: int, estimator: EstimatorBase, ntimes_resource_wait: int = 100, length2sleep: int = 360, **kwargs):
    """
    Runs ``estimator.fit(**kwargs)``. If ``backoff_wait_time > 0``, this is
    wrapped into the ``backoff_boto_clienterror`` decorator, which makes sure
    that if ``fit`` fails with ``ClientError`` of type "ResourceLimitExceeded",
    we wait for ``backoff_wait_time`` seconds and try again (up to 100 times).
    If ``backoff_wait_time <= 0``, the call of ``fit`` is not wrapped.
    :param backoff_wait_time: See above.
    :param estimator: SageMaker estimator to call ``fit`` for
    :param kwargs: Arguments for ``estimator.fit``
    """

    # If backoff_wait_time is None, run standard fitting
    if backoff_wait_time == 0:
        estimator.fit(**kwargs)

    for idx in range(ntimes_resource_wait):
        try:
            return estimator.fit(**kwargs)
        except Exception as e:
            if not e == botocore.exceptions.ClientError:
                raise (e)
        logger.info(
            f"botocore.exceptions.ClientError[{errorname}] detected "
            f"when calling <{some_function.__name__}>. Waiting "
            f"{length2sleep / 60} minutes before retrying"
        )
        time.sleep(length2sleep)
```

jgolebiowski: Alternatively, we can adapt the previously available decorator and then use it:

```
def backoff(error: Exception, ntimes_resource_wait: int = 100, length2sleep: float = 600):
    """
    Decorator that back offs for a fixed about of s after a given error is detected
    """
    def errorcatch(some_function):
        @functools.wraps(some_function)
        def wrapper(*args, **kwargs):
            for idx in range(ntimes_resource_wait):
                try:
                    return some_function(*args, **kwargs)
                except Exception as e:
                    if not e == error:
                        raise (e)

                logger.info(
                    f"{error} detected when calling <{some_function.__name__}>, waiting {length2sleep / 60} minutes before retring"
                )
                time.sleep(length2sleep)
                continue
        return wrapper

    return errorcatch
```

and the insides of the function becomes

```
    @backoff_boto_clienterror(
        error=botocore.exceptions.ClientError, length2sleep=backoff_wait_time
    )
    def fit_sagemaker_estimator_with_backoff(estimator: EstimatorBase, **kwargs):
        estimator.fit(**kwargs)

    if backoff_wait_time > 0:
        fit_sagemaker_estimator_with_backoff(estimator, **kwargs)
    else:
        estimator.fit(**kwargs)
```

jgolebiowski (changes requested):
I have combined all the comments under the fit_sagemaker_estimator function

mseeger:
OK, I think this is it. Please do check it again. I have not tested this (should try and run with instance type for which I have 0 limit)

mseeger:
Tested this:

`(st_with_ray) 88665a46461b:syne-tune matthis$ python benchmarking/examples/launch_local/launch_remote.py --method BO --benchmark transformer_wikitext2 --experiment_tag blurb1 --instance_type ml.g5.24xlarge --estimator_fit_backoff_wait_time 10
Failed to import apex. You can still train with --precision {float|double}.
coolname is not installed, will not be used
Master random_seed = 163844129
  0%|                                                                                                              | 0/1 [00:00<?, ?it/s]blurb1-BO-0
hyperparameters = {'experiment_tag': 'blurb1', 'benchmark': 'transformer_wikitext2', 'method': 'BO', 'save_tuner': 0, 'num_seeds': 1, 'start_seed': 0, 'random_seed': 163844129, 'scale_max_wallclock_time': 0, 'launched_remotely': 1, 'instance_type': 'ml.g5.24xlarge', 'verbose': 0, 'remote_tuning_metrics': 1, 'delete_checkpoints': 0, 'num_gpus_per_trial': 1}
Results written to s3://sagemaker-us-west-2-719355911555/syne-tune/blurb1/BO-0/
botocore.exceptions.ClientError[ResourceLimitExceeded] detected when calling estimator.fit. Waiting 0.16666666666666666 minutes before retrying
botocore.exceptions.ClientError[ResourceLimitExceeded] detected when calling estimator.fit. Waiting 0.16666666666666666 minutes before retrying
`

mseeger:
Nice, I have 0 quota for `ml.g5.24xlarge`. Seems to work

jgolebiowski (approved):
This solution looks good, thanks for adapting.

PR #677, opened 2023-05-16
chore: Sklearn estimator now fits the surrogate model without considering pending evaluations.

jgolebiowski:
The behaviour for `syne_tune.optimizer.schedulers.searchers.bayesopt.models.gp_model.GaussProcEstimator.fit_from_state` is 

If ``self.state.pending_evaluations`` is not empty, we proceed as follows:
* Compute posterior for state without pending evals
* Draw fantasy values for pending evals
* Recompute posterior (without fitting)

In the case of Sklearn Estimators we do not support fantasizing therefore the process should stop at step 1. The method can be later overriden if fantasizing is to be added.

----

### DIFF
--- a/syne_tune/optimizer/schedulers/searchers/bayesopt/models/sklearn_estimator.py
+++ b/syne_tune/optimizer/schedulers/searchers/bayesopt/models/sklearn_estimator.py
@@ -59,11 +59,22 @@ class SklearnEstimatorWrapper(Estimator):
         If your surrogate model is not Bayesian, or does not have hyperparameters,
         you can ignore the ``update_params`` argument,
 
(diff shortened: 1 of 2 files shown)


mseeger (approved):
Makes sense. I was kind of surprised you tried to deal with fantasizing before. It is not so important in practice.

jgolebiowski:
> Makes sense. I was kind of surprised you tried to deal with fantasizing before. It is not so important in practice.

Yes agreed, I tried to disable fantasizing on the Searcher level before but I think this is a better way.

jgolebiowski:
Replaced by https://github.com/awslabs/syne-tune/pull/681

PR #678, opened 2023-05-16
chore: fix unit test

wesk:
### Problem
Unit test started failing after https://github.com/awslabs/syne-tune/pull/672 was merged, which adds an assert about the config space for the MOREA method. The unit test was returning an invalid config space, which failed the assert. 

### Solution
Have the mock blackbox return a specific config space. 

Alternative considered: we can actually remove the `mock_blackbox` entirely, but then the test would run for >1sec.

----

### DIFF
--- a/tst/benchmarking/benchmark_commons/test_hpo_main_local.py
+++ b/tst/benchmarking/benchmark_commons/test_hpo_main_local.py
@@ -23,6 +23,7 @@ from benchmarking.nursery.benchmark_multiobjective.benchmark_definitions import 
     nas201_mo_benchmark,
 )
 from benchmarking.nursery.benchmark_multiobjective.hpo_main import main
+from syne_tune.config_space import randint
 
 
 class HPOMainLocalTests(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


tst/benchmarking/benchmark_commons/test_hpo_main_local.py
```diff
@@ -50,27 +51,37 @@ def test_tuner_is_run_the_expected_number_of_times(
             "nas201-ImageNet16-120": nas201_mo_benchmark("ImageNet16-120"),
         }
 
+        config_space = {
```
mseeger, line 54: Let us at least use the correct space for these benchmarks:

```
    hp_cols = ["hp_x0", "hp_x1", "hp_x2", "hp_x3", "hp_x4", "hp_x5"]

    config_space = {
        node: choice(
            ["avg_pool_3x3", "nor_conv_3x3", "skip_connect", "nor_conv_1x1", "none"]
        )
        for node in hp_cols
    }
    config_space["epochs"] = 200
```

mseeger (changes requested):

tst/benchmarking/benchmark_commons/test_hpo_main_local.py
wesk: Updated

mseeger (approved):

PR #680, opened 2023-05-17
fix: Remote launcher must not pass extra params to hpo_main

mseeger:
The script remote_launch.py has some args which must not be passed to hpo_main.py, since they are specific to remote launching. Currently, remote launching fails because of that.

----

### DIFF
--- a/benchmarking/commons/hpo_main_local.py
+++ b/benchmarking/commons/hpo_main_local.py
@@ -246,7 +246,9 @@ def start_benchmark_local_backend(
         random_seed = effective_random_seed(master_random_seed, seed)
         np.random.seed(random_seed)
         print(
-            f"Starting experiment ({method}/{benchmark_name}/{seed}) of {experiment_tag}"
+            f"Starting experiment ({method}/{benchmark_name}/{seed}) of {experiment_tag}\n"
+            f"  max_wallclock_time = {benchmark.max_wallclock_time}, "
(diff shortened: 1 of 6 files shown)


wesk:
Rather than creating this new `skip_initial_jobs` parameter, an alternative would be to make the command safe to retry. For example, in a 5-job experiment, if jobs 1, 3, 4, 5 all succeed but job 2 fails, the command should be smart enough so that if we re-invoke the experiment, we can re-launch just job 2.

This would be useful because not all jobs fail immediately with a sync error, some failures may be async, for example due to server errors.

wesk (changes requested):
To quickly fix remote launching in main, we reverted https://github.com/awslabs/syne-tune/pull/662. We can circle-back to determine the best way to add this feature and then open a new `feat:` PR.

mseeger:
OK, we need to decide whether we want to allow launch_remote.py to have additional args which are not passed down to hpo_main.py, or not.

If yes, and we don't want the solution I put in (which is ugly), we need a better one (@jgolebiowski ?). If no, this is OK for the short term, but I feel this is going to come back. Take the mechanism of Jacek to try again when launching a job failed due to service quota exceeded. You want to configure that one, but that option not be passed down.

mseeger:
OK, #688 is doing what #662 was supposed to do, without introducing issues. Can close this one here.

PR #683, opened 2023-05-19
fix: Revert "feat: Allow to skip initial jobs in remote launching (#662)"

wesk:
This reverts commit d5cf1ae8beca969e07f21c7fb4bce45842473c4f.

----

### DIFF
--- a/benchmarking/commons/launch_remote_common.py
+++ b/benchmarking/commons/launch_remote_common.py
@@ -67,13 +67,3 @@ def sagemaker_estimator_args(
         if benchmark.estimator_kwargs is not None:
             sm_args.update(benchmark.estimator_kwargs)
     return sm_args
-
-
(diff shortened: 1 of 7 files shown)


jgolebiowski:
This PR introduced breaking changes and the fix for the issues is not trivial, see comments in https://github.com/awslabs/syne-tune/pull/680. We should avoid having isues in the library but also should not put developing the right fix under time pressure. The best plan is to 1/ revert the original change 2/ work on an improvments to https://github.com/awslabs/syne-tune/pull/662 without a ticking clock and 3/ merge the #662 functionality without issues.

wesk:
Verified that this revert fixes remote launching. The three `romantic-dodo` jobs have the revert, whereas the three `viridan-malkoha` jobs do not. 

Screen Shot 2023-05-19 at 10 40 30 AM

jgolebiowski (approved):

PR #684, opened 2023-05-19
feat: Adding new searcher that uses contributed surrogate models. 

jgolebiowski:
Adding new searcher that uses contributed surrogate models.
This comes with a new test and an example showing how it can be done

----

### DIFF
--- a/examples/launch_contributed_surrogate_bo.py
+++ b/examples/launch_contributed_surrogate_bo.py
@@ -21,17 +21,17 @@ from syne_tune import Tuner, StoppingCriterion
 from syne_tune.backend import LocalBackend
 from syne_tune.config_space import randint
 from syne_tune.optimizer.schedulers import FIFOScheduler
-from syne_tune.optimizer.schedulers.searchers.bayesopt.sklearn.contributed_surrogate_searcher import (
-    ContributedSurrogateSearcher,
+from syne_tune.optimizer.schedulers.searchers.bayesopt.models.meanstd_acqfunc_impl import (
+    EIAcquisitionFunction,
+)
+from syne_tune.optimizer.schedulers.searchers.sklearn.sklearn_surrogate_searcher import (
+    SklearnSurrogateSearcher,
 )
 from syne_tune.optimizer.schedulers.searchers.bayesopt.sklearn.estimator import (
     SklearnEstimator,
 )
 from syne_tune.optimizer.schedulers.searchers.bayesopt.sklearn.predictor import (
     SklearnPredictor,
-)
-from syne_tune.optimizer.schedulers.searchers.bayesopt.tuning_algorithms.bo_algorithm_components import (
-    IndependentThompsonSampling,
 )
 
 
@@ -95,11 +95,11 @@ def main():
     }
     # Scheduler (i.e., HPO algorithm)
     myestimator = BayesianRidgeEstimator()
-    searcher = ContributedSurrogateSearcher(
+    searcher = SklearnSurrogateSearcher(
         config_space=config_space,
         metric="mean_loss",
         estimator=myestimator,
-        scoring_class_and_args=IndependentThompsonSampling,
+        scoring_class_and_args=EIAcquisitionFunction,
     )
 
     scheduler = FIFOScheduler(
@@ -120,7 +120,7 @@ def main():
         trial_backend=LocalBackend(entry_point=entry_point),
         scheduler=scheduler,
         stop_criterion=StoppingCriterion(max_wallclock_time=300),
-        n_workers=4,  # how many trials are evaluated in parallel
+        n_workers=1,  # how many trials are evaluated in parallel
     )
     tuner.run()
 
--- a/examples/launch_contributed_surrogate_bo.py
+++ b/examples/launch_contributed_surrogate_bo.py
@@ -1,0 +1,129 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+import copy
+from pathlib import Path
+from typing import Tuple
+
+import numpy as np
+from sklearn.linear_model import BayesianRidge
+
+from syne_tune import Tuner, StoppingCriterion
+from syne_tune.backend import LocalBackend
+from syne_tune.config_space import randint
+from syne_tune.optimizer.schedulers import FIFOScheduler
+from syne_tune.optimizer.schedulers.searchers.bayesopt.sklearn.contributed_surrogate_searcher import (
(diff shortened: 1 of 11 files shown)


syne_tune/optimizer/schedulers/searchers/bayesopt/sklearn/contributed_surrogate_searcher.py
```diff
@@ -0,0 +1,127 @@
...
+)
+from syne_tune.optimizer.schedulers.searchers.model_based_searcher import (
+    BayesianOptimizationSearcher,
+)
+from syne_tune.optimizer.schedulers.searchers.utils import make_hyperparameter_ranges
+
+logger = logging.getLogger(__name__)
+
+
+class ContributedSurrogateSearcher(BayesianOptimizationSearcher):
```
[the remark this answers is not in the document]
mseeger: This is very nice!

Once we have particular instances of this searcher, for example the ICML thing, we should then also invest in a factory, so user just has to specify `searcher` and `search_options` in the scheduler. But for now, this is exactly what we need.

examples/launch_contributed_surrogate_bo.py
```diff
@@ -0,0 +1,129 @@
...
+        "height": randint(1, 20),
+        "epochs": 10,
+    }
+    # Scheduler (i.e., HPO algorithm)
+    myestimator = BayesianRidgeEstimator()
+    searcher = ContributedSurrogateSearcher(
+        config_space=config_space,
+        metric="mean_loss",
+        estimator=myestimator,
+        scoring_class_and_args=IndependentThompsonSampling,
```
mseeger, line 102: Careful! This is precisely not what you want here! For ITS, this does not even use the acquisition function.

We should not allow the choice of `initial_scoring` for this searcher, because ITS does not make sense, and the only other thing is to use the AF for scoring.

syne_tune/optimizer/schedulers/searchers/bayesopt/sklearn/contributed_surrogate_searcher.py
```diff
@@ -0,0 +1,127 @@
...
+    Most of the implementation is generic in
+    :class:`~syne_tune.optimizer.schedulers.searchers.model_based_searcher.BayesianOptimizationSearcher`.
+    """
+
+    def __init__(
+        self,
+        config_space: Dict[str, Any],
+        metric: str,
+        estimator: SklearnEstimator,
+        scoring_class_and_args: ScoringClassAndArgs = None,
```
mseeger, line 59: This should be `acquisition_class`.

To explain this. In GP-BO, you can use the AF for (1) scoring initial cands, and (2) fine-tune by local search. But for (1), you can also use something else. For example, ITS is *independent* of the AF.

mseeger: Sure we need `ScoringClassAndArgs`? I'd just use `AcquisitionClassAndArgs`.

examples/launch_contributed_surrogate_bo.py
```diff
@@ -0,0 +1,129 @@
...
+from syne_tune.optimizer.schedulers.searchers.bayesopt.sklearn.contributed_surrogate_searcher import (
+    ContributedSurrogateSearcher,
+)
+from syne_tune.optimizer.schedulers.searchers.bayesopt.sklearn.estimator import (
+    SklearnEstimator,
+)
+from syne_tune.optimizer.schedulers.searchers.bayesopt.sklearn.predictor import (
+    SklearnPredictor,
+)
+from syne_tune.optimizer.schedulers.searchers.bayesopt.tuning_algorithms.bo_algorithm_components import (
```
mseeger, line 33: Don't use this. Sorry this is a bit confusing, see below

tst/schedulers/test_schedulers_api.py
```diff
@@ -376,6 +382,27 @@ def make_transfer_learning_evaluations(num_evals: int = 10):
         base_scheduler_factory=FIFOScheduler,
         searcher="random",
     ),
+    FIFOScheduler(
+        config_space,
+        searcher=ContributedSurrogateSearcher(
+            config_space=config_space,
+            metric="mean_loss",
+            estimator=TestEstimator(),
+            scoring_class_and_args=IndependentThompsonSampling,
```
mseeger, line 391: No, don't use this

tst/util_test.py
```diff
@@ -35,6 +43,19 @@
 )
 
 
+class TestPredictor(SklearnPredictor):
```
mseeger, line 46: Do you have an example somewhere, which also outputs predictive variances?

mseeger (changes requested):

examples/launch_contributed_surrogate_bo.py
```diff
@@ -0,0 +1,129 @@
...
+            * "mean": Predictive means in shape of ``(n,)``
+            * "std": Predictive stddevs, shape ``(n,)``
+        """
+        print(f"Predicting with BayesianRidgePredictor using X.shape={X.shape}")
+        return self.ridge.predict(X, return_std=True)
+
+
+class BayesianRidgeEstimator(SklearnEstimator):
+    """
+    Base class for the sklearn Estimators
```
wesk, line 65: `SklearnEstimator` is the base class, not `BayesianRidgeEstimator`, right?

syne_tune/optimizer/schedulers/searchers/bayesopt/sklearn/contributed_surrogate_searcher.py
```diff
@@ -0,0 +1,127 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
wesk: I also like the specificity of `sklearn` and not `contributed`, which I find confusing

examples/launch_contributed_surrogate_bo.py
jgolebiowski: yes, `BayesianRidgeEstimator` is only defined here for the purpose of the example

syne_tune/optimizer/schedulers/searchers/bayesopt/sklearn/contributed_surrogate_searcher.py
jgolebiowski: I have added a the `ScoringClassAndArgs` in `syne_tune/optimizer/schedulers/searchers/bayesopt/tuning_algorithms/base_classes.py`. Overall I think the skleanr searcher will rely on acquisition functions w.o gradients. The local optimizatin is already being skipped in https://github.com/awslabs/syne-tune/pull/684/files#diff-63315ef366b2e2b3cde9ea92943d6c9e8db65fcd630e8f4fce379f754fdc406eR99 so I dont think the gradients would be useful.

Note that `AcquisitionFunction` inherits from`ScoringFunction` so any acquitision function can be used here. See the launch example that uss EI.

tst/schedulers/test_schedulers_api.py
jgolebiowski: changed to `LCBAcquisitionFunction`

tst/util_test.py
jgolebiowski: The example with a proper model is in launch examples. I wanted to keep this as simple as possible but updated the stds to be non-zero to make sure it does not crash anything

mseeger (approved):
Just a small thing: I'd write SKLearn instead of Sklearn. Everywhere.

mseeger (approved):

PR #687, opened 2023-05-22
fix: typos in "Getting started" code example

ystein:
Fix the following typos in the example code in the Getting Started section from the README:

* `train_height_simple` expects parameter `--steps` but `launch_height_simple.py` defines `epochs` as HPO parameter only and not `steps` -> rename `steps` to `epochs` in `train_height_simple`
* in `launch_height_simple.py`: `entry_point` set to `train_height.py` which does not match the name `launch_height_simple.py` from the first comment in the previous code cell -> changed `entry_point` to `launch_height_simple.py`

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,12 +64,12 @@ if __name__ == '__main__':
     root = logging.getLogger()
     root.setLevel(logging.INFO)
     parser = ArgumentParser()
-    parser.add_argument('--steps', type=int)
+    parser.add_argument('--epochs', type=int)
     parser.add_argument('--width', type=float)
     parser.add_argument('--height', type=float)
     args, _ = parser.parse_known_args()
     report = Reporter()
-    for step in range(args.steps):
+    for step in range(args.epochs):
         time.sleep(0.1)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,16 +64,16 @@ if __name__ == '__main__':
     root = logging.getLogger()
     root.setLevel(logging.INFO)
     parser = ArgumentParser()
-    parser.add_argument('--steps', type=int)
+    parser.add_argument('--epochs', type=int)
     parser.add_argument('--width', type=float)
     parser.add_argument('--height', type=float)
     args, _ = parser.parse_known_args()
     report = Reporter()
-    for step in range(args.steps):
```
mseeger, line 72: Please revert changes from here on until the end of the loop. `dummy_score` depends on `step`, which starts from 0, and then we report `epoch=step + 1`.

mseeger (changes requested):
Thanks for this fix!

README.md
ystein: Done

PR #688, opened 2023-05-22
feat: Allow to skip initial jobs in remote launching

mseeger:
Does what #662 was supposed to do. We can add options only used by `launch_remote.py`, they'll just be ignored when used for `hpo_main.py`.

----

### DIFF
--- a/docs/source/tutorials/benchmarking/bm_simulator.rst
+++ b/docs/source/tutorials/benchmarking/bm_simulator.rst
@@ -273,6 +273,25 @@ In this case, experiments are sliced along the axis
 ``("nas201", "fcnet", "lcbench")`` to be run in parallel in different SageMaker
 training jobs.
 
+Skipping Initial Jobs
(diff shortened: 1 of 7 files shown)


docs/source/tutorials/benchmarking/bm_simulator.rst
```diff
@@ -273,6 +273,25 @@ In this case, experiments are sliced along the axis
 ``("nas201", "fcnet", "lcbench")`` to be run in parallel in different SageMaker
 training jobs.
 
+Skipping Initial Jobs
+~~~~~~~~~~~~~~~~~~~~~
+
+When running a remote launching command which starts many SageMaker training
+jobs, it can happen that the loop breaks after a certain number of jobs have
+been started. In this case, you can restart the command, asking it to skip the
```
wesk, line 281: What are the typical errors that cause the loop to break?

PR #692, opened 2023-05-25
fix: Remove (CITATION?). Currently, these results are not published a…

mseeger:
…nywhere

----


wesk (approved):

wesk:
Thanks for the fix @mseeger! And thanks for reporting this @ystein!

PR #694, opened 2023-05-26
feat: Some extra plotting features

mseeger:
Extends plotting in two ways:

* In `TrialsOfExperimentResults`, can now plot single and multi fidelity methods in the same figure
* Can select the subfigures to be plotted. This is useful if you group into (say) 4 or 8 buckets, but then want to create a sequence of plots comparing pairs of these

This is stuff I needed for the ODSC tutorial, they are just coming in now

----

### DIFF
--- a/syne_tune/experiments/plotting.py
+++ b/syne_tune/experiments/plotting.py
@@ -643,7 +643,7 @@ class ComparativeResults:
         plot_params: Optional[PlotParameters] = None,
         file_name: Optional[str] = None,
         extra_results_keys: Optional[List[str]] = None,
-    ) -> Optional[Dict[str, Any]]:
+    ) -> Dict[str, Any]:
         """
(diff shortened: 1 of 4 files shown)


syne_tune/experiments/plotting.py
```diff
@@ -569,13 +575,19 @@ def _plot_figure(
...
+                else subplots.subplot_indices
+            )
         else:
             subplot_xlims = None
             nrows = ncols = 1
             subplots_kwargs = dict(nrows=nrows, ncols=ncols)
             subplot_titles = None
             legend_no = [0]
             title_each_figure = False
+            subplot_indices = [0]
```
wesk, line 590: Should this be set here?

mseeger: Yes, there is a loop over this list below.

wesk (approved):

PR #695, opened 2023-05-26
chore: enable caching for unit tests, fix README typo

wesk:
Two small changes: 
1. Enable caching for unit tests, this should make them run faster. This setting is already enabled for our integ and end-to-end tests, see https://github.com/awslabs/syne-tune/blob/main/.github/workflows/run-syne-tune.yml#L51
2. Fix broken link in README

Before this change, duration of most recent 3 test runs: 12m 10s, 14m 55s, 19m 7s.

After this change: 8m 37s, 7m 54s, 7m 12s

Roughly 30% improvement. 

Confounding factors: 
1. Sometimes there's a delay until a machine is available to pick up the new job
2. If there's a cache-miss with a new library version, it will download it from the internet 

----

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -26,6 +26,7 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
+        cache: 'pip'
     - name: Display Python version
       run: python -c "import sys; print(sys.version)"
     - name: Install dependencies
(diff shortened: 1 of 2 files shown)


mseeger (approved):
OK, so does this work? Certainly cannot hurt, I suppose

wesk:
In my testing it reduces the amount of time that the unit tests run for by ~30%. Refer to PR description above for more details.

mseeger:
Thanks @wesk It seems to me even faster than what you say here. This is great!

PR #696, opened 2023-05-27
feat: Allow to transform result dataframe before aggregation in plotting

mseeger:
This allows to plot a derived metric, which is not contained in the results dataframe, but which depends on the content there. The first application of this is to plot the hypervolume index in the multi-objective context.

----


mseeger:
In fact, I added the hypervolume indicator code. @jgolebiowski , would have to confirm that this gives the same results as your special code

syne_tune/optimizer/schedulers/multiobjective/utils.py
```diff
@@ -23,41 +23,54 @@
...
+    :return: Cumulative hypervolume array, shape ``(npoints,)``
     """
     if reference_point is None:
-        reference_point = results_array.max(axis=0) * (1 + EPSILON) + EPSILON
+        reference_point = default_reference_point(results_array)
     indicator_fn = HV(ref_point=reference_point)
     hypervolume_indicator = np.zeros(shape=len(results_array))
     for idx in range(len(results_array)):
-        hypervolume_indicator[idx] = indicator_fn(results_array[0:idx])
+        hypervolume_indicator[idx] = indicator_fn(results_array[0 : (idx + 1)])
```
mseeger, line 75: To me, `0:idx` seems a bug. The first entry returned is always 0, and this does not return `HV` for the full `results_array`.

syne_tune/experiments/plotting.py
```diff
@@ -433,6 +434,7 @@ def _aggregrate_results(
         df: pd.DataFrame,
         plot_params: PlotParameters,
         extra_results_keys: Optional[List[str]],
+        dataframe_transform: Optional[DataFrameTransform],
```
jgolebiowski, line 437: i think adding the transform here conflates too many functionalities - derivation of new metrics and aggregation in the same function. I would suggest to add a separate function that enriches each results dataset with a new column with the derived metric (e.g. hypervoume). That will effectively remove this functionality from _aggregate and allow it to focus on aggregating a metric (original or newly derived)

Users (or plotting function) would then apply the transformation first and then call the aggregate function

jgolebiowski: On a second note, using a transformation is very flexible but also can be less streamlined + there is a potetial problem of mutating the original results frame, especially if we split new metrics addition from aggregation as suggested.

Maybe it would be enough to have a function that appends a new column to the dataframe based on the original dataframe. Enriching data by appending columns (instead of mutating the whole frame) is a pretty nice pattern since the original data is always available and the new columns are appended one by one which makes it very easy to debug. We can use something with signature 
`new_column_generator: Callable[[pd.DataFrame], pd.Series]`
with the following usage
`dataframe["new column"] = new_column_generator(dataframe)`

syne_tune/experiments/multiobjective.py
```diff
@@ -0,0 +1,65 @@
...
+import pandas as pd
+import numpy as np
+
+from syne_tune.optimizer.schedulers.multiobjective.utils import hypervolume_cumulative
+
+
+DEFAULT_HYPERVOLUME_INDICATOR_NAME = "hypervolume_indicator"
+
+
+def append_hypervolume_indicator(
```
jgolebiowski, line 24: I think this is ok for the current solution but please see the comments on alternatives to the trsnsform.

jgolebiowski (changes requested):
Looks like a very nice change, I have left some comments on disentangling enriching the data and aggregating the data + potentally simpler way to append new derived metrics.

wesk:
Rather than transforming the dataframe, as an alternative approach, have you considered appending the new information into the `results.csv.zip` file, using the feature you added earlier this year? https://syne-tune.readthedocs.io/en/latest/faq.html#how-can-i-write-extra-results-for-an-experiment

mseeger:
@wesk The problem is that you often would like to plot something you have not recorded explicitly during experimentation, but which can be derived from what you recorded. I think this is legitimate use.

syne_tune/experiments/plotting.py
mseeger: About your earlier comment: `_aggregate_results` is an internal method, the user calls `plot`. Note that data is loaded and aggregated for each call of `plot`, there is no dataframe which is maintained. So, while I will change it to create a single column, I will not change the fact that is passed to `plot`.

syne_tune/experiments/plotting.py
```diff
@@ -434,11 +435,28 @@ def _figure_shape(plot_params: PlotParameters) -> Tuple[int, int]:
             nrows = ncols = 1
         return nrows, ncols
 
+    def _transform_dataframe(
+        self,
+        df: pd.DataFrame,
+        plot_params: PlotParameters,
```
jgolebiowski, line 441: We should just pass metric directly to this function.

syne_tune/experiments/plotting.py
```diff
@@ -434,11 +435,28 @@ def _figure_shape(plot_params: PlotParameters) -> Tuple[int, int]:
...
+        )
+        new_column = dataframe_column_generator(df)
+        return df.assign(**{metric: new_column})
+
     def _aggregrate_results(
         self,
         df: pd.DataFrame,
         plot_params: PlotParameters,
         extra_results_keys: Optional[List[str]],
+        dataframe_column_generator: Optional[DataFrameColumnGenerator],
```
jgolebiowski, line 459: I think this should be extracted from `_aggregate_results` to separate aggregation and transformation. I believe it is important to have one function/module fulfil one purpose as much as we can (see single responsibility principle). In this case, I think `plot` can accept the transformation given its current design but we should separate aggregation and enrichement. In fact `plot` should first update the setup_df using something like (can be inside the transform function with extra asserts):
```
setup_df["new column"] = None
for tuner_name, tuner_df in setup_df.groupby("tuner_name"):
    setup_df.loc[tuner_df.index, "new column"] = dataframe_column_generator(tuner_df)
                
```

and then `_aggregate` should not worry about any transformations anymore. This way both functions can be modified/extended independently as they do not share any concerns.

jgolebiowski (changes requested):
I still believe we should separate enrichment and aggregation in this change to make sure our code is modular.

syne_tune/experiments/plotting.py
mseeger: OK

syne_tune/experiments/plotting.py
mseeger: I disagree. First of all, these are some internal methods.
Second, there is grouping happening first w.r.t. setup, then w.r.t. tuner (i.e., seed). Then, you do 2 things on the dataframes grouped this way: add column, then aggregate.

A copy&paste job doing this 2-level grouping in two methods is not only inefficient, but also not clean.

I can rename the method, and introduce another helper method for the aggregation only. But let us remain realistic.

mseeger: My latest refactoring should address the concerns

syne_tune/experiments/plotting.py
```diff
@@ -434,20 +435,83 @@ def _figure_shape(plot_params: PlotParameters) -> Tuple[int, int]:
...
+            else:
+                key_sequence = [setup_name]
+            final_pos = df[ST_TUNER_TIME].idxmax()
+            final_row = dict(df.loc[final_pos])
+            for key in extra_results_keys:
+                _insert_into_nested_dict(
+                    dictionary=extra_results,
+                    key_sequence=key_sequence + [key],
+                    value=final_row[key],
+                )
```
jgolebiowski, line 501: I am not sure if I understand this part. Is this modifying the `extra_results` input argument? I think it woukd be good to add a docstring ot the function to make it clear that this is modified.

jgolebiowski (commented):
I think there is a fundamental difference in opinion of single responsibility design vs swift implementation. If you feel strongly about this topic we can discuss more offline, I understand that we want to push those changes quickly before the mid-June talks so I hope we can find a sensible solution.

syne_tune/experiments/plotting.py
mseeger: My feeling is you get a bit worked up on a pretty isolated point. In much of our code, lots of different things happen in a loop.

I can now pick any of these codes, and ask for two or three loops. What does this add?

We have something like this

Grouping loop:
- Do thing_a (by calling a method called thing_a)
- Do thing b on output of thing_a (by calling a method called thing_b)

You ask me to rewrite this into:

do_thing_a_with_grouping:
- Grouping loop:
   - Do thing_a
   
 do_thing_b_with_grouping:
 - Grouping loop:
    - Do thing_b
    
I really fail to see why the latter is better design.

It is not the case that slower code means better design.

mseeger: I know this single responsibility idea. My code does that. But as methods become higher-level, their single responsibility becomes more complex.

It is important to note a few things:
- The column to add must be computed from the grouped data for a single experiment only
- The dataframe with the extra column is not maintained, it is just an intermediate for the aggregation
- I've been adding extra methods to separate concerns

mseeger: I am also not sure if you looked at the most recent code. The method is not called `_aggregate_results` anymore.

mseeger:
OK, I think I have now addressed the concerns

jgolebiowski (approved):
LGTM, thanks for making the changes.

PR #700, opened 2023-05-30
fix: Avoid MOO warning message

wistuba:
```from syne_tune.experiments import load_experiment``` triggers 
```
Multi Objective Optimization dependencies are not imported since dependencies are missing. You can install them with
   pip install 'syne-tune[moo]'
or (for everything)
   pip install 'syne-tune[extra]'
```

Moving the import will avoid that.

### DIFF
--- a/syne_tune/experiments/experiment_result.py
+++ b/syne_tune/experiments/experiment_result.py
@@ -104,8 +104,10 @@ class ExperimentResult:
         ), f"All metrics must be maximized but the following modes were selected: {metric_modes}"
 
         if self.results is not None and len(self.results) > 0:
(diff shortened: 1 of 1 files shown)


mseeger:
Hello Martin,

we have these warnings output all over the place. This is a compromise between (a) being able to import stuff at the top, which seems the clean way to do, and (b) you know what you need to do if you run ST with restricted dependencies and get an exception.

I leave this to Jacek, but I got comments in the past about not including stuff except at the top, and anyway this local including does not always work, for example if it is about classes you want to derive from.

wistuba:
I'm fine addressing it a different way as long as the warning disappears. Are there any alternatives?

mseeger (approved):

PR #701, opened 2023-05-31
refactor: Clean up transformer_wikitext2 code

mseeger:
Cleans up benchmarking code. This is because I'd like to show it in the tutorial

----

### DIFF
--- a/benchmarking/commons/benchmark_definitions/transformer_wikitext2.py
+++ b/benchmarking/commons/benchmark_definitions/transformer_wikitext2.py
@@ -10,7 +10,6 @@
 # on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
-from typing import Dict, Any
 from pathlib import Path
 
 from benchmarking.commons.benchmark_definitions.common import RealBenchmarkDefinition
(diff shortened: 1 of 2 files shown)


benchmarking/commons/benchmark_definitions/transformer_wikitext2.py
```diff
@@ -26,41 +25,40 @@
...
+        ffn_ratio=1,
+        nlayers=2,
+        nhead=2,
+        bptt=35,
+        optimizer_name="sgd",
+        input_data_dir="./",
+        use_cuda=1,
+        seed=1111,
+        precision="float",
+        log_interval=200,
```
wesk, line 46: Where do these params come from?

mseeger: They are default values in the arg parser in the script. This just makes them more accessible, and shows that these ones can also be set.

benchmarking/training_scripts/transformer_wikitext2/training_script.py
wesk, line 1: Would it make sense to add a unit test as part of the refactor? Noticed that the test coverage is fairly low for this file: https://app.codecov.io/gh/awslabs/syne-tune/blob/main/benchmarking%2Ftraining_scripts%2Ftransformer_wikitext2%2Ftraining_script.py

mseeger: Maybe another time. This is just a benchmark, and anyway needs GPU

mseeger: We don't have any tests for benchmark training scripts. I'd not know how they would look like. Also, we need more of them, so do not want to discourage people.

wesk: Hmm, ok, we can consider this separately. In general, I just see the coverage is low for benchmarking, and the way to raise it is just to add tests, especially for low-coverage files. But it sounds like this might need a dedicated effort to establish a testing pattern.

wesk (approved):

PR #702, opened 2023-06-01
fix: Command line arguments of benchmarking

mseeger:
Command line arguments did not work. I also pulled together common CL args between local and sagemaker

----


wesk:
When did the command line arguments break?

mseeger:
We introduces `num_gpus_per_trial` for the local backend. Up from then, launching jobs with SM backend through benchmark would have been affected.

Yes, we are still lacking tests for benchmarking

wesk (approved):

PR #704, opened 2023-06-01
refactor: Simplification in scikit-learn based estimators

mseeger:
Small changes. Move wrapper code into one file, which is in line with all other wrapper code. Also, simplify imports.
Needed before BO developer tutorial PR goes out.

----


jgolebiowski (approved):
LGTM, having both in the same file is probably a good idea.

PR #708, opened 2023-06-07
fix: Some fixes for multi-objective benchmarking

mseeger:
Several fixes for multi-objective benchmarking. I still need to test the plotting, but this is already doing the right thing in terms of the random seeds.

- Wrappers in `baselines.py` need to respect a certain form, otherwise random seeds are not properly passed through
- Some fixes related to `metric` and `mode` being lists sometimes
- New code to filter results before plotting, so that we only have one row per trial (otherwise, HV indicator computation is way too slow). For example: The toy benchmarks evaluate 400 trials, each reports 200 times (200 epochs). This gives 80000 results per experiment. The HV indicator plotting as it is, would probably take several hours. Filtering leads to 400 results, and plotting is reasonably fast.
- Make sure plotting scripts show log output
- Better documentation
- Comments in developer tutorial on how to extend `baselines.py`

----


mseeger:
I am still working on confirming the plotting works. Aggregation seems to take a long time ...

mseeger:
I am still fighting with plotting the HV indicator. I'll have a look at Jacek's code. But I implemented the filtering for one row per trial, after which the plotting is fast.

syne_tune/optimizer/schedulers/multiobjective/utils.py
```diff
@@ -47,18 +48,32 @@ def hypervolume(
     return indicator_fn(results_array)
 
 
+def _linear_interpolate(hv_indicator: np.ndarray, indices: List[int]):
```
wesk, line 51: What about a unit test that checks if the new linear interpolation option works correctly? 

Also more broadly, were any of the fixes in this PR for issues with the core library logic itself? If yes, those would also be good candidates for additional tests.

PR #710, opened 2023-06-07
feat: Code for OTHPO paper

sighellan:
Adds the code for the paper "Obeying the Order: Introducing Ordered Transfer Hyperparameter Optimisation".

This includes some edits to the BoTorch searcher and the addition of a transfer version. Otherwise, the code has been added to a directory in the nursery directory. 

This is all the data we included with the submission, so it includes some large result files. It might be better to store these in a fork @geoalgo 

----


mseeger:
Could you please remove all the pdf files? Syne Tune cannot host this kind of data, you should put them in a paper and upload to arxiv or so. Then, this can be linked.

wesk:
Nice @sighellan! 

Initial comment: the diff is large (191 files), many of which are csv and binary files. How can we cut down the size of the PR? One initial direction would be to not include the dataset directly in Syne Tune, but to instead write an import script, so users can download it if they want.

sighellan:
I've removed a lot of the large files and included a link to my fork where they can be accessed. 

There are a small number of small ones remaining which I would like to keep:
- simopt files
    - benchmarking/nursery/othpo/simopt/generated_files/default_fixed_factors.p
    - benchmarking/nursery/othpo/simopt/generated_files/opt-price-random-walk-utility-context-2022-11-08.p
    - These are small and hard coded elsewhere in the code and due to config spaces were tricky to expose to the user. 
-  xgboost experiment results
    - 10 pickle files
    - These are the results from long sagemaker runs which are then used in the Syne Tune experiments. They're not intended to be recollected by new users.

benchmarking/nursery/othpo/bo_warm_transfer.py
```diff
@@ -0,0 +1,225 @@
...
+            nn,
+            de_str_dict,
+            sort_by_task_id,
+            shuffle_order,
+        )
+
+    return [de_string_config(point, de_str_dict) for point in points_to_evaluate]
+
+
+class WarmStartBayesianOptimization(BoTorch):
```
geoalgo, line 173: Nice one, not needed for this PR but we could put this one in `syne_tune/schedulers/` as I assume it could be generally useful.

syne_tune/optimizer/schedulers/searchers/botorch/botorch_transfer_searcher.py
```diff
@@ -0,0 +1,196 @@
...
+        return val
+
+
+def configs_from_df(df) -> List[dict]:
+    return [
+        {key: parse_value(df[key][ii]) for key in df.keys()} for ii in range(len(df))
+    ]
+
+
+class BoTorchTransfer(BoTorch):
```
geoalgo, line 52: Very useful thanks.

benchmarking/nursery/othpo/simopt/generated_files/default_fixed_factors.p
geoalgo, line 1: Sorry for the late reply, I was OOTO.

We cannot include pickle files in the main repo as those can create serious security issue.

What I would suggest is:
1. bundle all pickle files in an archive and put it in your fork 
2. add instructions to `wget` and extract the files
3. importantly include instructions to check the expected hash to make sure the pickle are not altered

geoalgo (commented):
LGTM regarding the code structure.

Regarding the pickle and binary files, I agree with @wesk and @mseeger, we cannot host them in syne-tune main branch.

This is because of size but mostly because of security as having pickle files introduce a security risk, I left a comment on how to address that. I hope it makes sense for you let me know if it does not.

benchmarking/nursery/othpo/bo_warm_transfer.py
sighellan: I'm tempted to add it here now and move it to syne_tune/schedulers with a more clear name once I have a bit more time.

benchmarking/nursery/othpo/simopt/generated_files/default_fixed_factors.p
sighellan: I didn't realise re the security issue. Would a json file have the same issue? It's really just a dict of some settings and lists/arrays of floats/ints.

geoalgo: A Json file does not have the same issue but it will bloat the repository (all files needs to be downloaded when you git clone). An alternative is to have the data in a panda dataframe, then you can save/load with csv.zip directly in pandas (`df.to_csv("file.csv.zip")` and `pd.read_csv("file.csv.zip")`).

Having the file on a fork is probably the easiest unless you can make all the files bellow a few KBs...

mseeger:
@sighellan Syne Tune cannot host experimental results for methods. I'd recommend just publishing them in a paper. If you do want to use files to be loaded here, you'd need to host them somewhere, and then load them.

Unfortunately, this also extends to datasets or even blackbox repository files. We cannot host them in the repo here.
We already make an exception for image files in the docs, but also there, we need to keep things small.

sighellan:
> @sighellan Syne Tune cannot host experimental results for methods. I'd recommend just publishing them in a paper. If you do want to use files to be loaded here, you'd need to host them somewhere, and then load them.
> 
> Unfortunately, this also extends to datasets or even blackbox repository files. We cannot host them in the repo here. We already make an exception for image files in the docs, but also there, we need to keep things small.

@mseeger 
We are publishing them in a paper, open-sourcing the code is part of that :) 

It's not a matter of hosting experimental results for our methods, but data used to compare methods. But I've moved it to a fork using wget like @geoalgo suggested. 

I've reduced the pull request to just .py, .md, .sh and .txt files.

I'm not sure what's happening with the docs test that's failing. I don't think I'm making any changes to the docs.

geoalgo (approved):
Thanks for removing the large files! Really cool to have this paper available in Syne Tune as well as your GP transfer learning extensions.

PR #711, opened 2023-06-09
feat: Multi Surrogate Multi Objective Searcher

jgolebiowski:
Adding the Multi Surrogate Multi Objective Searcher to SyneTune. 
To make this happen, added a few small changes:
1. NoOptimization optimizer now does not setup itself, this is necessary to use the same optimizer for multiple metrics
2. If no estimator for INTERNAL_METRIC_NAME can be found, ModelBasedSearcher uses the debug log from a randomly selected one
3. _assert_same_keys check only if skip_optimization is not None. None is the default case where all is skipped.
4. target_metric is passed to SKLearnEstimatorWrapper to differentiate estimators

----

### DIFF
--- a/examples/launch_asha_yahpo.py
+++ b/examples/launch_asha_yahpo.py
@@ -71,7 +71,7 @@ class BenchmarkInfo:
     resource_attr: str
 
 
-if __name__ == "__main__":
+def main():
     logging.getLogger().setLevel(logging.INFO)
 
     benchmark_infos = {
--- a/examples/launch_mb_mo_optimization.py
+++ b/examples/launch_mb_mo_optimization.py
@@ -10,16 +10,15 @@
 # on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
-import copy
 from pathlib import Path
-from typing import Tuple
+from typing import Dict, Any, Optional
 
 import numpy as np
-from sklearn.linear_model import BayesianRidge
 
 from syne_tune import Tuner, StoppingCriterion
 from syne_tune.backend import LocalBackend
 from syne_tune.config_space import randint, uniform
+from syne_tune.optimizer.baselines import BayesianOptimization
 from syne_tune.optimizer.schedulers import FIFOScheduler
 from syne_tune.optimizer.schedulers.multiobjective.multi_surrogate_multi_objective_searcher import (
     MultiObjectiveMultiSurrogateSearcher,
@@ -27,60 +26,30 @@ from syne_tune.optimizer.schedulers.multiobjective.multi_surrogate_multi_objecti
 from syne_tune.optimizer.schedulers.multiobjective.random_scalarization import (
     MultiObjectiveLCBRandomLinearScalarization,
 )
-from syne_tune.optimizer.schedulers.searchers.bayesopt.models.sklearn_model import (
-    SKLearnEstimatorWrapper,
-)
-from syne_tune.optimizer.schedulers.searchers.bayesopt.sklearn import (
-    SKLearnEstimator,
-    SKLearnPredictor,
-)
 
 
-class BayesianRidgePredictor(SKLearnPredictor):
-    """
-    Base class for the sklearn predictors
-    """
+def create_gaussian_process_estimator(
+    config_space: Dict[str, Any],
+    metric: str,
+    mode: Optional[str] = None,
+    random_seed: Optional[int] = None,
+    search_options: Optional[Dict[str, Any]] = None,
+):
+    scheduler = BayesianOptimization(
+        config_space=config_space,
+        metric=metric,
+        mode=mode,
+        random_seed=random_seed,
+        search_options=search_options,
+    )
+    searcher = scheduler.searcher  # GPFIFOSearcher
+    state_transformer = searcher.state_transformer  # ModelStateTransformer
+    estimator = state_transformer.estimator  # GaussProcEmpiricalBayesEstimator
 
-    def __init__(self, ridge: BayesianRidge):
-        self.ridge = ridge
-
-    def predict(
-        self, X: np.ndarray, return_std: bool = True
-    ) -> Tuple[np.ndarray, np.ndarray]:
-        """
-        Returns signals which are statistics of the predictive distribution at
-        input points ``inputs``.
-
-
-        :param inputs: Input points, shape ``(n, d)``
-        :return: Tuple with the following entries:
-            * "mean": Predictive means in shape of ``(n,)``
-            * "std": Predictive stddevs, shape ``(n,)``
-        """
-        return self.ridge.predict(X, return_std=True)
-
-
-class BayesianRidgeEstimator(SKLearnEstimator):
-    """
-    Base class for the sklearn Estimators
-    """
-
-    def __init__(self, *args, **kwargs):
-        self.ridge = BayesianRidge(*args, **kwargs)
-
-    def fit(
-        self, X: np.ndarray, y: np.ndarray, update_params: bool
-    ) -> BayesianRidgePredictor:
-        """
-        Implements :meth:`fit_from_state`, given transformed data.
-
-        :param X: Training data in ndarray of shape (n_samples, n_features)
-        :param y: Target values in ndarray of shape (n_samples,)
-        :param update_params: Should model (hyper)parameters be updated?
-        :return: Predictor, wrapping the posterior state
-        """
-        self.ridge.fit(X, y.ravel())
-        return BayesianRidgePredictor(ridge=copy.deepcopy(self.ridge))
+    # update the estimator properties
+    estimator.active_metric = metric
+    estimator._no_fantasizing = True
+    return estimator
 
 
 def main():
@@ -90,35 +59,38 @@ def main():
         "theta": uniform(0, np.pi / 2),
         "sleep_time": 0.01,
     }
-    # Scheduler (i.e., HPO algorithm)
-    sklearn_myestimators = {
-        "y1": BayesianRidgeEstimator(),
-        "y2": BayesianRidgeEstimator(),
-    }
+    metrics = ["y1", "y2"]
+    modes = ["min", "min"]
+    # Create Gaussian process estimators
+    # In ``search_options``, the GP model can be configured, see comments
+    # of ``GPFIFOSearcher``
+    search_options = {"debug_log": False}
     myestimators = {
-        metric: SKLearnEstimatorWrapper(estim, target_metric=metric)
-        for metric, estim in sklearn_myestimators.items()
+        metric: create_gaussian_process_estimator(
+            config_space=config_space,
+            metric=metric,
+            mode=mode,
+            search_options=search_options,
+        )
+        for metric, mode in zip(metrics, modes)
     }
     searcher = MultiObjectiveMultiSurrogateSearcher(
         config_space=config_space,
-        metric=["y1", "y2"],
+        metric=metrics,
         estimators=myestimators,
         scoring_class_and_args=MultiObjectiveLCBRandomLinearScalarization,
     )
 
     scheduler = FIFOScheduler(
         config_space,
-        metric=["y1", "y2"],
-        mode=["min", "min"],
+        metric=metrics,
+        mode=modes,
         searcher=searcher,
-        search_options={"debug_log": False},
+        search_options=search_options,
     )
 
     entry_point = str(
-        Path(__file__).parent
-        / "training_scripts"
-        / "mo_artificial"
-        / "mo_artificial.py"
+        Path(__file__).parent / "training_scripts" / "mo_artificial" / "mo_artificial.py"
     )
     tuner = Tuner(
         trial_backend=LocalBackend(entry_point=entry_point),
--- a/examples/launch_mb_mo_optimization.py
+++ b/examples/launch_mb_mo_optimization.py
@@ -1,0 +1,133 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+import copy
+from pathlib import Path
+from typing import Tuple
+
+import numpy as np
+from sklearn.linear_model import BayesianRidge
+
+from syne_tune import Tuner, StoppingCriterion
+from syne_tune.backend import LocalBackend
+from syne_tune.config_space import randint, uniform
+from syne_tune.optimizer.schedulers import FIFOScheduler
+from syne_tune.optimizer.schedulers.multiobjective.multi_surrogate_multi_objective_searcher import (
+    MultiObjectiveMultiSurrogateSearcher,
+)
+from syne_tune.optimizer.schedulers.multiobjective.random_scalarization import (
+    MultiObjectiveLCBRandomLinearScalarization,
+)
+from syne_tune.optimizer.schedulers.searchers.bayesopt.models.sklearn_model import (
+    SKLearnEstimatorWrapper,
+)
+from syne_tune.optimizer.schedulers.searchers.bayesopt.sklearn import (
+    SKLearnEstimator,
+    SKLearnPredictor,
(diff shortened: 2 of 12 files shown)


syne_tune/optimizer/schedulers/multiobjective/multi_surrogate_multi_objective_searcher.py
```diff
@@ -0,0 +1,155 @@
...
+
+        super().__init__(
+            config_space,
+            metric=metric,
+            mode=mode,
+            points_to_evaluate=points_to_evaluate,
+            random_seed_generator=kwargs.get("random_seed_generator"),
+            random_seed=kwargs.get("random_seed"),
+        )
+        self.estimator = estimators
```
mseeger, line 96: Couldn't you use `self.state_transformer.estimator`?

syne_tune/optimizer/schedulers/multiobjective/multi_surrogate_multi_objective_searcher.py
```diff
@@ -0,0 +1,155 @@
...
+            # Internal constructor, bypassing the factory
+            # Note: Members which are part of the mutable state, will be
+            # overwritten in ``_restore_from_state``
+            self._create_internal(**kwargs.copy())
+
+    def _update(self, trial_id: str, config: Dict[str, Any], result: Dict[str, Any]):
+        # Gather metrics
+        metrics = {name: result[name] for name in self._metric}
+
+        # Cost value only dealt with here if ``resource_attr`` not given
```
mseeger, line 124: Is your searcher using `self._cost_attr`? If not, you can remove this here. This is just a way for me in `GPFIFOSearcher` to allow for certain cost-aware variants in subclasses. No need to copy that here.

I think that even if "cost" is one of your metrics, you just want to not single it out here.

syne_tune/optimizer/schedulers/multiobjective/multi_surrogate_multi_objective_searcher.py
```diff
@@ -0,0 +1,155 @@
...
+            _trial_id = self._trial_id_string(trial_id, result)
+            msg = f"Update for trial_id {_trial_id}: metric = {metrics:.3f}"
+            if cost_val is not None:
+                msg += f", cost_val = {cost_val:.2f}"
+            logger.info(msg)
+
+    def clone_from_state(self, state):
+        # Create clone with mutable state taken from 'state'
+        init_state = decode_state(state["state"], self._hp_ranges_in_state())
+        estimator = self.state_transformer.estimator
```
mseeger, line 145: See, you don't use `self.estimator` here. I'd prefer not to have `self.estimator` at all.

syne_tune/optimizer/schedulers/multiobjective/random_scalarization.py
```diff
@@ -0,0 +1,94 @@
...
+        weights_sampler: Optional[Callable[[], Dict[str, float]]] = None,
+        kappa: float = 0.5,
+    ):
+        self.kappa = kappa
+        self.predictor = predictor
+        if active_metric is None:
+            active_metric = [name for name in predictor.keys()]
+        self.active_metric = active_metric
+
+        if weights_sampler is None:
```
mseeger, line 59: No. We want to control all random effects with a seed. I'd do it like this:

Add argument `random_seed: Optional[int] = None`.
This argument is used only if `weights_sampler` is not given. In that case:
`state = RandomState(random_seed)`

I comment above how to assign this random seed.

This is quite important to me. I'd like to retain the property that if a method is re-run with the same seed, then the random sampling behaviour should be exactly the same, as long as this is at all possible.

mseeger: If `random_seed` is `None`, you can still set it to 123 here if you like.

examples/launch_mb_mo_optimization.py
```diff
@@ -0,0 +1,109 @@
...
+            mode=mode,
+            search_options=search_options,
+        )
+        for metric, mode in zip(metrics, modes)
+    }
+    searcher = MultiObjectiveMultiSurrogateSearcher(
+        config_space=config_space,
+        metric=metrics,
+        estimators=myestimators,
+        scoring_class_and_args=MultiObjectiveLCBRandomLinearScalarization,
```
mseeger, line 82: Please use `scoring_class_and_args=(MultiObjectiveLCBRandomLinearScalarization, {"random_seed": random_seed})` to control the random seed in the acq func.

syne_tune/optimizer/schedulers/multiobjective/multi_surrogate_multi_objective_searcher.py
```diff
@@ -0,0 +1,155 @@
...
+            config_space,
+            metric=metric,
+            mode=mode,
+            points_to_evaluate=points_to_evaluate,
+            random_seed_generator=kwargs.get("random_seed_generator"),
+            random_seed=kwargs.get("random_seed"),
+        )
+        self.estimator = estimators
+        if scoring_class_and_args is None:
+            scoring_class_and_args = MultiObjectiveLCBRandomLinearScalarization
```
mseeger, line 98: Please use this, so that the random sampling in the acq func is controlled:

Higher up:
`from syne_tune.optimizer.schedulers.searchers.searcher_base import extract_random_seed`

Below:

```
if scoring_class_and_args is None:
    random_seed = extract_random_seed(**kwargs)
    scoring_class_and_args = (MultiObjectiveLCBRandomLinearScalarization, {"random_seed": random_seed})
```

mseeger: Together with my comment below, this will work

syne_tune/optimizer/schedulers/multiobjective/random_scalarization.py
```diff
@@ -0,0 +1,94 @@
...
+        predictor: Dict[str, Predictor],
+        active_metric: Optional[List[str]] = None,
+        weights_sampler: Optional[Callable[[], Dict[str, float]]] = None,
+        kappa: float = 0.5,
+    ):
+        self.kappa = kappa
+        self.predictor = predictor
+        if active_metric is None:
+            active_metric = [name for name in predictor.keys()]
+        self.active_metric = active_metric
```
mseeger, line 57: Please call `super().__init__` instead of this here

syne_tune/optimizer/schedulers/searchers/bayesopt/models/sklearn_model.py
```diff
@@ -82,9 +82,16 @@ class SKLearnEstimatorWrapper(Estimator):
     Wrapper class for sklearn estimators.
     """
 
-    def __init__(self, sklearn_estimator: SKLearnEstimator, *args, **kwargs):
+    def __init__(
+        self,
+        sklearn_estimator: SKLearnEstimator,
+        target_metric: str = None,
```
mseeger, line 88: Maybe call this `active_metric` as well?

syne_tune/optimizer/schedulers/searchers/bayesopt/tuning_algorithms/bo_algorithm_components.py
```diff
@@ -140,6 +140,12 @@ def f_df(x):
 
 
 class NoOptimization(LocalOptimizer):
+    def __init__(self, *args, **kwargs):
```
mseeger, line 143: Can you explain why this is needed?

syne_tune/optimizer/schedulers/searchers/model_based_searcher.py
```diff
@@ -157,7 +158,11 @@ def _create_internal(
             )
         self.acquisition_class = acquisition_class
         if isinstance(estimator, dict):
-            estimator_main = estimator[INTERNAL_METRIC_NAME]
+            # If no estimator for INTERNAL_METRIC_NAME can be found,
+            # use the debug log from a randomly selected one
+            estimator_main = estimator.get(
+                INTERNAL_METRIC_NAME, random.choice(list(estimator.values()))
```
mseeger, line 164: Why not just pick the first? Seems simpler to me.

mseeger (changes requested):

syne_tune/optimizer/schedulers/multiobjective/random_scalarization.py
```diff
@@ -0,0 +1,94 @@
...
+            ...
+        }
+    """
+
+    def __init__(
+        self,
+        predictor: Dict[str, Predictor],
+        active_metric: Optional[List[str]] = None,
+        weights_sampler: Optional[Callable[[], Dict[str, float]]] = None,
+        kappa: float = 0.5,
```
mseeger, line 51: Doc missing

syne_tune/optimizer/schedulers/multiobjective/random_scalarization.py
mseeger: An alternative is to provide a default implementation for the `weights_sampler` mapping, as a call with `__call__` method. The class would be generated with `random_seed` and would create `state` as member, and then use it in each `__call__`. This is maybe cleaner.

syne_tune/optimizer/schedulers/multiobjective/random_scalarization.py
```diff
@@ -0,0 +1,94 @@
...
+            predictor = self.predictor
+        weights = self.weights_sampler()
+        scores = np.zeros(len(candidates))
+        for metric, metric_predictor in predictor.items():
+            predictions = metric_predictor.predict_candidates(candidates)
+            predicted_mean = predictions[0]["mean"]
+            predicted_std = predictions[0]["std"]
+
+            metric_score = self._single_objective_score(predicted_mean, predicted_std)
+            metrics_score_normalized = scipy.stats.rankdata(metric_score)
```
mseeger, line 83: Can you comment on why `rankdata` is used here? I do not get it.

examples/launch_mb_mo_optimization.py
```diff
@@ -0,0 +1,109 @@
...
+        mode=mode,
+        random_seed=random_seed,
+        search_options=search_options,
+    )
+    searcher = scheduler.searcher  # GPFIFOSearcher
+    state_transformer = searcher.state_transformer  # ModelStateTransformer
+    estimator = state_transformer.estimator  # GaussProcEmpiricalBayesEstimator
+
+    # update the estimator properties
+    estimator.active_metric = metric
```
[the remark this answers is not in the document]
jgolebiowski: yes unfortunately we need this metric inside the estimator to gather the right outputs from results. This is not optimal but I could not find a better way to pass the metric names since they get automatically adjusted to `INTERNAL_METRIC_NAME`.

examples/launch_mb_mo_optimization.py
```diff
@@ -0,0 +1,109 @@
...
+    estimator._no_fantasizing = True
```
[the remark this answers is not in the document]
jgolebiowski: Good idea, thanks!

syne_tune/optimizer/schedulers/multiobjective/random_scalarization.py
jgolebiowski: This is just rank normalization of the outputs, otherwise some objectives can be on different scale than others making it difficult to compare

syne_tune/optimizer/schedulers/searchers/bayesopt/tuning_algorithms/bo_algorithm_components.py
jgolebiowski: The higher level class `LocalOptimizer` needs the active metric specified and expectes to ony work for this one metric, e.g. checks if the metric if there exists a predictor for this metric. We want to re-use the same NoOptimization class for all MO metrics which does not work if the default __init__ is used. We could alternatively refactor how the NoOptimization is specificed in the BO algorithm but I thought this should work for now

syne_tune/optimizer/schedulers/multiobjective/multi_surrogate_multi_objective_searcher.py
jgolebiowski: Actually we dont need that at all, its already handled in `_create_internal`

syne_tune/optimizer/schedulers/multiobjective/multi_surrogate_multi_objective_searcher.py
jgolebiowski: Agreed, removed.

examples/launch_mb_mo_optimization.py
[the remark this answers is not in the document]
mseeger: I am also not very happy about this internal renaming (this is old code), but I so far did not find an easy way to refactor this, because this internal code should work also if no name is given (except multi-objective).

I'd leave it as is, as long as it works.

syne_tune/optimizer/schedulers/searchers/bayesopt/tuning_algorithms/bo_algorithm_components.py
mseeger: Is this something we can refactor in the base class?

mseeger:
There is a very weird error in integ test. It should not even attempt to download the YAHPO data. Maybe Wes can help

syne_tune/optimizer/schedulers/multiobjective/random_scalarization.py
mseeger: As you prefer, just we'd like to control the seed.

syne_tune/optimizer/schedulers/searchers/bayesopt/tuning_algorithms/bo_algorithm_components.py
jgolebiowski: I think its good the base class checks the metric it wants ot optimize is present, it helps catch errors at construction of the optimizer. The NoOptimization dos not need any construction so its ok. I think the fix would be to adapt th BaesianOptimization constructor (create_internal) to assign optimization primitives separately to every metric. I am happy to look into it after this PR but did not want to ballon this change.

examples/launch_asha_yahpo.py
```diff
@@ -71,7 +71,7 @@ class BenchmarkInfo:
     resource_attr: str
 
 
-if __name__ == "__main__":
```
jgolebiowski, line 74: This disables the example in our integ tests, see https://github.com/awslabs/syne-tune/issues/714 for details on fixing.

mseeger: I'd prefer if you edited the github action to disable this one, and not change the script here.

examples/launch_mb_mo_optimization.py
```diff
@@ -0,0 +1,112 @@
...
+            mode=mode,
+            search_options=search_options,
+        )
+        for metric, mode in zip(metrics, modes)
+    }
+    searcher = MultiObjectiveMultiSurrogateSearcher(
+        config_space=config_space,
+        metric=metrics,
+        estimators=myestimators,
+        scoring_class_and_args=partial(
```
mseeger, line 82: I think this just has to be a tuple, partial is not needed

syne_tune/optimizer/schedulers/multiobjective/random_scalarization.py
```diff
@@ -0,0 +1,95 @@
...
+    :param predictor: Surrogate predictor for statistics of predictive distribution
+    :param weights_sampler: Callable that can generate weights for each objective.
+        Once called it will return a dictionary mapping metric name to scalarization weight as
+        {
+            <name of metric 1> : <weight for metric 1>,
+            <name of metric 2> : <weight for metric 2>,
+            ...
+        }
+    :param kappa: Hyperparameter used for the LCM portion of the scoring
+    """
```
mseeger, line 45: Add comment about `random_seed`, and state this is used only if `weights_sampler` is not given

examples/launch_mb_mo_optimization.py
jgolebiowski: I think both can accompish the same goal of passing a constructor with some arguments fixed.
```
class_constr = partial(MultiObjectiveLCBRandomLinearScalarization, random_seed=random_seed)
clas_instance = class_constr(*args, **kwargs)
```
is the same as
```
class_constr = MultiObjectiveLCBRandomLinearScalarization
clas_instance = class_constr(*args, random_seed=random_seed, **kwargs)
```

but the former is more modern and IMHO cleaner.

syne_tune/optimizer/schedulers/multiobjective/random_scalarization.py
jgolebiowski: They use different normalization schemes. But their main point is that this framework is flexible so approaches like this should work very well. I have put it behind a flag and make it optional.

mseeger:
Your use of `partial` is much better than what I did before, namely to allow for a tuple. I'll send a PR to remove this tuple thing, along with a comment to use `partial` instead.

mseeger (approved):

PR #715, opened 2023-06-14
refactor: Remove tuple option for Type[AcquisitionFunction], as this …

mseeger:
…can be done by partial

Follow-up on recent PR by Jacek: Use of `partial` is more elegant.

----

### DIFF
--- a/syne_tune/optimizer/schedulers/multiobjective/multi_surrogate_multi_objective_searcher.py
+++ b/syne_tune/optimizer/schedulers/multiobjective/multi_surrogate_multi_objective_searcher.py
@@ -22,7 +22,7 @@ from syne_tune.optimizer.schedulers.searchers.bayesopt.datatypes.common import (
 )
 from syne_tune.optimizer.schedulers.searchers.bayesopt.models.estimator import Estimator
 from syne_tune.optimizer.schedulers.searchers.bayesopt.tuning_algorithms.base_classes import (
-    ScoringFunctionCreator,
+    ScoringFunctionConstructor,
 )
 from syne_tune.optimizer.schedulers.searchers.bayesopt.tuning_algorithms.defaults import (
     DEFAULT_NUM_INITIAL_CANDIDATES,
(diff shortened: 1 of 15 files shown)


syne_tune/optimizer/schedulers/multiobjective/multi_surrogate_multi_objective_searcher.py
```diff
@@ -75,7 +75,7 @@ def __init__(
         estimators: Dict[str, Estimator],
         mode: Optional[List[str]] = None,
         points_to_evaluate: Optional[List[Dict[str, Any]]] = None,
-        scoring_class_and_args: Optional[ScoringClassAndArgs] = None,
```
jgolebiowski, line 78: I think the right typing is actually: `Callable[[Any], ScoringFunction]` since we are passing a constructor to the scoring function not an instance of the scoring function

jgolebiowski (changes requested):
Very nice change, thanks for doing it. One comment I have is that I beliebve the right typing is actually: `Callable[[Any], ScoringFunction]` since we pass a constructor which will generate an instance of the ScoringFunction

syne_tune/optimizer/schedulers/multiobjective/multi_surrogate_multi_objective_searcher.py
mseeger: I see, good point

syne_tune/optimizer/schedulers/searchers/bayesopt/tuning_algorithms/base_classes.py
```diff
@@ -260,22 +260,10 @@ def score(
...
-AcquisitionClassAndArgs = Union[
-    Type[AcquisitionFunction], Tuple[Type[AcquisitionFunction], Dict[str, Any]]
-]
+AcquisitionFunctionCreator = Callable[[Any], AcquisitionFunction]
 
-ScoringClassAndArgs = Union[
-    Type[ScoringFunction], Tuple[Type[ScoringFunction], Dict[str, Any]]
-]
 
-
```
jgolebiowski, line 271: NIT: `ScoringFunctionConstructor`

jgolebiowski (commented):
Could we change `ScoringFunctionCreator` to `ScoringFunctionConstructor`. In most cases what is passed is actually the vanilla constructor and it will make it clearer.

jgolebiowski (approved):
LGTM

PR #717, opened 2023-06-15
fix: Fix bug in RemoteLauncher with SageMakerBackend

### LINKED ISSUE #716: examples/launch_height_sagemaker_remotely.py does not work if distribute_trials_on_sagemaker = True
**Describe the bug**
The script uses `RemoteLauncher` to launch an experiment remotely, one can select between local and SageMaker backend. The default is local, and the script works.

But if the SageMaker backend is chosen instead, the script fails.

What happens is that while `tuner.trial_backend.tuner_name` is set properly on the host the script is executed, it becomes `None` inside the SageMaker training job. The backend then tries to launch a trial training job of name "0-None", which fails.

`self.tuner_name` is initialized to `None` when `SageMakerBackend` is created, but it set when the `Tuner` object is created. The `RemoteLauncher` serializes the `Tuner` object along with backend and all other dependencies. Just before serialization, the `self.tuner_name` is properly set. But when this object is deserialized on the client side, `self.tuner_name` of the backend is `None`.

**To reproduce**  
2. Run the following command '....'
3. See error

**Expected behavior**  

**Paste the output of the `pip freeze` command below:**

**Desktop (please complete the following information):**  
 - OS: [e.g. MacOS Monterey 12.6.1 / Ubuntu Linux / Windows 11]
 - Syne Tune Version [e.g. v0.3.4]

**Screenshots**  

**Additional context**

mseeger:
Fixes a bug (and adds test) which blocks one of our customers.

Apart from this, I also replaced `logging` with `logger` at several places.

Closes #716 

----

### DIFF
--- a/examples/launch_height_sagemaker_remotely.py
+++ b/examples/launch_height_sagemaker_remotely.py
@@ -15,6 +15,7 @@ This example show how to launch a tuning job that will be executed on Sagemaker 
 """
 import logging
 from pathlib import Path
+from argparse import ArgumentParser
 
 from sagemaker.pytorch import PyTorch
 
@@ -42,6 +43,11 @@ from syne_tune.remote.remote_launcher import RemoteLauncher
 if __name__ == "__main__":
     logging.getLogger().setLevel(logging.INFO)
 
+    parser = ArgumentParser()
+    parser.add_argument("--use_sagemaker_backend", type=int, default=0)
+    args = parser.parse_args()
+    use_sagemaker_backend = bool(args.use_sagemaker_backend)
+
     max_steps = 100
     n_workers = 4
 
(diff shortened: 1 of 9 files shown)


syne_tune/backend/sagemaker_backend/sagemaker_backend.py
```diff
@@ -522,8 +529,12 @@ def delete_checkpoint(self, trial_id: int):
     def set_path(
         self, results_root: Optional[str] = None, tuner_name: Optional[str] = None
     ):
-        # we use the tuner-name to set the checkpoint directory
-        self.tuner_name = tuner_name
+        """
+        For this backend, it is mandatory to call this method passing ``tuner_name``
+        before the backend is used. ``results_root`` is ignored here.
+        """
+        if tuner_name is not None:
```
mseeger, line 536: This is the bugfix. This is called in `remote_main.py`, and it set `self.tuner_name` back to `None`, which is clearly wrong.

examples/launch_height_sagemaker_remotely.py
```diff
@@ -42,6 +43,11 @@
 if __name__ == "__main__":
     logging.getLogger().setLevel(logging.INFO)
 
+    parser = ArgumentParser()
```
mseeger, line 46: Allows to run this example w/o SM backend, so it is tested.

jgolebiowski (approved):
LGTM.

mseeger:
https://github.com/awslabs/syne-tune/actions/runs/5277103643

end-to-end workflow runs

Issue #714, opened 2023-06-14
examples/launch_asha_yahpo.py not working


jgolebiowski:
**Describe the bug**  
examples/launch_asha_yahpo.py fails with the following bug: https://github.com/awslabs/syne-tune/actions/runs/5260727444/jobs/9508020339?pr=711#logs

**To reproduce**  
1. Go to syne tune root
2. Run the following command `python examples/launch_asha_yahpo.py`
3. See error

```
Traceback (most recent call last):
  File "examples/launch_asha_yahpo.py", line 115, in <module>
    plot_yahpo_learning_curves(
  File "examples/launch_asha_yahpo.py", line 36, in plot_yahpo_learning_curves
    bb = trial_backend.blackbox
  File "/home/runner/work/syne-tune/syne-tune/syne_tune/blackbox_repository/simulated_tabular_backend.py", line 335, in blackbox
    self._blackbox = load_blackbox(
  File "/home/runner/work/syne-tune/syne-tune/syne_tune/blackbox_repository/repository.py", line 169, in load_blackbox
    return instantiate_yahpo(name, **yahpo_kwargs)
  File "/home/runner/work/syne-tune/syne-tune/syne_tune/blackbox_repository/conversion_scripts/scripts/yahpo_import.py", line 373, in instantiate_yahpo
    bench = benchmark_set.BenchmarkSet(scenario, active_session=False)
  File "/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/site-packages/yahpo_gym/benchmark_set.py", line 50, in __init__
    self.config_space = self._get_config_space()
  File "/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/site-packages/yahpo_gym/benchmark_set.py", line 371, in _get_config_space
    cs = CS_json.read(json_string)
  File "/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/site-packages/ConfigSpace/read_and_write/json.py", line 444, in read
    _construct_hyperparameter(
  File "/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/site-packages/ConfigSpace/read_and_write/json.py", line 5[24](https://github.com/awslabs/syne-tune/actions/runs/5260727444/jobs/9508020339?pr=711#step:9:25), in _construct_hyperparameter
    q=hyperparameter["q"],
KeyError: 'q'
Error: Process completed with exit code 1.
```

**Expected behavior**  
Complete run

[...]

PR #718, opened 2023-06-15
fix: Removed the broken example from github actions

jgolebiowski:
Helps https://github.com/awslabs/syne-tune/issues/714

----


mseeger (approved):
Could you edit the issue, that once this is fixed, the example is re-introduced here?

PR #719, opened 2023-06-16
refactor: Move code from benchmarking/commons to syne_tune/experiments

mseeger:
This is a major refactoring, but conceptually simple:

* Move most code in `benchmarking/commons` to `syne_tune/experiments`, where it is installed by `pip`
* The code remaining in `benchmarking` concerns the "real benchmarks", so the examples for training code. However, all surrogate benchmarks are available now in `syne_tune/experiments/benchmark_definitions`, because they only rely on the blackbox repo, which is also installed by `pip`
* Small change in `launch_remote.py` codes: Argument `source_dependencies` allows to pass dependencies for the SM estimator other than `syne_tune` itself. Often, these extra source deps are `benchmarking.__path__` (as before), but for users who have not installed the sources, they can put their source deps here, and it should work

I added comments to where a real change happens. Everything else is moving files, renaming imports, and changing docs.

Testing:

* All benchmarking examples still work
* `benchmarking/nursery/demo_experiment` is not using `benchmarking.__path__` as source dependence, and uses local imports. This works fine, so this means `benchmarking` is not needed
* What I could not test is whether this all works when Syne Tune is installed from `pip`, because the new code is not there yet

----


syne_tune/experiments/launchers/hpo_main_sagemaker.py
```diff
@@ -146,7 +147,6 @@ def start_benchmark_sagemaker_backend(
     )
     del sm_args["checkpoint_s3_uri"]
     sm_args["sagemaker_session"] = default_sagemaker_session()
-    sm_args["dependencies"] = benchmarking.__path__
```
mseeger, line 149: This is replaced by `source_dependencies`, as the user may not even have `benchmarking` (if not installed frmo source).

syne_tune/experiments/launchers/launch_remote_common.py
```diff
@@ -32,6 +31,7 @@ def sagemaker_estimator_args(
     tuner_name: str,
     benchmark: Optional[BenchmarkDefinition] = None,
     sagemaker_backend: bool = False,
+    source_dependencies: Optional[List[str]] = None,
```
mseeger, line 34: Look at this: New `source_dependencies`

syne_tune/experiments/launchers/launch_remote_common.py
```diff
@@ -56,13 +59,16 @@ def sagemaker_estimator_args(
...
+    if source_dependencies is not None:
+        dependencies = dependencies + source_dependencies
     sm_args = dict(
         entry_point=entry_point.name,
         source_dir=str(entry_point.parent),
         checkpoint_s3_uri=checkpoint_s3_uri,
         max_run=max_run,
         role=get_execution_role(),
-        dependencies=syne_tune.__path__ + benchmarking.__path__,
+        dependencies=dependencies,
```
mseeger, line 71: Needed, because we want people to use their own source deps, even if `benchmarking` is not even there.

mseeger:
That file is there in the new branch here, but of course not in what is currently released.

mseeger:
OK, the trick is to avoid sphinx 7.0.0.

syne_tune/experiments/launchers/utils.py
```diff
@@ -13,7 +13,7 @@
 from pathlib import Path
 from typing import Optional
 
-from syne_tune.experiments.results_utils import sync_from_s3_command
+from syne_tune.experiments.visualization.results_utils import sync_from_s3_command
```
wesk, line 16: Should `sync_from_s3_command` be under visualization?

syne_tune/experiments/launchers/launch_remote_local.py
```diff
@@ -18,25 +18,25 @@
...
+from syne_tune.experiments.benchmark_definitions import RealBenchmarkDefinition
+from syne_tune.experiments.launchers.hpo_main_common import (
     extra_metadata,
     ExtraArgsType,
     ConfigDict,
     DictStrKey,
     config_from_argparse,
 )
-from benchmarking.commons.hpo_main_local import (
+from syne_tune.experiments.launchers.hpo_main_local import (
```
wesk, line 30: It is a bit counterintuitive that benchmarking logic (for example RealBenchmarkDefinitions) is available inside `launchers`. Is it possible to decouple it further?

wesk:
Regarding:
> The code remaining in benchmarking concerns the "real benchmarks", so the examples for training code. However, all surrogate benchmarks are available now in syne_tune/experiments/benchmark_definitions, because they only rely on the blackbox repo, which is also installed by pip

One advantage of the status-quo, from a benchmarking point of view, is that all of the benchmarking code is together under `benchmarking`. However, this change as it stands now splits benchmarking into multiple places. 

Which leads me to the question of, what would it look like to keep all of the benchmarking logic where it is, and just move the experimentation logic?

My concern is that moving part of the benchmarking logic may create confusion, so better to either move all of it, or none of it.

docs/source/getting_started.rst
```diff
@@ -64,6 +64,16 @@ This example runs `ASHA <tutorials/multifidelity/mf_asha.html>`__ with
...
 
+Experimentation with Syne Tune
+==============================
+
+If you plan to use advanced features of Syne Tune, such as different execution
+backends or running experiments remotely, writing launcher scripts like
+``examples/launch_height_simple.py`` can become tedious. Syne Tune provides an
+advanced experimentation framework, which you can learn about in
+`this tutorial <tutorials/experimentation/README.html>`__, or also in
+`this one <tutorials/odsc_tutorial/README.html>`__.
```
wesk, line 75: Looking at `examples/launch_height_simple.py`, I see it uses LocalBackend: https://github.com/awslabs/syne-tune/blob/main/examples/launch_height_simple.py#L16. Isn't that in fact a good reference example? Or is there an alternative approach that we recommend instead?

Do any of the examples in the `examples/` folder that show usage of the experiment launchers? 

Maybe a more general question - what do these experimentation launchers provide that the backends themselves don't already provide directly?  https://github.com/awslabs/syne-tune/tree/main/syne_tune/backend

And would anything stop us from adding those features to the backends directly, rather than creating this new folder under `syne_tune.experiments.launchers.` and putting launcher scripts there.

Otherwise, it seems like there are two competing ways of doing things. And from a user experience point of view, it may not be clear what the recommended path is.

wesk:
I would recommend a more incremental approach, where we define what the problem is with the status-quo, and then make the changes that are necessary in order to fix those issues we identify.

syne_tune/experiments/launchers/utils.py
mseeger: Moved to `launchers/utils`

syne_tune/experiments/benchmark_definitions/fcnet.py
```diff
@@ -10,7 +10,7 @@
 # on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
-from benchmarking.commons.benchmark_definitions.common import (
+from syne_tune.experiments.benchmark_definitions.common import (
```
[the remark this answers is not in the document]
mseeger: One of the main reasons of this refactoring is that most folks do not use this very useful methodology, because they somehow think this is only useful for benchmarking (i.e., comparing different HPO methods).

I kind of thought of a "benchmark" as an HPO problem to be addressed, but since this term seems to trigger the wrong association, I'll think of another one.

docs/source/getting_started.rst
mseeger: I don't really understand this.

mseeger: The idea of `syne_tune.experiments.launchers` is for users not to have to write complex launcher scripts for every case. This has nothing to do with backends, which are just a part of the definition of a tuning experiment.

mseeger: There are two competing ways of doing things. One is most flexible, but very hard to run complicated experiments. The other is much simpler and allows to quickly obtain simple launcher scripts with CLI.

mseeger: The launcher scripts in `examples/` are bare bones, and error prone. You see how often things are wrong with them. They are not useful to run many experiments. They are basically using the essentials, which is OK, but tedious for most users.

The stuff in `syne_tune/experiments` is for users who do not care to create a scheduler with a searcher and a backend, and make sure they all fit together with their training function. Each backend has its own ways in which you need to create them, it is quite complex. With `experiments`, this is independent of internals, you just specify things about your training function.

mseeger: I am a bit tired of these discussions. Please have a look at Optuna, where they provide simple ways of describing a study and then execute it. We should always strive for simplicity.

mseeger:
OK, so the reason that nobody except for me uses all this code, is that everybody thinks this is just about "benchmarking". This is why people like Vittorio take ages to get stuff going in Syne Tune, wasting their time writing launcher scripts, and have to use legacy code like `RemoteLauncher` (which serializes a tuner with all dependencies for no reasons -- the new mechanism does not need any of that).

mseeger:
The only way out is to not call this stuff "benchmarking" anymore. There is no incremental path to this, and as we saw both with Vittorio, but also with the ICML paper, and with the MO experiments towards my talk slides: Nobody except for me is using "benchmarking". So there won't be confusions renaming and moving it.

I had to spend quite a bit of work addressing issues, which would all be avoided if people used the code I already provided. It is fair enough to make this code easier to access.

mseeger:
@wesk It is not true that all the code required for benchmarking is in `benchmarking`. This code of course depends in many ways on `syne_tune`, and code for plotting in `syne_tune/experiments` is specific for results coming out of `benchmarking`.

mseeger:
Moving code for experimentation to `syne_tune` has a clear rationale: It becomes available when people do `pip install` only. This is what Vittorio needs, to avoid having to depend on legacy code.

With the same arguments, I could also say let's move `RemoteLauncher` to `benchmarking`, because my feeling is it does not belong to `syne_tune`.

mseeger:
What I'll do. Since "benchmark" is apparently a fixed term, I'll change it to "tuning problem". A "tuning problem" is a training/evaluation script plus some meta-data, providing defaults for config space and others.

"Benchmarking" will then be running several HPO methods on one or more "tuning problems".

mseeger:
The reason for tuning problems for surrogate benchmarks to be in `syne_tune` is the same as for `syne_tune/blackbox_repository`. Either, these surrogate blackboxes are only useful for benchmarking. Then, blackbox repository should move there and not installed by `pip`. Or they are generally useful, for example for testing and experimentation. Then, everything should be in `syne_tune`.

mseeger:
@wesk I tried to replace "benchmark" by "problem" everywhere, but in the end, this is a massive change for little real gain, except maybe it would make people start using this if they have no interest in benchmarking HPO methods at all.

mseeger:
It's been a historical mistake to call a tuning problem a "benchmark", but instead of trying to correct it, I suggest to make it very clear in all our docs that we simply call a tuning problem a "benchmark":

* Even if you do not care a bit about "benchmarking", the experimentation setup is useful for you, and in order to use it, you need to specify defaults for your training script in something called a "benchmark"

mseeger:
But just to be clear. With this PR, I want to get out of this place between a rock and a hard place, where I feel a lot of my efforts have been put in:

* Oh, so this is called "benchmarking"? So it must be in `benchmarking/`, cannot "pollute" Syne Tune (while `blackbox_repository` and `RemoteLauncher` are in Syne Tune)
* Oh, so this is not in `syne_tune/` and anyway it is called "benchmarking"? I cannot / will not use it for my experimentations. All these tutorials are nice, but why read them if this is just for comparing HPO methods?

syne_tune/experiments/benchmark_definitions/fcnet.py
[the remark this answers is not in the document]
mseeger: Well no, I will keep this term, it is too ingrained. But I'll make it clear that a "benchmark" is nothing but a tuning problem.

syne_tune/experiments/launchers/launch_remote_local.py
mseeger: It is not in `launchers`, but in `benchmark_definitions`

mseeger:
@wesk In the end, let us not forget that the changes here make life easier (and more stable) for Vittorio. I am not going to debug `RemoteLauncher` any much longer, or debug complex launcher scripts of customers. I'd like to point them to something simple, not just for "benchmarking"

wesk:
You've made great points! Thanks for the additional context. 

I agree that the fact that the code is all currently under `benchmarking/` really limits who is using it. 

Whatever we can do to make it easier to use Syne Tune is worthwhile.

I'm looking more at this again now and will follow-up

mseeger:
About "either moving all of benchmarking logic or none of it".
The separation is between (a) logic and (b) content. We decided a while ago that examples and training scripts for them should not be part of the pip installed code (e.g., `examples/`, `benchmarking/`).

The point about this PR is to move the logic (a) to `syne_tune/experiments`, but keep the content (b) where it is.

The exception are the surrogate benchmarks. They are moved, because `syne_tune/blackbox_repository` is also pip installed, so this part of the content is already there, and we should make it accessible.

docs/source/tutorials/multifidelity/README.rst
```diff
@@ -8,7 +8,7 @@ between different methods, and how to choose the best approach for your own
 problems.
 
 .. note::
-   This tutorial makes use of the *benchmarking* framework, you need to have
+   In order to run the code in this tutorial, you need to have
    `installed Syne Tune from source <../../getting_started.html#installation>`__
    to run it.
```
wesk, line 13: What components does this tutorial still need that requires it to be installed from source? 

(Fine to leave it like this, but want to verify that indeed it still needs to be installed from source.)

docs/source/tutorials/odsc_tutorial/getting_started.rst
```diff
@@ -170,3 +170,9 @@ The most important fields are:
 Also, note the role of ``**kwargs`` in the function signature, which allows
 to overwrite any of the default values (e.g., for ``max_wallclock_time``,
 ``n_workers``, or ``instance_type``) with command line arguments.
+
+.. note::
+   In the Syne Tune experimentation framework, a tuning problem (i.e., training and
+   evaluation script together with defaults) is called a *benchmark*.
+   This terminology is used even if the goal of experimentation is not benchmarking
+   (i.e., comparing different HPO methods), as is the case in this tutorial here.
```
wesk, line 178: After this refactor is completed, what will remain in the top-level `benchmarking/` folder, and what is the purpose of it? (aside from `benchmarking/nursery`) 

And a question arises: what is the difference between a benchmark and a study?

My understanding of existing terminology: 
- `experiment`: a single run of HPO
- `study`: a set of experiments (e.g., different HPO methods, different config spaces, different seeds)

By the existing naming scheme, is a benchmark an implementation of a study?

docs/source/tutorials/multifidelity/README.rst
mseeger: Good catch, this is wrong.

mseeger: I'll go over all tutorials and check.

benchmarking/benchmark_definitions/resnet_cifar10.py
```diff
@@ -25,39 +25,28 @@
...
+        script=Path(__file__).parent.parent
         / "training_scripts"
         / "resnet_cifar10"
         / "resnet_cifar10.py",
         config_space=config_space,
         max_wallclock_time=3 * 3600,
-        n_workers=params["num_workers"],
-        instance_type=params["instance_type"],
+        n_workers=4,
+        instance_type=instance_type,
```
wesk, line 49: Can these continue to be specified through `params` or do they need to be hardcoded?

benchmarking/benchmark_definitions/mlp_on_fashionmnist.py
```diff
@@ -22,33 +21,22 @@
...
+        script=Path(__file__).parent.parent
         / "training_scripts"
         / "mlp_on_fashion_mnist"
         / "mlp_on_fashion_mnist.py",
         config_space=config_space,
         max_wallclock_time=3 * 3600,  # TODO
-        n_workers=params["num_workers"],
-        instance_type=params["instance_type"],
+        n_workers=4,
+        instance_type=DEFAULT_CPU_INSTANCE,
```
wesk, line 39: Likewise here, should these still be set through `params`?

wesk: A few other instances of similar changes in `benchmarking/benchmark_definitions/lstm_wikitext2.py` and `benchmarking/benchmark_definitions/distilbert_on_imdb.py`

docs/source/tutorials/multifidelity/README.rst
mseeger: I'll also make sure that all benchmarking examples which do not depend on training code in `benchmarking`, use local imports, which means that only their own code has to be installed, not all of `benchmarking`.

docs/source/tutorials/odsc_tutorial/getting_started.rst
mseeger: There are two purposes of `benchmarking` folder:

* Collect training codes for benchmarks (in `benchmarking/training_scripts`)
* Collect examples for benchmarking studies (in `benchmarking/examples`, `benchmarking/nursery`)

mseeger: Naming:

* A "benchmark" is a tuning problem, consisting of a training code or surrogate blackbox, plus a default config space
* An experiment is a single run of HPO, configured by a benchmark (tuning problem), a method, a seed, etc
* A study is a set of experiments, designed to answer some statistical question or gather insights
* Some studies are about "benchmarking",  meaning comparing how different HPO methods compare on one or more benchmarks
* Other studies are about something else, so for example use a single HPO method only

benchmarking/benchmark_definitions/resnet_cifar10.py
mseeger: These are just defaults, you can change them on the CL. The refactoring here is only to go from two functions to one, this is legacy stuff.

benchmarking/benchmark_definitions/mlp_on_fashionmnist.py
mseeger: See above.

wesk (approved):

PR #720, opened 2023-06-18
fix: Avoid sphinx 7.0.0, since the readthedocs build fails with that

mseeger:
You could investigate more, but fact is our docs do not currently build on readthedocs, and downgrading sphinx to <7.0.0 solves that issue. This is what I found:

https://github.com/readthedocs/readthedocs.org/issues/10279

----


wesk (approved):
Thanks! Will keep an eye on https://github.com/readthedocs/sphinx_rtd_theme/issues/1463

PR #727, opened 2023-06-26
refactor: Wrappers for multi-objective methods

mseeger:
I got a request on what we offer as multi-objective methods. It is really important for adoption that when we implement something new, we add wrappers and update the docs.

----

### DIFF
--- a/syne_tune/optimizer/baselines.py
+++ b/syne_tune/optimizer/baselines.py
@@ -26,6 +26,7 @@ from syne_tune.optimizer.schedulers.multiobjective import (
     NSGA2Searcher,
     MultiObjectiveMultiSurrogateSearcher,
     MultiObjectiveLCBRandomLinearScalarization,
+    LinearScalarizedScheduler,
 )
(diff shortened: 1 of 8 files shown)


syne_tune/optimizer/baselines.py
```diff
@@ -733,6 +738,142 @@ def __init__(
...
+            config_space=config_space,
+            metric=metric,
+            mode=mode,
+            searcher=searcher,
+            random_seed=random_seed,
+            **kwargs,
+        )
+
+
+class NSGA2(FIFOScheduler):
```
aaronkl, line 840: Has anything changed here? Because this wrapper is already in` baselines.py`

aaronkl (approved):

syne_tune/optimizer/baselines.py
mseeger: Do you mean `NSGA2`? Yes, that is already there. I think diff is messing things up.

PR #728, opened 2023-06-26
fix: ExperimentResult.best_config() always used max mode

ystein:
## Problem
* In [ExperimentResult.best_config()](https://github.com/awslabs/syne-tune/blob/254ad0708703245e51da0b113f1067a2da92039a/syne_tune/experiments/experiment_result.py#L181C26-L181C26) `metric_mode` is resolved by calling `_metric_name_mode`
* In [_metric_name_mode](https://github.com/awslabs/syne-tune/blob/254ad0708703245e51da0b113f1067a2da92039a/syne_tune/experiments/experiment_result.py#L223-L225): in case `metric_mode` has a non-zero length it is replaced by `metric_mode[metric]`. In case `metric_mode` is a string (for example `min`) and not a list, it is replaced by the character in the string at position `metric`, with `metric` defaulting to `0`. In case of `metric_mode = "min"` it is replaced by `m`.
* back in [best_config](https://github.com/awslabs/syne-tune/blob/254ad0708703245e51da0b113f1067a2da92039a/syne_tune/experiments/experiment_result.py#L184), we test whether `metric_mode` equals `min` and otherwise go into the `max` case. In case `metric_mode` was originally a string `"min"` it got replaced by `"m"` and we falsely return the trial that maximized the metric.

----

### DIFF
--- a/syne_tune/experiments/experiment_result.py
+++ b/syne_tune/experiments/experiment_result.py
@@ -221,7 +221,7 @@ class ExperimentResult:
             )
 
         metric_mode = self.metric_mode()
-        if len(metric_mode) > 1:
+        if isinstance(metric_mode, list):
             metric_mode = metric_mode[metric]
 
         return metric, metric_name, metric_mode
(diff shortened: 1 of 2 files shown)


mseeger (approved):
Thanks!

mseeger (approved):

PR #734, opened 2023-06-29
fix: Clean up imports so that code runs with core dependencies. Updat…

mseeger:
…e tests to enforce this

Need to be more careful with imports in the future!

----


mseeger:
There are still some `end-to-end-tests` failing

mseeger:
OK, `end-to-end-tests` are working now

.github/workflows/run-syne-tune.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           python -m pip install --upgrade pip
       - name: Install Syne Tune with all dependencies
```
wesk, line 56: Nit
```suggestion
      - name: Install Syne Tune with core dependencies only
```

wesk (approved):
Nice change, and the stricter tests will help prevent this from happening again.

I verified that with this change it's now possible to run `from syne_tune import Tuner` with just core dependencies.

wesk (approved):

PR #741, opened 2023-07-08
feat: Automatic streamlining of configuration space

### LINKED ISSUE #380: Auto-suggest domains with better encoding
Syne Tune provides a number of data types for hyperparameters:

https://github.com/awslabs/syne-tune/blob/main/syne_tune/config_space.py

Some of them are for finite ranges: `choice`, `finrange`, `logfinrange`, `ordinal`, `logordinal`.

For numerical values, users often choose `choice` even though one of the others would be more appropriate, and lead to better results with Bayesian optimization.

The goal of this task is to provide recommendations when `choice` is used, on what could be a better type. For now, this would simply just be printed, but in the future. Upon call of `choice(categories)`, the recommender would check whether the value type is int or float. If so:
- categories not sorted increasing order? Print message
- If sorted increasing order? Check whether equispaced. If so, recommend `finrange`. Check whether equispaced in log. If so, recommend `logfinrange`
- Advanced. If increasing, but not equispaced. Recommend `ordinal(kind="nn")` or `logordinal`, depending on spacing closer to equal in normal or log domain. Also `logordinal` requires positive values

This task does not require internal knowledge about Syne Tune, and it would simplify adoption.

mseeger:
This is automating some best practices for choosing domains in ``config_space``. Particularly useful for people who come from other tools, like Optuna, where you don't have the finite numerical types we have.

Closes #380.

----

### DIFF
--- a/benchmarking/training_scripts/mlp_on_fashion_mnist/mlp_on_fashion_mnist.py
+++ b/benchmarking/training_scripts/mlp_on_fashion_mnist/mlp_on_fashion_mnist.py
@@ -35,7 +35,13 @@ except ImportError:
     )
 
 from syne_tune import Reporter
-from syne_tune.config_space import randint, uniform, loguniform, add_to_argparse
+from syne_tune.config_space import (
+    randint,
+    lograndint,
+    uniform,
+    loguniform,
+    add_to_argparse,
+)
 from syne_tune.utils import (
     resume_from_checkpointed_model,
     checkpoint_model_at_rung_level,
@@ -57,8 +63,8 @@ ELAPSED_TIME_ATTR = "elapsed_time"
 
 
 _config_space = {
-    NUM_UNITS_1: randint(4, 1024),
-    NUM_UNITS_2: randint(4, 1024),
+    NUM_UNITS_1: lograndint(4, 1024),
+    NUM_UNITS_2: lograndint(4, 1024),
     "batch_size": randint(8, 128),
     "dropout_1": uniform(0, 0.99),
     "dropout_2": uniform(0, 0.99),
--- a/docs/source/search_space.rst
+++ b/docs/source/search_space.rst
@@ -103,7 +103,7 @@ Recommendations
 
 How to choose the domain for a given hyperparameter? Obviously, we want to
 avoid illegal values: learning rates should be positive, probabilities lie
-in ``[0, 1]``, and numbers of unit must be integers. Apart from this, the
+in ``[0, 1]``, and numbers of units must be integers. Apart from this, the
 choice of domain is not always obvious, and different choices can affect
 search performance significantly in some cases.
 
--- a/docs/source/search_space.rst
+++ b/docs/source/search_space.rst
@@ -4,7 +4,10 @@ How to Choose a Configuration Space
 One important step in applying hyperparameter optimization to your tuning
 problem is to define a configuration space (or search space). Doing this
 optimally for any given problem is more of an art than a science, but in this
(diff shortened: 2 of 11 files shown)


benchmarking/training_scripts/mlp_on_fashion_mnist/mlp_on_fashion_mnist.py
```diff
@@ -57,8 +63,8 @@
 
 
 _config_space = {
-    NUM_UNITS_1: randint(4, 1024),
-    NUM_UNITS_2: randint(4, 1024),
+    NUM_UNITS_1: lograndint(4, 1024),
```
mseeger, line 66: Made this change, because our own replacement logic maps `randint` to `lograndint` when `upper >= 100 * lower`.

docs/source/search_space.rst
```diff
@@ -100,9 +103,39 @@ Recommendations
 
 How to choose the domain for a given hyperparameter? Obviously, we want to
 avoid illegal values: learning rates should be positive, probabilities lie
-in ``[0, 1]``. Apart from this, the choice of domain is not always obvious,
-and different choices can affect search performance significantly in some
-cases. Here, we provide some recommendations:
+in ``[0, 1]``, and numbers of unit must be integers. Apart from this, the
```
aaronkl, line 106: typo: units

syne_tune/utils/convert_domain.py
```diff
@@ -0,0 +1,211 @@
...
+    converted to :func:`~syne_tune.config_space.finrange`,
+    :func:`~syne_tune.config_space.logfinrange`,
+    :func:`~syne_tune.config_space.ordinal`, or
+    :func:`~syne_tune.config_space.logordinal`. Otherwise, ``domain`` is
+    returned as is.
+    """
+    num_values = len(domain)
+    if num_values <= 2:
+        return domain
+    if not isinstance(domain.categories[0], Real):
```
aaronkl, line 89: Why do we only check the first category to be Real?

syne_tune/utils/convert_domain.py
```diff
@@ -0,0 +1,211 @@
...
+    lograndint,
+)
+from syne_tune.util import is_integer
+
+logger = logging.getLogger(__name__)
+
+
+def fit_to_regular_grid(x: np.ndarray) -> Dict[str, float]:
+    r"""
+    Computes the least squares fit of :math:`a * j + b` to ``x[j]``, where
```
aaronkl, line 42: What is x[j] here?

syne_tune/utils/convert_domain.py
```diff
@@ -0,0 +1,211 @@
...
+    if num_values <= 2:
+        return domain
+    if not isinstance(domain.categories[0], Real):
+        return domain
+    values_are_int = is_integer(domain.categories)
+    sorted_values = sorted(domain.categories)
+    x = np.array(sorted_values)
+    best_fit = fit_to_regular_grid(x)
+    best_is_log = False
+    if sorted_values[0] >= POSITIVE_EPS:
```
aaronkl, line 96: Can you add some comments to better explain what is happening here?

syne_tune/utils/convert_domain.py
mseeger: The j-th entry of array x

syne_tune/utils/convert_domain.py
aaronkl: Maybe add a comment to make this clear?

aaronkl (approved):

PR #756, opened 2023-09-06
ci: automatically test all notebooks under `examples/notebooks` and render them in docs page

wesk:
This PR: 
1. Adds automation to test notebooks with every pull request
2. Adds automation to render notebooks in our [documentation website](https://syne-tune.readthedocs.io/en/latest/examples_toc.html)
3. Adds a new example notebook

More detail on the three aspects below.

## Automatically test notebooks with every pull request

To add a new Jupyter (.ipynb) file to the repository, simply add it to the `examples/notebooks` folder. 

Every notebook in the `examples/notebooks` folder will be converted to a python script and executed by a GitHub Actions workflow. 

If you don't want specific cells in a notebook to be run as part of our automatic tests, tag  them with the `ci-skip-cell` tag, and the CI will not execute the cell. The reason this feature exists is because some cells that are used to install dependencies, such as `!pip install xgboost` do not convert easily into directly executable scripts. The first code cell in the XGBoost notebook is the motivating example for this feature. 

## Automatically render notebooks in Syne Tune documentation

The idea with this PR is to keep a single source of truth: `examples/notebooks` as the location where all example notebooks are kept.

To have your notebook show up in documentation, first add it to the `examples/notebooks` folder, and then update the following syntax inside the `examples_toc.rst` page to make sure that your notebook is included in the documentation.

Example showing adding the xgboost notebook:

```
Examples
========

.. toctree::
   :name: Example Notebooks
   :caption: Example Notebooks
   :maxdepth: 1

   notebooks/xgboost_example.ipynb


.. toctree::
   :name: Example Tuning Scripts
   :caption: Example Tuning Scripts
   :maxdepth: 1

   examples

```

Before you add a new notebook, be sure to clear the cell outputs, unless you also want them to render in the resulting documentation.

Note that the documentation build process copies the notebooks to the `docs/source/notebooks` folder. This is because the [nbsphinx extension](https://nbsphinx.readthedocs.io/en/0.9.3/) only knows about files inside the `docs/source/` folder.

## New example notebook: `Tune XGBoost`

This PR adds a new example notebook demonstrating how you can use Syne Tune to tune XGBoost's hyperparameters on the UCI hand-written digits dataset.

----

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -21,6 +21,16 @@ import syne_tune
 sys.path.insert(0, os.path.abspath("../../"))
 
 
+def copy_notebooks_into_docs_folder(app):
+    # .ipynb files must be inside the docs/ folder for Jupyter to be able to convert them
+    destination = "source/notebooks"
+    try:
+        shutil.rmtree(destination)
+    except OSError as e:
+        print("Error: %s - %s." % (e.filename, e.strerror))
+    shutil.copytree("../examples/notebooks", destination)
+
+
 def run_apidoc(app):
     """Generate doc stubs using sphinx-apidoc."""
     module_dir = os.path.join(app.srcdir, "../../")
@@ -61,6 +71,7 @@ def run_apidoc(app):
 
 def setup(app):
     """Register our sphinx-apidoc hook."""
+    app.connect("builder-inited", copy_notebooks_into_docs_folder)
     app.connect("builder-inited", run_apidoc)
 
 
@@ -110,6 +121,7 @@ default_role = "py:obj"
 templates_path = ["_templates"]
 exclude_patterns = []
 
+nbsphinx_execute = 'never'
 
 # -- Options for HTML output -------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-output
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -92,6 +92,7 @@ extensions = [
     "sphinxcontrib.bibtex",
     "myst_parser",
     "sphinxcontrib.jquery",  # can be removed as soon as the theme no longer depends on jQuery
+    "nbsphinx",
 ]
 
 myst_heading_anchors = 2
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -26,7 +26,9 @@ def copy_notebooks_into_docs_folder(app):
     currpath = os.path.dirname(os.path.realpath(__file__))
     source = os.path.join(currpath, "../../examples/notebooks")
     destination = os.path.join(currpath, "notebooks")
-    print(f"current path: {currpath}; source path: {source}; destination path: {destination}")
+    print(
+        f"current path: {currpath}; source path: {source}; destination path: {destination}"
+    )
     try:
         shutil.rmtree(destination)
     except OSError as e:
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -14,6 +14,7 @@ import datetime
 import os
 import shutil
 import sys
+from pathlib import Path
 
 import syne_tune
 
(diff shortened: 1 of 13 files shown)


docs/source/conf.py
```diff
@@ -21,6 +21,21 @@
 sys.path.insert(0, os.path.abspath("../../"))
 
 
+def copy_notebooks_into_docs_folder(app):
+    # .ipynb files must be inside the docs/ folder for Jupyter to be able to convert them
+    currpath = os.path.dirname(os.path.realpath(__file__))
+    source = os.path.join(currpath, "../../examples/notebooks")
```
geoalgo, line 27: Can we use `Pathlib` `parent` rather than "../../" which is not portable?

docs/source/conf.py
```diff
@@ -21,6 +21,21 @@
...
+    destination = os.path.join(currpath, "notebooks")
+    print(
+        f"current path: {currpath}; source path: {source}; destination path: {destination}"
+    )
+    try:
+        shutil.rmtree(destination)
```
geoalgo, line 33: Do we need this `rmtree`? It seems potentially dangerous, most of the time we would execute on the remote machine but I could see a case where we would execute locally and may get bad surprise.

geoalgo (commented):
It looks good to me overall. 

One question would be what is the rationale for having a notebook and not an example? 

I think there should be a clear way to say how functionalities are committed otherwise you may have a mix-and-match of notebook/python examples (for instance a bunch of the current examples could be notebooks).

examples/notebooks/tune_xgboost.ipynb
```diff
@@ -0,0 +1,180 @@
...
+    "scheduler = BayesianOptimization(\n",
+    "    config_space,\n",
+    "    metric=\"accuracy\",\n",
+    "    mode=\"max\"\n",
+    ")\n",
+    "\n",
+    "tuner = Tuner(\n",
+    "    trial_backend=PythonBackend(tune_function=train, config_space=config_space),\n",
+    "    scheduler=scheduler,\n",
+    "    stop_criterion=StoppingCriterion(max_wallclock_time=30),\n",
```
mseeger, line 131: This is pretty short, in case you give this script to MLU

examples/notebooks/tune_xgboost.ipynb
```diff
@@ -0,0 +1,180 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Hyperparameter configuration space\n",
+    "config_space = {\n",
+    "    \"max_depth\": randint(1,10),\n",
+    "    \"gamma\": uniform(1,10),\n",
+    "    \"reg_lambda\": uniform(.0000001, 1),\n",
```
mseeger, line 117: Use `loguniform` instead of `uniform`

examples/notebooks/tune_xgboost.ipynb
```diff
@@ -0,0 +1,180 @@
...
+    "    For more information about the dataset, see https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_digits.html\n",
+    "    '''\n",
+    "\n",
+    "    from sklearn.datasets import load_digits\n",
+    "    from sklearn.model_selection import train_test_split\n",
+    "    from syne_tune import Reporter\n",
+    "    import xgboost\n",
+    "    import numpy as np\n",
+    "\n",
+    "    eval_metric = \"merror\"\n",
```
mseeger, line 75: Is this needed?

examples/notebooks/tune_xgboost.ipynb
```diff
@@ -0,0 +1,180 @@
...
+    "    report = Reporter()\n",
+    "\n",
+    "    clf = xgboost.XGBClassifier(\n",
+    "        n_estimators=n_estimators,\n",
+    "        reg_lambda=reg_lambda,\n",
+    "        gamma=gamma,\n",
+    "        max_depth=max_depth,\n",
+    "        eval_metric=eval_metric,\n",
+    "    )\n",
+    "    clf.fit(X_train, y_train, eval_set=[(X_val, y_val)])\n",
```
mseeger, line 91: Is the `eval_set` parameter needed?

mseeger (approved):
I just have some small comments.

examples/notebooks/tune_xgboost.ipynb
wesk: No - removed in next revision

examples/notebooks/tune_xgboost.ipynb
wesk: Updated

docs/source/conf.py
wesk: Good point, I updated it to `shutil.copytree(source, destination, dirs_exist_ok=True)`, the `dirs_exist_ok=True` is what I was missing previously. 

https://docs.python.org/3/whatsnew/3.8.html#shutil

docs/source/conf.py
wesk: Updated

wesk:
> It looks good to me overall.
> 
> One question would be what is the rationale for having a notebook and not an example?
> 
> I think there should be a clear way to say how functionalities are committed otherwise you may have a mix-and-match of notebook/python examples (for instance a bunch of the current examples could be notebooks).

Thanks for the feedback -- I updated `CONTRIBUTING.md` with some general guidance in the next revision. 

I intentionally kept the guidance general - we can update/refine it over time. 

Another consideration is that not every example can easily be run today as a notebook.

My hypothesis is that users, especially those who are new to Syne Tune, will appreciate notebook examples, and we may  want to consider converting some existing script examples into notebooks over time.

geoalgo (approved):

PR #765, opened 2023-10-03
feat: Add SMAC wrapper and examples

geoalgo:
Adds SMAC to syne tune using their ask & tell interface. I still need to benchmark the method but this would allow to easily share the wrapper and discuss/improve it with the author.

----


geoalgo:
The issue remaining comes from a problem with yahpo which is incompatible with a recent onnx change: https://github.com/slds-lmu/yahpo_gym/issues/82

examples/launch_smac_example.py
```diff
@@ -0,0 +1,115 @@
...
+    max_resource_attr = benchmark.max_resource_attr
+    trial_backend = BlackboxRepositoryBackend(
+        elapsed_time_attr=benchmark.elapsed_time_attr,
+        max_resource_attr=max_resource_attr,
+        blackbox_name=benchmark.blackbox_name,
+        dataset=dataset_name,
+    )
+
+    # Asynchronous successive halving (ASHA)
+    blackbox = trial_backend.blackbox
```
wesk, line 87: Is this comment accurate? That ASHA is somehow being used in this example?

wesk (approved):
Looks good!

PR #767, opened 2023-10-09
feat: Add method to get the best configuration directly from Tuner, add com…

geoalgo:
…ments on how to rerun the best configuration found

Also updates `metric_name_mode` to:
1) test the logic as it is quite tricky with union types
2) stop returning `metric` argument which is an input
3) don’t pass `verbose` which is always set to true and overlaps logging verbosity
4) move to an util where it can be imported to multiple places without creating import conflicts

----

### DIFF
--- a/examples/launch_plot_results.py
+++ b/examples/launch_plot_results.py
@@ -72,5 +72,5 @@ if __name__ == "__main__":
     tuning_experiment.plot()
 
     # Print the best configuration found from the tuner and retrain it
-    best_config, trial_id = tuner.best_config()
+    trial_id, best_config = tuner.best_config()
     tuner.trial_backend.start_trial(config=best_config)
--- a/examples/launch_plot_results.py
+++ b/examples/launch_plot_results.py
@@ -76,4 +76,3 @@ if __name__ == "__main__":
 
     # plots values found by all trials over time
     tuning_experiment.plot_trials_over_time()
-
--- a/examples/launch_plot_results.py
+++ b/examples/launch_plot_results.py
@@ -63,6 +63,9 @@ if __name__ == "__main__":
 
     tuner.run()
 
+    # shows how to print the best configuration found from the tuner and retrains it
+    trial_id, best_config = tuner.best_config()
+
     tuning_experiment = load_experiment(tuner.name)
 
     # prints the best configuration found from experiment-results
@@ -73,8 +76,4 @@ if __name__ == "__main__":
 
     # plots values found by all trials over time
     tuning_experiment.plot_trials_over_time()
-
-    # prints the best configuration found from the tuner and retrains it
-    trial_id, best_config = tuner.best_config()
-    tuner.trial_backend.start_trial(config=best_config)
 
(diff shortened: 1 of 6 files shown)


syne_tune/util.py
```diff
@@ -319,3 +323,43 @@ def find_first_of_type(a: Iterable[Any], typ) -> Optional[Any]:
...
+        ), f"Attempted to use metric index={metric} with {len(metric_names)} available metrics"
+        metric_name = metric_names[metric]
+    else:
+        raise AttributeError(
+            f"metric must be <int> or <str> but {type(metric)} was provided"
+        )
+
+    if len(metric_names) > 1:
+        logger.warning(
+            "Several metrics exists, this will "
```
wesk, line 355:
```suggestion
            "Several metrics exist, this will "
```

syne_tune/tuner.py
```diff
@@ -683,3 +684,30 @@ def _default_callback():
...
+            self.tuning_status, metric_names=[metric_name], mode=metric_mode
+        )
+        config = self.trial_backend._trial_dict[trial_id].config
+
+        logger.info(
+            f"If you want to retrain the best configuration found, you can run: \n"
+            f"```tuner.trial_backend.start_trial(config={config})``` to start training from scratch\n"
+            f"or\n"
+            f"```tuner.trial_backend.start_trial(config={config}, checkpoint_trial_id={trial_id})``` to start from "
+            f"last checkpoint (your script should have stored a checkpoint)"
```
wesk, line 711: Here or in the FAQ entry, would it make sense to explain when you would use `best_config()` versus a checkpoint?

examples/launch_plot_results.py
```diff
@@ -64,8 +64,13 @@
...
 
+    # Print the best configuration found from experiment-results
     print(f"best result found: {tuning_experiment.best_config()}")
 
+    # Plot the best value found over time
     tuning_experiment.plot()
+
+    # Print the best configuration found from the tuner and retrain it
+    trial_id, best_config = tuner.best_config()
+    tuner.trial_backend.start_trial(config=best_config)
```
wesk, line 76: Maybe plot again, and hopefully show improvement? Or consider splitting out into a separate retraining example? 

Otherwise it feels a bit random - why train again and then do nothing with it after?

syne_tune/tuner.py
geoalgo: It is not really versus, you can only restart from a checkpoint if your script supports checkpointing which may not be the case.

I do not think it would make sense to explain checkpointing there as it has its own set of FAQ items, for instance:
* https://syne-tune.readthedocs.io/en/latest/faq.html#how-can-i-enable-trial-checkpointing

examples/launch_plot_results.py
geoalgo: One use-case could be to run with a larger budget, I do not have one use-case personally but I know some people asks for this so it probably have an example showing how it can be done.

wesk (approved):

PR #784, opened 2023-10-31
fix: Fix sagemaker backend

geoalgo:
Newer version of SageMaker Estimator added a field which is not serializable, removing it.

Should fix integration tests issues https://github.com/awslabs/syne-tune/actions/runs/6704556262.

----


wesk (approved):
LGTM!

PR #798, opened 2023-12-20
feat: Obtain Pareto Set from results

aaronkl:
Parses the results file from Tuner to return Pareto optimal points.

----


jgolebiowski:
I think the code looks good. Could you make sure the linting and unit tests are passing?

NIT: Perhaps add a test to this if you have the time :)

jgolebiowski (approved):
reviewed before, still good!

PR #810, opened 2024-01-31
chore: Uncommented test which needs SageMaker. It is still commented out

### LINKED ISSUE #476: Files seem to be corrupted (hash mismatch)
Hello all,

I updated to the most recent library version and noticed an issue in using the benchmarks. I am getting the following: 
`Files seem to be corrupted (hash mismatch) `, which keeps persisting across multiple code reruns. The issue was fixed in PR #428 initially.

The hash is consistent between the calls, however, it seems not to match the hardcoded one and not to be the same between different operating systems. For example for PD1:

Hardcoded one in syne-tune:

`bd5b599179b1c5163d146a26dd2d559e5cb561f491ef48a22503e821651fd4d1`

On Windows (11) I get:

`f693fed481e344267c3a897eb8e629056e93304ee21e6e90955029c9804cdfda`

On Linux (CentOS 7.9) I get:

`ca162d81cacadb1e177ec319e65d68f812140bbc5864b0dceac28bbcca328a70`

I am able to overcome the problem by hardcoding the hash to my local specific value, but it seems the function that calculates the hash is not working as intended maybe, unless I am doing something wrong.

mseeger:
This is to close an issue. We currently do not execute unit tests which require SageMaker. We run code in `examples/` which uses SageMaker, but not tests.

Closes #476.

----

### DIFF
--- a/tst/remote_launcher/test_remote_launcher_path.py
+++ b/tst/remote_launcher/test_remote_launcher_path.py
@@ -22,33 +22,36 @@ from syne_tune import StoppingCriterion
 from syne_tune import Tuner
 
 
-root = Path(__file__).parent
-sm_estimator = PyTorch(
-    entry_point="folder2/main.py",
-    source_dir=str(root / "folder1"),
-    instance_type="local",
-    instance_count=1,
-    py_version="py3",
-    framework_version="1.7.1",
-    role="dummy",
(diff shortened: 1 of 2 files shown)


jgolebiowski (approved):
LGTM

PR #817, opened 2024-02-07
feat: Allow to restrict GPUs to be used in LocalBackend to a subset o…

### LINKED ISSUE #812: Set custom GPU Ids for LocalBackend
Thank you syne-tune team for your hard work.

**Is your feature request related to a problem? Please describe.**
I would like to set specific GPU IDs to run the tuning jobs, instead of allowing the syne-tune to use all available GPUs, in order, indexing from 0. E.g. I have an 8 GPU machine but I can only use the GPUs at indices 4, 5, 6, 7. Right now, syne-tune will use the first 4 GPUs (0, 1, 2, 3), even with `setenv CUDA_VISIBLE_DEVICES 4,5,6,7` set.

**Describe the solution you'd like**
In the [LocalBackend,](https://github.com/awslabs/syne-tune/blob/6748205788fe791be8daaeb99440574121653052/syne_tune/backend/local_backend.py#L42) I would like to be able to set the GPU IDs to use for the tuning jobs. This could be done by adding a `gpu_ids` list parameter to the `LocalBackend` class, and then using this list to set the `CUDA_VISIBLE_DEVICES` environment variable down the line when allocating the resources for the tuning jobs.

**Describe alternatives you've considered**
Adding this hack at the beginning of my entry script for a job works:

```python
import os
visible_device = os.environ.get("CUDA_VISIBLE_DEVICES")
# Map 0 to 4, 1 to 5, 2 to 6, 3 to 7
if visible_device is not None:
    os.environ["CUDA_VISIBLE_DEVICES"] = str(int(visible_device) + 4)
```

Thank you guys for your hard work.

mseeger:
…f all available ones.

Covered by unit tests.

Closes #812.

----


wesk (approved):
LGTM!

PR #825, opened 2024-02-15
feat: implementation of expected hypervolume improvement

aaronkl:
Implements expected hypervolume improvement, a popular Bayesian optimization strategy for multi-objective problems using the BOTorch implementation. It also fixes a small bug in the FIFO scheduler.

----

### DIFF
--- a/syne_tune/optimizer/schedulers/multiobjective/expected_hyper_volume_improvement.py
+++ b/syne_tune/optimizer/schedulers/multiobjective/expected_hyper_volume_improvement.py
@@ -1,0 +1,366 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


syne_tune/optimizer/schedulers/multiobjective/expected_hyper_volume_improvement.py
```diff
@@ -0,0 +1,336 @@
...
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+from typing import Optional, List, Dict, Any, Union
+import logging
+
+import numpy as np
+
+from syne_tune.try_import import try_import_botorch_message
+
+try:
```
geoalgo, line 20: Not for now and this PR but we really need to get rid of this try import pattern.

geoalgo (approved):

syne_tune/optimizer/schedulers/multiobjective/expected_hyper_volume_improvement.py
aaronkl: +1

PR #839, opened 2024-04-15
feat: add ask/tell interface

aaronkl:
Add a simple ask/tell interface to use SyneTune within a python script.

----

### DIFF
--- a/syne_tune/optimizer/schedulers/ask_tell_scheduler.py
+++ b/syne_tune/optimizer/schedulers/ask_tell_scheduler.py
@@ -1,0 +1,86 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


syne_tune/optimizer/schedulers/ask_tell_scheduler.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class AskTellScheduler:
+    base_scheduler: TrialScheduler
+    trial_counter: int
+    completed_experiments: Dict[int, TrialResult]
+
+    def __init__(self, base_scheduler: TrialScheduler):
+        """
+        Simple interface to use SyneTune schedulers in a custom loop, for example:
```
geoalgo, line 27: Not necessarily for now but perhaps we could add an example.

geoalgo (approved):

syne_tune/optimizer/schedulers/ask_tell_scheduler.py
aaronkl: there is already an example. I will update it in another PR

PR #860, opened 2024-07-19
fix: removes logging in imports

aaronkl:
Removes logging in try/catch blocks for missing dependencies to remove clutter in logs.
----


geoalgo (commented):
Oh sorry I had review this a while ago but did not send my comment...

My main comment is to use logging.debug for now as otherwise those try/catch would swallow exception silently which would be a bit dangerous.
In another step/PR I think we should get rid of most of those try/import to simplify the code (which would also be easier when we simplify the requirements.txt)

examples/launch_asha_yahpo.py
```diff
@@ -22,12 +22,11 @@
 from syne_tune.optimizer.baselines import ASHA
 from syne_tune import Tuner, StoppingCriterion
 from syne_tune.config_space import Domain
-from syne_tune.try_import import try_import_visual_message
 
 try:
     import matplotlib.pyplot as plt
 except ImportError:
-    print(try_import_visual_message())
```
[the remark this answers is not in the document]
aaronkl: What do you think of moving the import to the end and moving all the plotting code into the try/catch block. This allows us to essentially run the code without matplotlib

aaronkl:
Agreed, I'll change to logging.debug but I'd like to still remove all the `try_import_X_messages()` to reduce complexity. This is just a hotfix until we consolidated all the dependencies

examples/launch_asha_yahpo.py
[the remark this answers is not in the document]
geoalgo: 👍 This seems better indeed, alternatively, we could make matplotlib required for this example.

examples/example_automatic_termination_criterion.py
```diff
@@ -23,7 +23,10 @@
 try:
     import matplotlib.pyplot as plt
 except ImportError:
-    pass
+    print(
+        "This example requires matplotlib. You can install via pip install matplotlib"
+    )
+    raise
```
geoalgo, line 29: What is the benefit of raising here if we have a try-catch? Isnt it better then to just remove it and import directly?

syne_tune/blackbox_repository/conversion_scripts/utils.py
```diff
@@ -12,25 +12,7 @@
...
-@lru_cache(maxsize=1)
-def s3_blackbox_folder(s3_root: Optional[str] = None):
-    if s3_root is None:
-        if "AWS_DEFAULT_REGION" not in os.environ:
-            # avoids error "Must setup local AWS configuration with a region supported by SageMaker."
-            # in case no region is explicitely configured
-            os.environ["AWS_DEFAULT_REGION"] = "us-west-2"
-        s3_root = sagemaker.Session().default_bucket()
-    return f"{s3_root}/blackbox-repository"
-
```
geoalgo: Thanks for the good catch!

examples/launch_asha_yahpo.py
```diff
@@ -23,40 +23,40 @@
...
-            k: v.sample() if isinstance(v, Domain) else v
-            for k, v in bb.configuration_space.items()
-        }
-        evals = bb(config)
-        time_index = next(
-            i for i, name in enumerate(bb.objectives_names) if name == time_col
-        )
-        accuracy_index = next(
-            i for i, name in enumerate(bb.objectives_names) if name == metric_col
+    try:
```
geoalgo, line 30: Lets do it this way for now, but I think we should just import normally matplotlib in this case since the example would not run anyway.

geoalgo (dismissed):

examples/example_automatic_termination_criterion.py
aaronkl: makes sense

examples/launch_asha_yahpo.py
aaronkl: I think you are right. I'll change it, because right now the it just clutters the code

geoalgo (approved):

PR #861, opened 2024-07-20
[fix] use free workers, remove dependabot as we dont use it, remove end2end…

geoalgo:
… tests which we cannot run anymore (as-is)

In the mid-term, we should select a list of examples to run. Running all of them as of now is not an option as we would exhaust our free worker time very fast.

Also update python version to more recent ones and use 3 instead of 4 to save compute.

----


.github/dependabot.yml
wesk, line 1: Let's keep dependabot?

.github/workflows/integ-tests.yml
wesk, line 1: How long do these integ tests take on the regular runners now? I'd recommend we try keeping them, or at least keep them as tests that we can invoke manually.

.github/dependabot.yml
geoalgo: I would be keen to remove this action as it creates a lot of PRs that we do not have time to handle (certainly not now with limited ressources but also before as we almost always had a bunch open).

.github/workflows/integ-tests.yml
geoalgo: Makes sense but I would prefer to do it in a second step. 
Right now, the pipeline is completely blocked, this PR deblocks it and allows to run unit-tests in PRs which is quite urgent. 

We could add test for some examples back later but we should be watchful with our runtime as we now have limits with github actions (if we exceeed our limit, we cannot run test anymore and the limit is relatively low => 30h per month).

wesk: Thanks for unblocking the tests with this!

One thing to note: I believe the 2000 minutes per month limit is for private repositories, not public repositories, see: 
* https://stackoverflow.com/questions/62724429/what-are-github-actions-minutes-month
* https://github.com/orgs/community/discussions/70492

Rather than deleting the integ tests, can you instead switch them so that they are invoked manually instead? We can then run them manually before releases.

At the top of the integ tests file, replace
```
on:
  push:
    branches:
      - main
  pull_request:
    branches:
      - main
```

with just the following: 

```
on:
  workflow_dispatch:
```

geoalgo: Makes sense, this is a good idea thanks for the suggestion.

wesk (approved):
LGTM!

PR #863, opened 2024-07-22
Blackbox repository to use HuggingFace as storage backend instead of S3

geoalgo:
Because the options of HF hub are different than what was supported in BB repository, the options now looks a bit different but I did my best to maintain option parity when possible.
The hash is not needed anymore since HF hub supports it.

I have noted an issue with LCBench data as an error is thrown now with a pandas update. I have sent an email to the authors and hotfixed the blackbox import code so that it can run with newer pandas code.

### DIFF
--- a/syne_tune/blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
+++ b/syne_tune/blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
@@ -237,7 +237,11 @@ class NASBench201Recipe(BlackboxRecipe):
 
                     session = requests.Session()
 
-                    params = {"id": id, "confirm": "t"}
+                    print(
+                        "TODO fix me, right now this download does not work as google added a warning, "
+                        "file must be downloaded manually for now"
+                    )
+                    params = {"id": id, "confirm": True}
                     response = session.get(URL, params=params, stream=True)
 
                     save_response_content(response, destination)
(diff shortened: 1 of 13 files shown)


syne_tune/blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
```diff
@@ -240,7 +237,7 @@ def download_file_from_google_drive(id, destination):
 
                     session = requests.Session()
 
-                    params = {"id": id, "confirm": True}
+                    params = {"id": id, "confirm": "t"}
```
aaronkl, line 240: Is this is typo or did the interface change?

syne_tune/blackbox_repository/repository.py
```diff
@@ -92,87 +76,52 @@ def load_blackbox(
...
+        )
+
         if yahpo_kwargs is None:
             yahpo_kwargs = dict()
         return instantiate_yahpo(name, **yahpo_kwargs)
     elif name.startswith("pd1"):
-        return deserialize_pd1(tgt_folder)
-    elif (tgt_folder / "hyperparameters.parquet").exists():
-        return deserialize_tabular(tgt_folder)
+        print(blackbox_path)
```
aaronkl, line 119: can we remove prints?

aaronkl (approved):
LGTM

syne_tune/blackbox_repository/conversion_scripts/scripts/nasbench201_import.py
geoalgo: Yes it changed, but this does not fix it unfortunately, I left a comment instead.

syne_tune/blackbox_repository/repository.py
geoalgo: Good catch.

aaronkl (approved):

PR #867, opened 2024-08-14
fix: update docs

aaronkl:
- Removes any references to the Sagemaker backend from the docs since it will be moved to another package
- Removes some of the Tutorial sections  because they are outdated.

----


geoalgo (approved):
Makes sense, I think we should create another package to still expose this content easily while specifically mentioning the version required.

geoalgo (approved):

PR #878, opened 2024-10-24
refactor: scheduler/searcher interface to handle both multi and single-objective optimization

aaronkl:
- Pass a list of metrics to `BaseSearcher` and `SingleFidelityScheduler` to allow for multi-objective optimization.
- Implement a new `SingleObjectiveSearcher`  and `SingleObjectiveScheduler` for sinlge-objective optimization.
- Adapt `RandomSearcher` to new interface
- Provide unit test for the searcher / scheduler interface
- Deprecate old `RandomSearcher` -> `LegacyRandomSearcher`

----


syne_tune/optimizer/schedulers/single_fidelity_scheduler.py
```diff
@@ -52,15 +52,18 @@ class SingleFidelityScheduler(TrialScheduler):
     def __init__(
         self,
         config_space: Dict[str, Any],
-        metric: str,
+        metric: List[str] | str,
```
geoalgo, line 55: Could we perhaps inherit rather? (Handling either list or single metrics was causing a lot of errors in our previous version)

aaronkl: I also thought about it. What do you think about allowing only a list of metrics to `SingleFidelityScheduler`, i.e making it multi-objective only, and inherit a `SingleObjectiveScheduler` that only expect a single metric? The idea here is that single-objective optimization is a special case of multi-objective optimization.

geoalgo: That can work also, if we can avoid to use `metric` to refer to a list of metrics it would be much better.

aaronkl: I am wondering if we should do the same thing for `BaseSearcher`?

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -38,6 +38,8 @@ def __init__(
         self.config_space = config_space
         if points_to_evaluate is None:
             self.points_to_evaluate = []
+        else:
+            self.points_to_evaluate = points_to_evaluate
```
geoalgo, line 42: Before we add new functionalities, can we test the ones we added in the previous PR? We should make sure that all our searchers and scheduler verify the same API (can suggest ~5 configurations, return suggestion which matches points to evaluate, returns suggestion that have the expected signature).

I believe that it would be better to do this before adding functionalities (such as multiobjective) as we want to make sure the functionality we work as expected.

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -74,7 +76,7 @@ def on_trial_result(
         self,
         trial_id: int,
         config: Dict[str, Any],
-        metric: float,
+        metric: List[float],
```
geoalgo, line 79: Can we have another class? Otherwise we have lots of tension in our base class, for instance right now, we have `metric`, `metrics`and the type are sometimes `list[float]` and sometimes `Union[list[float], float]`.

I think we should have another class also because the multiobjective case is something that we only want to use *sometimes*. And we should not put the union of things that we need *sometimes* in the base class.

syne_tune/optimizer/schedulers/searchers/random_searcher.py
```diff
@@ -0,0 +1,38 @@
...
+
+class RandomSearcher(BaseSearcher):
+    """
+    Searcher which randomly samples configurations to try next.
+    """
+
+    def __init__(
+        self,
+        config_space: Dict[str, Any],
+        points_to_evaluate: Optional[List[dict]] = None,
```
geoalgo, line 17: With recent enough python, we can do directly:
```suggestion
        points_to_evaluate: list[dict] | None = None,
```
which avoids the import of optional and list, I would suggest to start doing this as it makes the code much leaner.

syne_tune/optimizer/schedulers/searchers/random_searcher.py
```diff
@@ -0,0 +1,38 @@
...
+        random_seed: int = None,
```
geoalgo, line 18: The correct type would be:
```suggestion
        random_seed: int | None = None,
```

syne_tune/optimizer/schedulers/searchers/random_searcher.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def suggest(self, **kwargs) -> Optional[dict]:
+        """Sample a new configuration at random
+
+        If ``allow_duplicates == False``, this is done without replacement, so
+        previously returned configs are not suggested again.
+
+        :return: New configuration, or None
+        """
+        new_config = self._next_initial_config()
```
geoalgo, line 32: Could we name it `_next_points_to_evaluate`? This would avoid naming the same things in two different ways.

syne_tune/optimizer/schedulers/searchers/random_searcher.py
```diff
@@ -0,0 +1,38 @@
...
+        self,
+        config_space: Dict[str, Any],
+        points_to_evaluate: Optional[List[dict]] = None,
+        random_seed: int = None,
+    ):
+        super().__init__(
+            config_space, points_to_evaluate=points_to_evaluate, random_seed=random_seed
+        )
+
+    def suggest(self, **kwargs) -> Optional[dict]:
```
geoalgo, line 24:
```suggestion
    def suggest(self, **kwargs) -> dict | None:
```

geoalgo, line 24: Can we avoid the unused kwargs here?

syne_tune/optimizer/schedulers/searchers/searcher.py
```diff
@@ -74,7 +77,7 @@ def on_trial_result(
...
```
geoalgo, line 80: The naming should be metrics then.
```suggestion
        metrics: list[float],
```

syne_tune/optimizer/schedulers/searchers/single_objective_searcher.py
```diff
@@ -0,0 +1,55 @@
...
+        determined by the midpoint heuristic. If ``[]`` (empty list), no initial
+        configurations are specified.
+    """
+
+    def on_trial_result(
+        self,
+        trial_id: int,
+        config: Dict[str, Any],
+        metric: float,
+        update: bool,
```
geoalgo, line 38: Could we remove `update`? (it seems that it should be specific class logic)

syne_tune/optimizer/schedulers/searchers/random_searcher.py
geoalgo: Rethinking about this one, we probably want to use the old typing in this PR as you did to avoid mixing different thing and having a very large diff.

PR #890, opened 2024-12-10
fix: use trial_complete

aaronkl:
Make single fidelity methods use `on_trial_complete` 

----


geoalgo (dismissed):

geoalgo:
Btw, just noticed that the test are failling, it seems to be due to quantile searcher and issues with sklearn updates, maybe we can skip the test and add an issue?

aaronkl:
Yes seems to be a XGBoost problem. Let's track it here #892

geoalgo (approved):

PR #896, opened 2024-12-21
fix: use specified tuner name

aaronkl:
Don't overwrite `tuner_name` given as input.

----


syne_tune/tuner.py
```diff
@@ -153,11 +153,12 @@ def __init__(
...
 
-        if tuner_name is not None:
+        if tuner_name is None:
             tuner_name = Path(self.trial_backend.entrypoint_path()).stem.replace(
                 "_", "-"
             )
-        if suffix_tuner_name or tuner_name is None:
+
+        if suffix_tuner_name:
             self.name = name_from_base(tuner_name, default="st-tuner")
```
geoalgo, line 162:
```suggestion
        if suffix_tuner_name:
            # suffix a timestamp to get a unique string
            self.name = name_from_base(tuner_name)
```

geoalgo: (`tuner_name` cant be None so passing default is not needed anymore)

geoalgo (approved):
Thanks for the fix, I see that this bug was introduced in #865 when moving out AWS code.

geoalgo (dismissed):

syne_tune/tuner.py
```diff
@@ -153,12 +153,14 @@ def __init__(
...
+        if tuner_name is None:
             tuner_name = Path(self.trial_backend.entrypoint_path()).stem.replace(
                 "_", "-"
             )
-        if suffix_tuner_name or tuner_name is None:
-            self.name = name_from_base(tuner_name, default="st-tuner")
+
+        if suffix_tuner_name:
+            # suffix a timestamp to get a unique string
+            self.name = name_from_base(tuner_name)
```
geoalgo, line 163: I realize now that we also should make `default` optional in `name_from_base`.

geoalgo (approved):

Issue #879, opened 2024-11-07
Refactor scheduler / searcher interface


aaronkl:
The current interfaces for schedulers and searchers are flexible and highly configurable, but they can be difficult to understand and maintain. The FIFOScheduler in particular is complex and not easily adaptable. To encourage wider adoption of the library—especially among HPO developers—and to reduce maintenance overhead, I suggest simplifying the current scheduler and searcher interfaces by:

- Removing boilerplate code
- Minimizing the base class and removing specific class logic
- Creating a cleaner separation between scheduler and searcher
- Avoiding mutation of kwargs objects
- Explicitly handling both single-objective and multi-objective searchers

# Tasks

- [x] add new interface for single-fidelity and multi-fidelity scheduling #877 
- [x] add new interface for single and multi-objective searcher #878 
- [x] deprecate current FIFO and Hyperband scheduler
- [x] migrate existing single-fidelity scheduler / searchers to new interface.
    - [x] RandomSearcher  
    - [x] BORE
    - [x] REA
    - [x] KDE
    - [x] CQR
    - [x] GP-BO
    - [x] SMAC

- [x] migrate existing multi-fidelity scheduler / searchers to new interface.
    - [x] ASHA
    - [x] Model-based ASHA
    - [x] Median Stopping Rule
    - [x] PBT
  
- [x] migrate multi-objective methods
   - [X] EHVI
   - [x] Linear scalarization
   - [x] MO-ASHA
   
- [x] migrate transfer-learning methods
   - [x] bounding box 
   - [x] zero-shot
   - [x] quantile
- [x] update doc-strings
- [x] update baselines.py

PR #898, opened 2025-01-11
refactor: pbt

aaronkl:
refactor PBT as described in #879 

----

### DIFF
--- a/syne_tune/optimizer/schedulers/pbt.py
+++ b/syne_tune/optimizer/schedulers/pbt.py
@@ -14,15 +14,8 @@ from syne_tune.optimizer.scheduler import (
     TrialSuggestion,
     TrialScheduler,
 )
-from syne_tune.config_space import cast_config_values
+from syne_tune.config_space import cast_config_values, postprocess_config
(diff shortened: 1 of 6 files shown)


syne_tune/optimizer/schedulers/pbt.py
```diff
@@ -34,32 +31,7 @@ class PBTTrialState:
...
-    "resource_attr": DA_String(),
-    "population_size": DA_Integer(1, None),
-    "perturbation_interval": DA_Float(0.01, None),
-    "quantile_fraction": DA_Float(0.0, 0.5),
-    "resample_probability": DA_Float(0.0, 1.0),
-}
-
-
-class PopulationBasedTraining(FIFOScheduler):
+class PopulationBasedTraining(TrialScheduler):
```
geoalgo, line 34: Could we add test for this one? (potentially dropping the tests for the other?)
I checked and I could not see any test for this one.

geoalgo (approved):

syne_tune/optimizer/schedulers/pbt.py
geoalgo: Its all good actually, I missed the test in my review.

PR #900, opened 2025-01-28
fix: pass single metric to SingleObjectiveBaseSearcher

aaronkl:
Makes sure that we only pass a single float as metric to searchers derived from SingleObjectiveBaseSearcher

----


syne_tune/optimizer/schedulers/single_fidelity_scheduler.py
```diff
@@ -103,7 +106,11 @@ def on_trial_result(self, trial: Trial, result: Dict[str, Any]) -> str:
         metric = [
             result[metric_name] * self.metric_multiplier for metric_name in self.metrics
         ]
-        self.searcher.on_trial_result(trial.trial_id, config, metric)
+
+        if isinstance(self.searcher, SingleObjectiveBaseSearcher):
```
geoalgo, line 110: Can we assert in the constructor that in such case we have a single metric?

syne_tune/optimizer/schedulers/single_fidelity_scheduler.py
```diff
@@ -103,7 +106,11 @@ def on_trial_result(self, trial: Trial, result: Dict[str, Any]) -> str:
         metric = [
             result[metric_name] * self.metric_multiplier for metric_name in self.metrics
         ]
```
geoalgo, line 108: Can we rename here and bellow with `metrics`? (given that we have a list, metric is a misnommer)
```suggestion
        metrics = [
            result[metric_name] * self.metric_multiplier for metric_name in self.metrics
        ]
```

syne_tune/optimizer/schedulers/single_fidelity_scheduler.py
```diff
@@ -120,7 +127,11 @@ def on_trial_complete(self, trial: Trial, result: Dict[str, Any]):
         metric = [
```
geoalgo, line 127: same here, could we use metrics?

geoalgo (commented):
LGTM, I just have minor comments.

geoalgo (approved):

syne_tune/optimizer/schedulers/single_fidelity_scheduler.py
```diff
@@ -77,9 +77,11 @@ def __init__(
             self.searcher = searcher
 
         if isinstance(self.searcher, SingleObjectiveBaseSearcher):
-            assert len(metrics) == 1, logging.error(
-                f"Searcher {self.searcher} "
-                f"only support single metric optimization but number of metrics is {len(metrics)}"
+            assert len(self.metrics) == 1, (
+                f"Searcher {self.searcher} only support single metric optimization "
+                f"{issubclass(type(self.searcher), BaseSearcher)} "
+                f"{issubclass(type(self.searcher), SingleObjectiveBaseSearcher)} "
```
geoalgo, line 83: Is it a typo? It sounds like we would not want to print those?

geoalgo (approved):

PR #903, opened 2025-02-02
refactor: ehvi

aaronkl:
refactors ehvi and deprecates old version. #879 

----

### DIFF
--- a/syne_tune/optimizer/schedulers/multiobjective/expected_hyper_volume_improvement.py
+++ b/syne_tune/optimizer/schedulers/multiobjective/expected_hyper_volume_improvement.py
@@ -1,4 +1,4 @@
-from typing import Optional, List, Dict, Any, Union
+from typing import Optional, List, Dict, Any
 import logging
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


syne_tune/optimizer/schedulers/multiobjective/expected_hyper_volume_improvement.py
```diff
@@ -99,36 +90,26 @@ def __init__(
...
-    def clone_from_state(self, state: Dict[str, Any]):
-        raise NotImplementedError
-
     def num_suggestions(self):
         return len(self.trial_configs)
 
-    def _get_config(self, trial_id: str, **kwargs) -> Optional[dict]:
-        trial_id = int(trial_id)
-        config_suggested = self._next_initial_config()
+    def suggest(self, **kwargs) -> Optional[dict]:
```
geoalgo, line 111: Can we remove the kwargs here if we dont need them?

tst/schedulers/test_schedulers_api.py
```diff
@@ -90,6 +93,17 @@
...
     ),
+    # Multi-objective methods
+    SingleFidelityScheduler(
+        config_space,
+        searcher=ExpectedHyperVolumeImprovement(
+            config_space=config_space,
+            random_seed=random_seed,
+        ),
+        metrics=[metric1, metric2],
+        do_minimize=False,
```
geoalgo, line 104: Not for this PR, but we should probably think on an easy way to specify mode per metrics in the case of multi-objective models.

geoalgo (commented):
LGTM, just one small comment.

geoalgo (approved):

PR #906, opened 2025-02-04
refactor: make asha compatibles with models

aaronkl:
- passes resource level to searcher
- adapts multi-fidelity BORE to make it compatible with ASHA
- allows to pass a model to ASHA

----


syne_tune/optimizer/schedulers/searchers/kde/multi_fidelity_kde_searcher.py
```diff
@@ -88,7 +88,7 @@ def on_trial_result(
         trial_id: int,
         config: Dict[str, Any],
         metric: float,
-        resource_level: float,
+        resource_level: int = None,
```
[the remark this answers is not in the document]
aaronkl: Actually I think it should be int, for methods such as BOHB that maintain different models per resource level it might be prone to rounding errors otherwise. Also runtime should probably be cast to integer for that reason.

syne_tune/optimizer/schedulers/searchers/single_objective_searcher.py
```diff
@@ -35,6 +35,7 @@ def on_trial_result(
         trial_id: int,
         config: Dict[str, Any],
         metric: float,
+        resource_level: int = None,
```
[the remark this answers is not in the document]
aaronkl: Yes I agree, but I think it require some restructuring

aaronkl:
I simplified the interface even further. Now, there is only a generic IndependentMultiFidelitySearcher class that can receive any SingleObjectiveBaseSearcher class as inputs and fits independent models on each resource level as described by Falkner et al.

syne_tune/optimizer/schedulers/searchers/multi_fidelity_searcher.py
```diff
@@ -0,0 +1,151 @@
...
+            self.searcher_kwargs = dict()
+        else:
+            self.searcher_kwargs = searcher_kwargs
+
+        if isinstance(searcher_cls, str):
+            assert searcher_cls in searcher_cls_dict
+            self.searcher_cls = searcher_cls_dict.get(searcher_cls)
+        else:
+            self.searcher_cls = searcher_cls
+        self.models = OrderedDict()
```
geoalgo, line 60: Is model the best name here? Perhaps `searchers` or `searchers_resource` are better names?
```suggestion
        self.searchers = OrderedDict()
```

syne_tune/optimizer/schedulers/searchers/multi_fidelity_searcher.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def initialize_model(self):
+        return self.searcher_cls(
+            config_space=self.config_space,
+            random_seed=self.random_seed,
+            **self.searcher_kwargs,
+        )
```
geoalgo, line 67: Wouldnt it be better to use defaultdict?

```
self.searchers = defaultdict(lambda : self.searcher_cls(
      config_space=self.config_space,
      random_seed=self.random_seed,
      **self.searcher_kwargs,
  ))
```

It seems cleaner but I dont mind if you prefer the current option.

syne_tune/optimizer/schedulers/searchers/multi_fidelity_searcher.py
```diff
@@ -0,0 +1,151 @@
...
+            stop. This happens if searchers never suggest the same config more
+            than once, and all configs in the (finite) search space are
+            exhausted.
+        """
+        suggestion = self._next_points_to_evaluate()
+        if suggestion is not None:
+            return suggestion
+
+        if len(self.models) == 0:
+            return {
```
geoalgo, line 88:
```suggestion
        # in case we have not seen observations, we have no searcher and sample at random
        if len(self.models) == 0:
            return {
```

syne_tune/optimizer/schedulers/searchers/multi_fidelity_searcher.py
```diff
@@ -0,0 +1,151 @@
...
+    def on_trial_result(
+        self,
+        trial_id: int,
+        config: Dict[str, Any],
+        metric: float,
+        resource_level: int,
+    ):
+        """Inform searcher about result
+
+        The scheduler passes every result. If ``update == True``, the searcher
```
geoalgo, line 105: Can you update the docstring? (we dont have `update`)

syne_tune/optimizer/schedulers/searchers/multi_fidelity_searcher.py
```diff
@@ -0,0 +1,151 @@
...
+        should update its surrogate model (if any), otherwise ``result`` is an
+        intermediate result not modelled.
+
+        The default implementation calls :meth:`_update` if ``update == True``.
```
geoalgo, line 137: Same comment here.

geoalgo (commented):
LGTM, have mostly minor comments.

syne_tune/optimizer/schedulers/searchers/multi_fidelity_searcher_factory.py
```diff
@@ -7,11 +7,13 @@
...
+    MultiFidelityRandomSearcher,
+)
 from syne_tune.optimizer.schedulers.searchers.searcher import BaseSearcher
 
 multi_fidelity_searcher_cls_dict = {
-    "random_search": RandomSearcher,
+    "random_search": MultiFidelityRandomSearcher,
     "bore": MultiFidelityBore,
     "kde": MultiFidelityKernelDensityEstimator,
     "cqr": SurrogateSearcher,
```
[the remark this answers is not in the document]
aaronkl: actually this function is deprecated since we can just pass any `SingleObjectiveSearcher` here. But I agree we should have a factory method that allows to generate different model-based multi-fidelity strategies

syne_tune/optimizer/schedulers/searchers/multi_fidelity_searcher.py
aaronkl: The problem here would be that we lose the order no? For sampling, we always want to select the last model that we trained

syne_tune/optimizer/schedulers/searchers/multi_fidelity_searcher_factory.py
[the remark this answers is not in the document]
geoalgo: Ok thanks, is there a reason for not removing it then?

syne_tune/optimizer/schedulers/searchers/multi_fidelity_searcher.py
geoalgo: Good point to check indeed, `defaultdict` is ordered for python 3.7+ so it is safe.

geoalgo (approved):
Two remaining comments: can we remove the factory if it is outdated? can we use defaultdict? We could do it in another PR on in this one as you prefer.

geoalgo (approved):

PR #908, opened 2025-02-06
refactor: bounding box transfer method

aaronkl:
- updates `BoundingBox` and `TransferLearningMixin` to new interface #879 
- deprecates old code 
- adds a simple unit test for `BoundingBox`

----


syne_tune/optimizer/schedulers/transfer_learning/bounding_box.py
```diff
@@ -66,47 +70,43 @@ def scheduler_fun(new_config_space: Dict[str, Any], mode: str, metric: str):
 
     def __init__(
         self,
-        scheduler_fun: Callable[[dict, str, str], LegacyTrialScheduler],
+        scheduler_fun: Callable[[dict, str, bool, int], SingleObjectiveScheduler],
```
geoalgo, line 73: Can you update the docstring?

syne_tune/optimizer/schedulers/transfer_learning/transfer_learning_mixin.py
```diff
@@ -0,0 +1,71 @@
...
+        :returns: Dict which maps from task name to list of hyperparameters in order.
+        """
+        assert num_hyperparameters_per_task > 0 and isinstance(
+            num_hyperparameters_per_task, int
+        ), f"{num_hyperparameters_per_task} is no positive integer."
+        assert mode in ["min", "max"], f"Unknown mode {mode}, must be 'min' or 'max'."
+        best_hps = dict()
+        for task, evaluation in transfer_learning_evaluations.items():
+            best_hps[task] = evaluation.top_k_hyperparameter_configurations(
+                num_hyperparameters_per_task, mode, self.metric
```
geoalgo, line 69: Could we change to use `do_minimize` rather than `mode`? It is slightly more annoying but would be a cleaner approach. Otherwise some code will keep "min", "max" as the mode internally, if we can avoid it, that would be nice (I guess for some scheduler it may be too complicated but for this one it sounds doable?).

geoalgo (commented):
LGTM, two small comments.

syne_tune/optimizer/schedulers/transfer_learning/transfer_learning_mixin.py
geoalgo: We can also do it in another step if you prefer (I could take a look), let me know.

geoalgo (approved):

PR #931, opened 2025-05-07
refactor: surrogate searcher

aaronkl:
- Renames SurrogateSearcher to ConformalQuantileRegression for clarity; it now only supports single-fidelity optimization.

- Adds a new LastValueMultiFidelitySearcher, which takes any SingleObjectiveSearcher as input and fits a model using only the last observed value from each trial.

----


geoalgo (approved):
neat!

PR #947, opened 2025-05-30
fix: clean up examples

### LINKED ISSUE #905: some examples are broken
Some examples are broken. I propose to remove all examples that are broken, we should also try to limit the number of examples to ease maintenance (right now, we have 33 examples)

```bash
git ls-files examples | grep launch | wc -l
      33
```

aaronkl:
Closes #905 

----


Kavlahkaff (dismissed):
Left only one minor comment. Looks good to me otherwise!

Kavlahkaff (approved):
Thank you, looks good!

PR #954, opened 2025-06-11
fix: typo in LastValueMultiFidelitySearcher

### LINKED ISSUE #952: Running Blackbox example: "TypeError: 'NoneType' object is not callable" in last_value_multi_fidelity_searcher.py
**Describe the bug**  
Python raises Exception TypeError: 'NoneType' object is not callable in last_value_multi_fidelity_searcher.py when running the train blackbox example from README.

Then see log output.

[log_output_TypeError_NoneType.txt](https://github.com/user-attachments/files/20672590/log_output_TypeError_NoneType.txt)

**To reproduce**  
Same steps  as  in Issue #951 .
Then I fixed the upper range in the randint in #951 to 2**31-1.
Then run train_blackbox.py.

**Expected behavior**  
The Python code in train_blackbox.py is properly executed. Python does not fail intermediately with any exception.

**Paste the output of the `pip freeze` command below:**
Same as  in Issue #951 

**Desktop (please complete the following information):**  
Same as  in Issue #951 

**Screenshots**  

**Additional context**

aaronkl:
Fix a typo that was introduced in #945 to correctly initialize the searcher.

Closes #952 

----

### DIFF
--- a/tst/schedulers/test_schedulers_api.py
+++ b/tst/schedulers/test_schedulers_api.py
@@ -263,7 +263,7 @@ list_schedulers_to_test = [
 @pytest.mark.timeout(20)
 @pytest.mark.parametrize("scheduler", list_schedulers_to_test)
 def test_schedulers_api(scheduler):
-    trial_ids = range(4)
+    trial_ids = range(6)
 
     # checks suggestions are properly formatted
     trials = []


geoalgo (commented):
Can we add a simple smoke test that checks the instantiation of this class to avoid regression?

geoalgo:
Running example as integration test makes sense, we would need to reduce the number though as it has increased a lot.

However, what I am suggesting is a unit-test that test the instantiations of this class that would have catch that some instantiations fail.

aaronkl:
Ah I see.  We have a unit-test that tests the instantiations [here](https://github.com/syne-tune/syne-tune/blob/main/tst/schedulers/test_schedulers_api.py). Unfortunately, the error only occurred after running the scheduler for a few iterations, that's why it wasn't caught by our unit tests.

PR #971, opened 2025-07-04
refactor: remove legacy code

aaronkl:
removes deprecated code https://github.com/syne-tune/syne-tune/issues/965

replaces #968 

----


tst/test_schedulers.py
```diff
@@ -1,190 +0,0 @@
...
-from syne_tune.optimizer.schedulers.legacy_fifo import LegacyFIFOScheduler
-from syne_tune.optimizer.schedulers.synchronous.hyperband_impl import (
-    SynchronousGeometricHyperbandScheduler,
-    GeometricDifferentialEvolutionHyperbandScheduler,
-)
-from syne_tune.config_space import choice
-from tst.util_test import run_experiment_with_height
-
-
-def make_async_scheduler(scheduler, searcher):
```
geoalgo, line 14: We have the equivalent new schedulers right?

examples/launch_tensorboard_example.py
```diff
@@ -21,16 +21,11 @@
 from pathlib import Path
```
geoalgo, line 21: For each of the deleted example, can you grep the codebase to make sure they dont appear in the doc or faq?

geoalgo (commented):
LGTM, could you also update the list of schedulers supported?

https://syne-tune.readthedocs.io/en/latest/getting_started.html#supported-hpo-methods

tst/test_schedulers.py
```diff
@@ -0,0 +1,63 @@
...
+asha_variants = ["cqr", "random_search", "kde", "regularized_evolution", "bore"]
+
+
+@pytest.mark.timeout(10)
+@pytest.mark.parametrize("searcher", asha_variants)
+def test_single_objective_schedulers_local(searcher):
+    run_experiment_with_height(
+        make_scheduler=make_asha_scheduler(searcher),
+        simulated=False,
+    )
```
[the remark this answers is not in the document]
aaronkl: We could skip it, but right now the unit-test take only <30 seconds to run which is much less than installing all the dependencies.

Issue #966, opened 2025-06-25
move nursery


aaronkl:
Clean up syne tune repo by moving nursery into a separate repo

PR #975, opened 2025-07-11
refactor: remove nursery

aaronkl:
#966 

----


geoalgo (commented):
LGTM, maybe we can put it in another repo? (it is in git but may be hard to access otherwise?)

geoalgo (approved):

PR #979, opened 2025-07-29
fix points_to_evaluate in ASHA baseline

geoalgo:
`points_to_evaluate` is ignored currrently we should fix it.
One thing though I was not sure about, should we make sure the scheduler handle `points_to_evaluate` or the searcher?

Right now, we have some tension as we are passing this to both the searcher and the scheduler so it is not clear who is responsible for handling it (for instance what happens if you pass `points_to_evaluate` to ASHA with a custom searcher with a different `points_to_evaluate`?)


geoalgo:
To be clear, this is WIP.

aaronkl:
Thanks for the catch! Right now `points_to_evaluate` is only handled by the searcher. I guess we can simply fix it by passing it as a `searcher_kwargs`?

geoalgo:
Hi @aaronkl thanks for taking a look I did in a cleaner fashion, agreed that using kwargs is better, also fixed the test which was not doing anything.

aaronkl (approved):

PR #990, opened 2025-08-01
refactor: remove searcher legacy code

aaronkl:
----

### DIFF
--- a/tst/trial_backend/test_local_trial_backend.py
+++ b/tst/trial_backend/test_local_trial_backend.py
@@ -2,11 +2,7 @@ import logging
 from pathlib import Path
 from typing import Any
 
-import pytest
-
-from syne_tune.backend.trial_status import Status
-from syne_tune.util import script_checkpoint_example_path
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -108,7 +108,7 @@ extra = [
 "Bug Tracker" = "https://github.com/awslabs/syne-tune/issues"
 
 [tool.setuptools.packages.find]
-include = ["syne_tune", "syne_tune.*"]
+include = ["syne_tune", "syne_tune.*", "examples", "examples.*"]
```
geoalgo, line 111: why include examples? we probably dont want to ship when pip installing right?

aaronkl: Right, the unit tests fails because of some files from examples cannot be imported

aaronkl: Ah sorry, I just added it for testing. It seems to be some import error. I removed the unit test that imported examples, because most of them where skipped anyway.

geoalgo (approved):

PR #995, opened 2025-08-18
feat: add hebo searcher

Kavlahkaff:
This PR will wrap the optuna implementation of HEBO in Syne Tune.

----
Closes Issue #996


syne_tune/optimizer/baselines.py
```diff
@@ -465,10 +465,46 @@ def __init__(
         )
 
 
+class HEBO(SingleObjectiveScheduler):
+    """
+    HEBO
```
aaronkl, line 470: Can you add a reference to the original paper? Also mention that we wrap the optuna code

syne_tune/optimizer/schedulers/searchers/hebo/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+        HEBOSearcher,
+    )
+
+    __all__.extend(
+        [
+            "HEBOSearcher",
+        ]
+    )
+except ImportError as e:
+    logging.debug(e)
```
aaronkl, line 16: Can you raise a warning instead of debug? Also mention in the warning that uses need to install optuna to use HEBO

syne_tune/optimizer/schedulers/searchers/hebo/hebo_searcher.py
```diff
@@ -0,0 +1,177 @@
...
+    Convert a Syne Tune domain (sp.*) or literal constant into an Optuna distribution.
+    Returns (optuna_dist or None for constants, metadata dict)
+    """
+    # literal constant
+    if isinstance(dom, (int, float, bool, str)):
+        return None, {"constant": dom}
+
+    # Categorical
+    if isinstance(dom, sp.Categorical):
+        choices = dom.categories
```
aaronkl, line 34: Can choices be None? They should be always defined for categoricals?

aaronkl (commented):
Thanks for implementing this wrapper! After thinking about it a bit more, I am wondering if we already want to have this in Syne Tune yet or if we should instead move it to our experiment repo (open_optformer)? We could move it later to Syne Tune once we finish the benchmarking. @geoalgo wdyt?

PR #997, opened 2025-08-21
reformat: Unify naming in schedulers

### LINKED ISSUE #918: unify naming of resource/time attribute
**Is your feature request related to a problem? Please describe.**
Some schedulers (e.g ASHA) use `time_attr` to define the fidelity space and others use `resource_attr` (MedianStoppingRule). We should unify this.

Kavlahkaff:
Unified naming convention of schedulers to use `time_attr` to define the fidelity space.

Closes #918 .

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ tuner = Tuner(
     scheduler=ASHA(
         config_space,
         metric='mean_loss',
-        time_attr='epoch',
+        resource_attr='epoch',
     ),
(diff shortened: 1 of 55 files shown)


geoalgo:
Could we use `time_attr` everywhere rather? It is the same as Raytune then https://docs.ray.io/en/latest/tune/api/doc/ray.tune.schedulers.AsyncHyperBandScheduler.html

aaronkl (approved):

PR #1025, opened 2026-05-13
fix: pass grace period

aaronkl:
pass grace period for ASHA baselines
----


Kavlahkaff (commented):
Thanks for the PR, LGTM! Just one question, is there a reason to only pass `grace_period ` and not other parameters like `reduction_factor` and `brackets`?

aaronkl:
Good catch! Added.

Kavlahkaff (approved):
Thanks, looks good!

PR #987, opened 2025-07-31
Remove typing imports

Kavlahkaff:
Removes type imports which will deprecate in the future, like List, Dict, Set, Tuple and replaces them with native types. Changed imports of Callable, Iterator and Iterable to import from collections.abc. 

Replaced Optional and Union, as described [here](https://peps.python.org/pep-0604/)

Used this [article](https://medium.com/@anuraagkhare_39064/typing-module-deprecated-in-python-what-you-need-to-know-d4348b694141) as linked in the issue.

Closes [#880](https://github.com/syne-tune/syne-tune/issues/880).

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1031, opened 2026-06-19
feat: add global optimiation problems

aaronkl:
- adds a new base class for blackboxes that are based on an artificial test functions
- implements several global optimization problems commonly used in the literature
- updates README and adds some unittests

----


Kavlahkaff (commented):
LGTM, thank you! Just one note, in #987 we decided to use native python types instead of importing from `typing`.

aaronkl:
ah right thanks for catching it

Kavlahkaff (approved):

PR #1032, opened 2026-06-26
feat: add foundation model searcher

aaronkl:
Adds a new search for foundation models that are trained on optimization trajectories of different optimization methods as described in the bbo-pile paper: https://arxiv.org/abs/2605.23417

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,6 +63,11 @@ extra = [
     # moo
     "pymoo>=0.6.0",
 ]
+fmbo = [
+    "huggingface_hub",
+    "transformers",
+    "vllm",
+]
 
 [project.urls]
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -63,6 +63,11 @@ extra = [
     # moo
     "pymoo>=0.6.0",
```
Kavlahkaff, line 64: Maybe we should add the `fmbo` dependencies to the `extra` section? This way I think the failing test will also run through.

Kavlahkaff (commented):
I only have a small comment regarding the failing test. Otherwise this looks good to me, thanks for adding it!

aaronkl:
good catch thanks!

Kavlahkaff (approved):

PR #1034, opened 2026-07-03
fix: infer noise level

aaronkl:
Adding additional noise leads to a fixed noise level for the GP (see [here](https://botorch.readthedocs.io/en/stable/models.html#module-botorch.models.gp_regression)). If we don't pass it, we let BOTorch infer the noise level automatically. 

----


Kavlahkaff (approved):
Thats a good catch! Did you test whether this now performs more competitive?

